PR #34, opened 2020-02-21
test: wait for expected replicas in WaitForDeployment

iaguis:
Let's expect the Kubernetes expected amount of replicas and remove the
expected replicas parameter from the function.

### DIFF
--- a/test/components/contour/contour_test.go
+++ b/test/components/contour/contour_test.go
@@ -42,7 +42,6 @@ func TestContourDeployment(t *testing.T) {
 	t.Parallel()
 	namespace := "projectcontour"
 	deployment := "contour"
-	replicas := 2
 
 	client, err := testutil.CreateKubeClient(t)
 	if err != nil {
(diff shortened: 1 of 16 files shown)


test/components/contour/contour_test.go
```diff
@@ -42,14 +42,13 @@ func TestContourDeployment(t *testing.T) {
...
 	client, err := testutil.CreateKubeClient(t)
 	if err != nil {
 		t.Errorf("could not create Kubernetes client: %v", err)
 	}
 	t.Log("got kubernetes client")
 
-	testutil.WaitForDeployment(t, client, namespace, deployment, replicas, time.Second*5, time.Minute*5)
-	t.Logf("Found required replicas: %d", replicas)
+	testutil.WaitForDeployment(t, client, namespace, deployment, time.Second*5, time.Minute*5)
+	t.Logf("Found required replicas")
```
invidian: This line can be moved to the `WaitForDeployment` function, to avoid duplicating it so many times. And to provide consistent output.

iaguis:
oh, forgot to rebase to remove the fixup commit :man_facepalming:

PR #55, opened 2020-02-26
Fix some typos

invidian:


blixtra:
But we should start following Golang convention for error strings. See https://github.com/golang/go/wiki/CodeReviewComments#error-strings

iaguis:
> But we should start following Golang convention for error strings. See https://github.com/golang/go/wiki/CodeReviewComments#error-strings

These are not errors that are returned: they are directly printed so in this case capitalization makes sense IMO.

PR #56, opened 2020-02-26
CopyWalker: truncate file when writing

### LINKED ISSUE #40: Writing assets to disk sometimes produces bad Terraform code
I'm not sure if this is only the issue when doing local development or also with standalone binaries, but I sometimes get some corrupted lines in assets files. Example:
```
$ cat lokomotive-assets/lokomotive-kubernetes/packet/flatcar-linux/kubernetes/bootkube.tf 
module "bootkube" {
  source       = "../../../bootkube"
  cluster_name = var.cluster_name

  # Cannot use cyclic dependencies on controllers or their DNS records
  api_servers          = [local.api_fqdn]
  api_servers_external = [local.api_external_fqdn]
  etcd_servers         = local.etcd_fqdn
  asset_dir            = var.asset_dir
  networking           = var.networking
  network_mtu          = var.network_mtu

  # Select private Packet NIC by using the can-reach Calico autodetection option with the first
  # host in our private CIDR.
  network_ip_autodetection_method = "can-reach=${cidrhost(var.node_private_cidr, 1)}"

  pod_cidr              = var.pod_cidr
  service_cidr          = var.service_cidr
  cluster_domain_suffix = var.cluster_domain_suffix
  enable_reporting      = var.enable_reporting
  enable_aggregation    = var.enable_aggregation

  certs_validity_period_hours = var.certs_validity_period_hours

  container_arch = var.os_arch
}
 expose_on_all_ports = true
}
```

invidian:
To avoid leaving old content untouched when updating files, which may
happen if new content is smaller than the old one.

Closes #40

### DIFF
--- a/pkg/util/walkers/copying.go
+++ b/pkg/util/walkers/copying.go
@@ -15,6 +15,7 @@
 package walkers
 
 import (
+	"fmt"
 	"io"
 	"os"
 	"path/filepath"
@@ -36,16 +37,23 @@ func CopyingWalker(path string, newDirPerms os.FileMode) assets.WalkFunc {
 			return errors.Wrap(err, "failed to create dir")
 		}
 
-		// TODO: If we start packing binaries, make sure they have executable bit set.
-		targetFile, err := os.OpenFile(fileName, os.O_RDWR|os.O_CREATE, 0644)
-		if err != nil {
-			return errors.Wrap(err, "failed to open target file")
-		}
-		defer targetFile.Close()
-
-		if _, err := io.Copy(targetFile, r); err != nil {
-			return errors.Wrap(err, "failed to write file")
-		}
-		return nil
+		return writeFile(fileName, r)
 	}
+}
+
+// writeFile writes data from given io.Reader to the file and makes sure, that
+// this is the only content stored in the file.
+func writeFile(p string, r io.Reader) error {
+	// TODO: If we start packing binaries, make sure they have executable bit set.
+	f, err := os.OpenFile(p, os.O_RDWR|os.O_CREATE, 0644)
+	if err != nil {
+		return fmt.Errorf("failed to open target file %s: %w", p, err)
+	}
+	defer f.Close()
+
+	if _, err := io.Copy(f, r); err != nil {
+		return fmt.Errorf("failed writing to file %s: %w", p, err)
+	}
+
+	return nil
 }
--- a/pkg/util/walkers/copying.go
+++ b/pkg/util/walkers/copying.go
@@ -45,7 +45,7 @@ func CopyingWalker(path string, newDirPerms os.FileMode) assets.WalkFunc {
 // this is the only content stored in the file.
 func writeFile(p string, r io.Reader) error {
 	// TODO: If we start packing binaries, make sure they have executable bit set.
-	f, err := os.OpenFile(p, os.O_RDWR|os.O_CREATE, 0644)
+	f, err := os.OpenFile(p, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0644)
 	if err != nil {
 		return fmt.Errorf("failed to open target file %s: %w", p, err)
 	}
--- a/pkg/util/walkers/copying_test.go
+++ b/pkg/util/walkers/copying_test.go
@@ -1,0 +1,50 @@
+// Copyright 2020 The Lokomotive Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package walkers
+
+import (
+	"io/ioutil"
+	"os"
+	"reflect"
+	"strings"
+	"testing"
+)
+
+const (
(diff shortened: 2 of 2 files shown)


pkg/util/walkers/copying.go
```diff
@@ -36,16 +37,24 @@ func CopyingWalker(path string, newDirPerms os.FileMode) assets.WalkFunc {
...
-			return errors.Wrap(err, "failed to write file")
-		}
-		return nil
+// writeFile writes data from given io.Reader to the file and makes sure, that
+// this is the only content stored in the file.
+func writeFile(p string, r io.Reader) error {
+	// TODO: If we start packing binaries, make sure they have executable bit set.
+	f, err := os.OpenFile(p, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0644)
+	if err != nil {
+		return fmt.Errorf("failed to open target file %s: %v", p, err)
```
iaguis: why not `errors.Wrapf()`?

pkg/util/walkers/copying_test.go
```diff
@@ -0,0 +1,62 @@
...
+		t.Fatalf("Writing to file should succeed, got: %v", err)
+	}
+
+	d, err := ioutil.ReadFile(f.Name())
+	if err != nil {
+		t.Fatalf("Reading tmp file should succeed, got: %v", err)
+	}
+
+	if !reflect.DeepEqual(testData, string(d)) {
+		t.Fatalf("expected: '%s', got '%s'", testData, string(d))
```
iaguis: For consistency
```suggestion
		t.Fatalf("Expected: '%s', got '%s'", testData, string(d))
```

pkg/util/walkers/copying_test.go
```diff
@@ -0,0 +1,62 @@
...
+		t.Fatalf("Updating file should succeed, got: %v", err)
+	}
+
+	d, err := ioutil.ReadFile(f.Name())
+	if err != nil {
+		t.Fatalf("Reading tmp file should succeed, got: %v", err)
+	}
+
+	if !reflect.DeepEqual(testData, string(d)) {
+		t.Fatalf("expected: '%s', got '%s'", testData, string(d))
```
iaguis: For consistency:

```suggestion
		t.Fatalf("Expected: '%s', got '%s'", testData, string(d))
```

pkg/util/walkers/copying.go
```diff
@@ -36,16 +37,24 @@ func CopyingWalker(path string, newDirPerms os.FileMode) assets.WalkFunc {
...
 	}
+
```
iaguis: I like having the defer right after the previous statement without any empty lines, is this something enforced by the linter?

pkg/util/walkers/copying_test.go
```diff
@@ -0,0 +1,62 @@
+package walkers
```
iaguis: Can we have a copyright header? Should we add a check for new files in the CI?

pkg/util/walkers/copying.go
invidian: Because from Go 1.13 `fmt.Errorf` is preferred method for wrapping errors. Earlier `errors.Wrapf` had to be used.

Though it should be `%w`, not `%v` to wrap properly.

pkg/util/walkers/copying.go
invidian: Fixed.

pkg/util/walkers/copying_test.go
invidian: >Can we have a copyright header?
Fixed.

>Should we add a check for new files in the CI?

Perhaps.

pkg/util/walkers/copying_test.go
invidian: Fixed.

pkg/util/walkers/copying_test.go
invidian: Fixed.

pkg/util/walkers/copying_test.go
invidian: Created https://github.com/kinvolk/lokomotive/issues/58 for tracking.

pkg/util/walkers/copying.go
iaguis: So should we do this then? :D

```suggestion
		return fmt.Errorf("failed to open target file %s: %w", p, err)
```

pkg/util/walkers/copying.go
```diff
@@ -36,16 +37,23 @@ func CopyingWalker(path string, newDirPerms os.FileMode) assets.WalkFunc {
...
+func writeFile(p string, r io.Reader) error {
+	// TODO: If we start packing binaries, make sure they have executable bit set.
+	f, err := os.OpenFile(p, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0644)
+	if err != nil {
+		return fmt.Errorf("failed to open target file %s: %v", p, err)
+	}
+	defer f.Close()
+
+	if _, err := io.Copy(f, r); err != nil {
+		return fmt.Errorf("failed writing to file %s: %v", p, err)
```
iaguis:
```suggestion
		return fmt.Errorf("failed writing to file %s: %w", p, err)
```

pkg/util/walkers/copying.go
iaguis: thanks for pointing me to this, TIL!

iaguis:
I'll modify the PR to use `%w` and create an issue to go through our uses of `fmt.Errorf` and `errors.Wrap*` and update them to the new convention.

pkg/util/walkers/copying.go
invidian: I eventually missed that :smile: Thanks for fixing.

BTW, if we print the error, e.g. in unit tests, `%v` should be used.

PR #71, opened 2020-02-28
aws workers: don't recreate the pool when launch configuration changes

### LINKED ISSUE #48: Changing kubernetes version triggers recreation of AWS worker pools
Since we use self-hosted `kubelet` and this is the strategy for node updates, we should probably ignore changes to pool `user_data` field.

invidian:
Closes #48


iaguis:
Can we have a bit more context in the commit message explaining that we don't want this because we use a self-hosted kubelet and so on?

invidian:
>Can we have a bit more context in the commit message explaining that we don't want this because we use a self-hosted kubelet and so on?

Done.

PR #32, opened 2020-02-20
Add ability to perform controlplane upgrades

invidian:

PR #72, opened 2020-02-28
Add ability to perform controlplane upgrades: alternative approach

invidian:
Alternative approach to #32. Opening separate PR to keep comments focused.

Contains commits from #71.


invidian:
Current issues:
- **Problem**: Currently we upgrade `podcheckpointer` together with `kube-apiserver`. If `podcheckpointer` fails, then `kube-apiserver` is lost and there is no way to recover from it, as `inactive-manifests` directory gets cleaned up.

  **Potential solution**: Update `podcheckpointer` separately.
- **Problem**: Unfortunately, as I said, pod checkpointer does not prevent or provide a way to recover if bad flag is specified for kube-apiserver, as there is no way to 
  distinguish between the "regular" crash, which is part of the upgrade process and the real crash, if e.g. bad flag has been specified.

  Additionally, bad flag is also persisted to inactive-manifests, so the recovery procedure is manual and difficult.

  **Proposed solution**: None so far

invidian:
I pushed the code, which separates pod checkpointer and kube-apiserver upgrades. However, it's still not working correctly, as pod checkpointer currently does not have any health checks defined and it seems, that Helm relies on them when doing `--atomic` upgrades.

Next, we should look how to avoid podcheckpointer caching crashing pods.

invidian:
>I pushed the code, which separates pod checkpointer and kube-apiserver upgrades. However, it's still not working correctly, as pod checkpointer currently does not have any health checks defined and it seems, that Helm relies on them when doing --atomic upgrades.

It looks like pod checkpointer does not implement any way of health checks, so we would have to implement one, to be able to provide reliable updates of it. Please note, that if pod checkpointer is not in a healthy state, `lokoctl` will not be able to restore it via upgrading helm release.

And if we try upgrading `kube-apiserver` without pod checkpointer running, cluster will break. Maybe we should look into https://github.com/kubernetes-sigs/bootkube/blob/master/Documentation/disaster-recovery.md and implement some of this functionality somehow.

>Next, we should look how to avoid podcheckpointer caching crashing pods.

I think that would again require some modification to pod checkpointer. Perhaps pod-checkpointer could not remove pods manifests, which are part of the daemonset and use them until new DaemonSet pod is not ready somehow. As far as I know, there is no such functionality at the moment.

#

As a side note, perhaps pod checkpointer could be modified to use kubelet's `kubeconfig` rather than service account token, so with Node restrictions, it would limit it to read only secrets of pods, which are actually assigned to this node.

invidian:
Created https://github.com/kinvolk/lokomotive/issues/83 to address lack of the probes on the workloads in general.

iaguis:
Should we close this?

invidian:
Yes, thanks @iaguis.

PR #73, opened 2020-02-28
Reordering the `pkg/components/util` package.

ipochi:
- helm.go and its test file mostly contained the code for charts. Hence
  renamed to charts.go and charts_test.go

- install.go contains mostly code for installing the chart as a release,
  hence renamed to release.go, so that in near future for uninstalling
  a release, we use the same file and not create another like
  `uninstall.go`

- release.go, charts.go and its test file is related to helm, hence its
  better to put them in a separate package under
  `pkg/components/utl/helmutil`

- lastly, made the changes required in the components which used the
  functions in the above files.

### DIFF
--- a/cli/cmd/component-install.go
+++ b/cli/cmd/component-install.go
@@ -21,7 +21,7 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/kinvolk/lokomotive/pkg/components"
-	"github.com/kinvolk/lokomotive/pkg/components/util"
+	"github.com/kinvolk/lokomotive/pkg/components/util/helmutil"
 	"github.com/kinvolk/lokomotive/pkg/config"
 )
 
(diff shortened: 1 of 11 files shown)


cli/cmd/component-install.go
```diff
@@ -21,7 +21,7 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/kinvolk/lokomotive/pkg/components"
-	"github.com/kinvolk/lokomotive/pkg/components/util"
+	"github.com/kinvolk/lokomotive/pkg/components/util/helmutil"
```
invidian: Maybe just `helm` to avoid duplication of `util`?

invidian:
ping @ipochi

ipochi:
Closing this, as discussed with @invidian. It doesn't make much sense to reorganize at this point in time.

PR #85, opened 2020-03-04
pkg/install/aws: use defaults from Terraform

invidian:
And only render values, which are actually defined. With this patch,
more defaults from Terraform are used, so it is easier to change those
defaults in the future.

This commit also reduces amount of Terraform configuration which is
generated, which should make debugging easier.


invidian:
>Magic number :D

That's one of the annoying linters I talked about :smile: Those numbers are not that "magic" :smile:

PR #91, opened 2020-03-04
Fix error wrapping across packages

invidian:
Even though we don't use it right now, since Go 1.13 it's recommended to use `%w` with combination of `fmt.Errorf`, so then receiver can unwrap underlying error.


iaguis:
Kicked the CI.

PR #100, opened 2020-03-05
Adding cert-manager component to AWS example configuration.

ipochi:


ipochi:
> Why having cert-manager if we cannot have ingress controller at the moment on AWS?

oh yes, that completely skipped my mind.

PR #107, opened 2020-03-05
Small documentation improvements

invidian:


iaguis:
We have a lot of links pointing to `configuration-guides/` still, we need to change those.

invidian:
>We have a lot of links pointing to configuration-guides/ still, we need to change those.

Oops, right. Will fix these.

invidian:
@iaguis fixed.

PR #108, opened 2020-03-05
Improve handling kubeconfig path

invidian:


invidian:
>Didn't check if the old logic is still in some other places (like components related functions, etc.), though. Did you? :)

Yes, all code use now `getKubeconfig()`.

PR #112, opened 2020-03-06
pkg/terraform/executor.go: include key in Output error message

invidian:
So it's easier to debug which key is wrong.


pkg/terraform/executor.go
```diff
@@ -293,7 +293,7 @@ func (ex *Executor) Plan() error {
 func (ex *Executor) Output(key string, s interface{}) error {
 	o, err := ex.ExecuteSync("output", "-json", key)
 	if err != nil {
-		return fmt.Errorf("failed getting Terraform output: %w", err)
+		return fmt.Errorf("failed getting Terraform output for key '%s': %w", key, err)
```
iaguis: maybe we can use `%q`?

invidian: TIL `%q`

PR #119, opened 2020-03-09
packet: improve hardware reservations support

rata:
This PR adds support for hardware reservations on Packet. Other patches in this PR were moved to another PR and are already merged (https://github.com/kinvolk/lokomotive/pull/146).

This patch set shows that we are lacking and is becoming more important to:
 - [x] Tests the function that validates valid configs, is really tricky. Will do if the approach looks reasonable to others
 - [x] Maybe the functionality regarding hardware reservations is too flexible and we can do some simplifications at the terraform layer. My guess is that it will be tricker, though, so haven't explored it much
 - [x] The syntax to specify `reservation_ids` and `reservation_ids_default` needs to have documentation (it needs a special format, etc.), where should that be? With terraform it was implicit in the variables description, but now? Any ideas?
 - [x] We are not checking that entries in the reservation_ids map are duplicated. I think this is not important (if it will cause a problem is detected, if entries are duplicated or not), but if you think otherwise let me know!

### logging

I modified the terraform executor (see commit "terraform: Add ApplyArgs() to reduce duplication") to log when running these operations. I don't like the log messages that I added, though, so if you have better ideas, please share them!

Right now, the logs when running cluster apply look like this:

```
INFO[0000] Initializing Terraform working directory      phase=infrastructure
INFO[0001] Applying part of the Terraform configuration. It might take a long time...  phase=infrastructure
INFO[0421] Applying Terraform configuration. This creates infrastructure so it might take a long time...  phase=infrastructure
```

The second line ("Applying part of the Terraform configuration...") is what I added, because that operation can take ~10 minutes before the third line is printed. But I don't like how similar they are lines 2 and 3 :-/

Also, while we were at it, this improves the logging when using DNS provider manual, as creating the controlers (and even workers creates! Another PR for that coming) can take ~10 minutes and therefore it made sense to use this new function there too.

### Not handling resource dependencies in terraform

What I really dislike about this approach, but decided to give it a try, is that it basically does `terraform -target` to avoid creating dependencies between terraform resources. If there are dependencies between resources, of course, the best thing is to express them in terraform.

This is not so straight forward, though, as the issue we are facing is an issue in terraform packet provider itself: https://github.com/terraform-providers/terraform-provider-packet/issues/176. The fix upstream, as there was no better alternative, was to document that you need to add dependencies between resources manually when you are using the hardware reservation feature and mixing "next-available" and specific UUIDs.

Using "next-available" and a specific UUID makes sense in lokomotive (we **need** both, one to spread nodes across racks for several reasons, like ECMP issues in packet, and the other one to simply use any reserved instance for pricing reasons). So, AFAICS, we can't ignore this problem in lokomotive.

The alternative will be to try to generate the dependencies programmatically, and if done correctly, just a `terraform apply` will work and simplify the logic on terraform creation.

I initially started going for the `terraform -target` approach as this feature was needed soon and there was already an instance of the same approach done for the manual DNS provider in Packet.

It is indeed easy to go after that approach, code is simple to reason about, etc. And it can be not very clean code the one to generate the terraform dependencies programmatically, but not completely sure.

While going the `terraform -target` approach, to make things sane, I started to simplify what was accepted as valid input and it might be easier with this added restrictions to go the "generate programmatically the terraform dependencies" route.

I'll try a little bit on this route and report back. Let me know if you have any thoughts about this.


rata:
@invidian @surajssd this is ready for review, please have a look! :)

invidian:
@rata run `make lint` locally, it's the linter which is failing. Or `make all` to also include all unit tests etc. And make sure your `master` branch is up to date. And see https://github.com/golangci/golangci-lint#install if you don't have linter installed locally.

pkg/platform/packet/packet.go
```diff
@@ -248,3 +277,151 @@ func (c *config) GetExpectedNodes() int {
...
+		d := checkEachReservation(w.ReservationIDs, w.ReservationIDsDefault, "worker", w.Name, w.Count)
+		diagnostics = append(diagnostics, d...)
+	}
+
+	return diagnostics
+
+}
+
+func checkEachReservation(reservationIDs map[string]string, resDefault, nodeRole, name string, count int) hcl.Diagnostics {
+	// This functions checks that hardware reservations are in the correct
```
invidian: Function documentation should be above the function declaration.

rata: I'm not sure I want that to be documentation (is not exported function also), but I can move it if you prefer to have it as doc

rata:
Added in a different commit (will squash later) changes to make the linter happy. The linter is still failing, however, with this error:

```
pkg/platform/packet/packet.go:425:23: mnd: Magic number: 2, in <condition> detected (gomnd)
		if len(resEntry) != 2 {
		                    ^
```

I think it is quite clear why it is using a 2 there and is not a magic number. I'm not sure how to make the linter happy and have a nice name (i.e. don't make the code worse). Any ideas?

pkg/platform/packet/packet.go
invidian: This makes it consistent with other function documentation I think.

invidian:
>I think it is quite clear why it is using a 2 there and is not a magic number. I'm not sure how to make the linter happy and have a nice name (i.e. don't make the code worse). Any ideas?

I think even `//nolint:gomnd` will do here :smile:

pkg/platform/packet/packet.go
rata: Moved it, thanks!

rata:
> > I think it is quite clear why it is using a 2 there and is not a magic number. I'm not sure how to make the linter happy and have a nice name (i.e. don't make the code worse). Any ideas?
> 
> I think even `//nolint:gomnd` will do here smile

Okay, done that. Also opened: #166 :)

rata:
@invidian @surajssd can you please review? Also, please look at the items in the PR description that are not solved yet, I'd appreciate any ideas regarding documentation :)

pkg/platform/packet/packet.go
```diff
@@ -34,20 +36,22 @@ import (
...
+	DisableBGP            bool              `hcl:"disable_bgp,optional"`
+	IPXEScriptURL         string            `hcl:"ipxe_script_url,optional"`
+	OSArch                string            `hcl:"os_arch,optional"`
+	OSChannel             string            `hcl:"os_channel,optional"`
+	OSVersion             string            `hcl:"os_version,optional"`
+	NodeType              string            `hcl:"node_type,optional"`
+	Labels                string            `hcl:"labels,optional"`
+	Taints                string            `hcl:"taints,optional"`
+	ReservationIDs        map[string]string `hcl:"reservation_ids,optional"`
+	ReservationIDsDefault string            `hcl:"reservation_ids_default,optional"`
```
invidian: Maybe `DefaultReservationID` ? It sounds better to me.

pkg/platform/packet/packet.go
```diff
@@ -237,6 +252,25 @@ func (c *config) terraformSmartApply(ex *terraform.Executor, dnsProvider dns.DNS
 	return ex.Apply()
 }
 
+// terraformCreateReservations creates nodes that use a specific hardware
+// reservation ID
```
```suggestion
// reservation ID.
```

Linter for that pending :smile: https://github.com/golangci/golangci-lint/pull/993

pkg/platform/packet/packet.go
```diff
@@ -237,6 +252,25 @@ func (c *config) terraformSmartApply(ex *terraform.Executor, dnsProvider dns.DNS
...
+func (c *config) terraformCreateReservations(ex *terraform.Executor) error {
+	arguments := []string{"apply", "-auto-approve"}
+
+	// Create workers that use specific UUIDS as hardware reservation
```
```suggestion
	// Create workers that use specific UUIDS as hardware reservation.
```

pkg/platform/packet/packet.go
```diff
@@ -237,6 +252,25 @@ func (c *config) terraformSmartApply(ex *terraform.Executor, dnsProvider dns.DNS
...
+	for _, w := range c.WorkerPools {
+		if len(w.ReservationIDs) > 0 {
+			arguments = append(arguments, fmt.Sprintf("-target=module.worker-%v.packet_device.nodes", w.Name))
+		}
+	}
+
+	// Create controllers that use specific UUIDS as hardware reservation
```
```suggestion
	// Create controllers that use specific UUIDS as hardware reservation.
```

pkg/platform/packet/packet.go
```diff
@@ -237,6 +252,25 @@ func (c *config) terraformSmartApply(ex *terraform.Executor, dnsProvider dns.DNS
...
+	if len(c.ReservationIDs) > 0 {
+		arguments = append(arguments, fmt.Sprintf("-target=module.packet-%v.packet_device.controllers", c.ClusterName))
+	}
+
+	return ex.Execute(arguments...)
+}
```
invidian: Functions should be separated with blank line.

pkg/platform/packet/packet.go
```diff
@@ -249,10 +283,12 @@ func (c *config) GetExpectedNodes() int {
 
 // checkValidConfig validates cluster configuration.
 func (c *config) checkValidConfig() hcl.Diagnostics {
+	// TODO: this function really needs to be tested
```
invidian: I'm not sure if this comment make sense.

Also, maybe add the tests then? :wink:

pkg/platform/packet/packet.go
```diff
@@ -294,3 +330,117 @@ func (c *config) checkWorkerPoolNamesUnique() hcl.Diagnostics {
 
 	return diagnostics
 }
+
```
```suggestion
// checkReservationIDs validates reservation IDs for controller nodes and all worker pools.
```

pkg/platform/packet/packet.go
```diff
@@ -294,3 +330,117 @@ func (c *config) checkWorkerPoolNamesUnique() hcl.Diagnostics {
...
+func (c *config) checkReservationIDs() hcl.Diagnostics {
```
```suggestion
func (c *config) validateReservationIDs() hcl.Diagnostics {
```

pkg/platform/packet/packet.go
```diff
@@ -294,3 +330,117 @@ func (c *config) checkWorkerPoolNamesUnique() hcl.Diagnostics {
...
+	}
+
+	return diagnostics
+}
+
+// checkEachReservation checks that hardware reservations are in the correct
+// format and, when it will cause problems, that reservation IDs values in this
+// pool are not mixed between using "next-available" and specific UUIDs, as this
+// can't work reliably.
+// For more info, see comment when calling terraformCreateReservations()
```
```suggestion
// For more info, see comment when calling terraformCreateReservations().
```

pkg/platform/packet/packet.go
```diff
@@ -294,3 +330,117 @@ func (c *config) checkWorkerPoolNamesUnique() hcl.Diagnostics {
...
+		if !strings.HasPrefix(key, resPrefix) {
+			diagnostics = append(diagnostics, hclErr)
+			// Don't duplicate the same error, show it one per key
+			continue
+		}
+
+		resEntry := strings.Split(key, "-")
+		if len(resEntry) != 2 { //nolint:gomnd
+			diagnostics = append(diagnostics, hclErr)
+			// Don't duplicate the same error, show it one per key
```
```suggestion
			// Don't duplicate the same error, show it one per key.
```

pkg/platform/packet/packet.go
```diff
@@ -294,3 +330,117 @@ func (c *config) checkWorkerPoolNamesUnique() hcl.Diagnostics {
...
+		resPrefix = "controller-"
+	}
+
+	d := checkResFormat(reservationIDs, name, errorPrefix, resPrefix)
+	diagnostics = append(diagnostics, d...)
+
+	return diagnostics
+}
+
+// checkResFormat checks that format for every key in reservationIDs is: <resPrefix>-<int>
```
```suggestion
// checkResFormat checks that format for every key in reservationIDs is: <resPrefix>-<int>.
```

pkg/platform/packet/packet.go
```diff
@@ -294,3 +330,117 @@ func (c *config) checkWorkerPoolNamesUnique() hcl.Diagnostics {
...
+func checkResFormat(reservationIDs map[string]string, name, errorPrefix, resPrefix string) hcl.Diagnostics {
```
invidian: This function should be easy to test and it would definitely improve having an idea how it works (to know valid and invalid input values. Can we add some?

pkg/platform/packet/packet.go
```diff
@@ -294,3 +330,117 @@ func (c *config) checkWorkerPoolNamesUnique() hcl.Diagnostics {
...
+
+	return diagnostics
+}
+
+// checkEachReservation checks that hardware reservations are in the correct
+// format and, when it will cause problems, that reservation IDs values in this
+// pool are not mixed between using "next-available" and specific UUIDs, as this
+// can't work reliably.
+// For more info, see comment when calling terraformCreateReservations()
+func checkEachReservation(reservationIDs map[string]string, resDefault, nodeRole, name string) hcl.Diagnostics {
```
invidian: Ditto here, it would be nice to have some tests for this function.

pkg/platform/packet/packet.go
invidian: It we want to be extra cautious, we should also check, that reservation IDs are not duplicated across the pools :see_no_evil:

invidian:
> Tests the function that validates valid configs, is really tricky. Will do if the approach looks reasonable to others

The approach looks good to me, I'd proceed with implementing tests, as mentioned in the review.

> The syntax to specify reservation_ids and reservation_ids_default needs to have documentation (it needs a special format, etc.), where should that be? With terraform it was implicit in the variables description, but now? Any ideas?

We have platform configuration reference document, here: https://github.com/kinvolk/lokomotive/blob/master/docs/configuration-reference/platforms/packet.md#argument-reference.

pkg/platform/packet/packet.go
rata: Yes, it's on the TODO list of the PR

rata: Marking as solved as it is already a TODO, we don't need to duplicate tracking

pkg/platform/packet/packet.go
rata: idem

pkg/platform/packet/packet.go
rata: Terraform fails saying it is already in use... do you think it is really needed? I doubt, to be honest

pkg/platform/packet/packet.go
rata: Yes, that was my internal TODO to add tests, not intended to be merged with that TODO. Sorry for the confusion :)

Test will be added, as mentioned in the PR description :)

pkg/platform/packet/packet.go
rata: Hm, I don't think it's worth diverging from terraform. For users coming from there, this would look simpler, and if we really want to, we can rename later (like deprecate, etc.)?

invidian: Well, I'm not sure if we should care so much about the Terraform here. `reservation_ids_default` suggests, that this takes multiple values, but I'm fine with it as it is. Feel free to mark as resolved if you think that's better for now.

pkg/platform/packet/packet.go
invidian: Well, IMO we should validate the configuration before even starting the configuration to follow "fail early" principle. But I don't think it's very important for now.

invidian:
ping @rata

rata:
@invidian I didn't forget about this. I won't have the BW to continue until next week, probably.

pkg/platform/packet/packet.go
rata: haha, cool :-D

rata: Fixed, will push soon.

pkg/platform/packet/packet.go
rata: Following the [link you posted](https://github.com/golangci/golangci-lint/pull/993) I see that [golang has documented the style](https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences) for *documentation* to have a period at the end. But haven't seen anything on *code* comments.

Did you find any place where it is documented? I had a quick look at go source code and it seems they use the same style for code comments too, except when comment is on the same line as code (i.e `a := "hola" // some comment with no period at the end`)

In any case, I think is fine to use always the same style. Just curious if you found it documented in some other place

pkg/platform/packet/packet.go
rata: Fixed

pkg/platform/packet/packet.go
rata: Added this instead:
```
// checkReservationIDs checks that reservations configured for controllers and
// workers are valid according to checkEachReservation().
```

pkg/platform/packet/packet.go
rata: Big no. All other functions are called `check*`, like `checkNotEmptyWorkers`, `checkWorkerPoolNamesUnique`. It's just not coherent with existing code

rata: Please re-open of you want to discuss this further

rata: I'm not sure I agree... this can open a can of worms and if we go down that route, start doing API calls to cloud providers to validate things (like if an DNS record already exists and will be possible to create it or fail because it exists).

If you want, I can create an issue to follow this later. But I'd like some guidance on what things you consider that should be catched, becase of course doing it for "all possible issues" can create the problem just mentioned.

pkg/platform/packet/packet.go
rata: Ok, thanks!

rata:
@invidian thanks again for the review! Marked a bunch of comments suggestions resolved as I fixed them locally (was easier due to other changes I already had).

It is still WIP, but should be close to it's final form (missing doc, as mentioned in the PR description).

pkg/platform/packet/packet_test.go
```diff
@@ -0,0 +1,169 @@
...
+			{
+				Name: "not",
+			},
+			{
+				Name: "dup",
+			},
+		},
+	}
+
+	hclErr := c.checkWorkerPoolNamesUnique()
```
invidian, line 56: Why not consolidate into single line? `if err := c.checkWorkerPoolNamesUnique(); err != nil {`

Also, if we test for diagnostics, we should use `err.HasErrors()`.

pkg/platform/packet/packet.go
```diff
@@ -294,3 +338,122 @@ func (c *config) checkWorkerPoolNamesUnique() hcl.Diagnostics {
...
+
+// checkResFormat checks that format for every key in reservationIDs is:
+// <resPrefix>-<int>.
+func checkResFormat(reservationIDs map[string]string, name, errorPrefix, resPrefix string) hcl.Diagnostics {
+	var diagnostics hcl.Diagnostics
+
+	for key := range reservationIDs {
+		hclErr := &hcl.Diagnostic{
+			Severity: hcl.DiagError,
+			Summary:  "Format used in reservation_ids invalid",
```
invidian, line 430: This error message sounds weird to me. How about:
```suggestion
			Summary:  "Invalid reservation ID",
```

or something?

pkg/platform/packet/packet_test.go
```diff
@@ -0,0 +1,169 @@
...
+
+	if hclErr != nil {
+		t.Errorf("Validation failed and shouldn't for worker pool: %v", c.WorkerPools[0].Name)
+	}
+}
+
+func TestCheckResFormatPrefixValidInput(t *testing.T) {
```
invidian, line 63: Tests for `checkResFormat` are not comprehensive. We have 4 execution branches and only 2 tests defined. If you look at the code coverage, it's clearly visible.

pkg/platform/packet/packet.go
```diff
@@ -294,3 +338,122 @@ func (c *config) checkWorkerPoolNamesUnique() hcl.Diagnostics {
...
+			Detail: fmt.Sprintf("%v: %v used \"%v\", format should be \"%v<int>\"",
```
invidian, line 431: For `name`, I would use `%q`.

pkg/platform/packet/packet_test.go
```diff
@@ -0,0 +1,169 @@
...
+	// Test with empty workers, it should fail.
+	c := config{}
+	hclErr := c.checkNotEmptyWorkers()
+
+	// Exactly 1 failure expected: no worker pools.
+	if len(hclErr) != 1 { // //nolint:gomnd
+		t.Errorf("Expected to detect 1 failure, but detected %v failures", len(hclErr))
+	}
+
+	// Test with one worker pool, it should work.
```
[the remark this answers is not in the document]
rata: I was on the fence because it is so short. But will do, thanks!

pkg/platform/packet/packet_test.go
rata: Good points!

I guess the first one is because I didn't write go code in a while :-(. Didn't know about the second one, but makes total sense and will check it out in case there other useful helpers!

pkg/platform/packet/packet.go
rata: You are right, now that I look at it after a few days seems weird. Your suggestion seems more clear, will change it. Thanks!

invidian:
> The syntax to specify reservation_ids and reservation_ids_default needs to have documentation (it needs a special format, etc.), where should that be? With terraform it was implicit in the variables description, but now? Any ideas?

I think this should be documented in configuration-reference for Packet: https://github.com/kinvolk/lokomotive/blob/master/docs/configuration-reference/platforms/packet.md.

pkg/platform/packet/packet.go
rata: Fixed, thanks!

pkg/platform/packet/packet.go
rata: Very nice! And for key! Started using it on several other places! Thanks :)

pkg/platform/packet/packet_test.go
rata: The downside of this is that you can't easily use the len of the slice, though.

I've moved it to use `HasErrors()` and lost precision because I only check if "HasErrors()", so checking that exactly the number of errors expected is returned (this will catch when an error is reported two times for the same line and we don't want it, for example) is not possible.

I'm not sure this is a big loss, though, but... I'm unsure if it's worth it.

Using the len of the slice is wrong, as it is possible that not all entries are errors. We can easily check the number of errors with a function like:

```
func ErrorsNr(d hcl.Diagnostics) int {
      ret := 0

      for _, diag := range d {
              if diag.Severity == hcl.DiagError {
                      ret++
              }
      }

      return ret
}
```

But I've had a quick look at terraform code and the test also use `HasErrors()`. So, I'm leaning towards just using that...

What do you think?

pkg/platform/packet/packet_test.go
rata: Thanks to point it out. Tests were not completed, but made sure it includes a test for each error path too now. Thanks :)

rata:
@invidian PTAL. I updated the PR description with some new things to look at, also. Please look that too :)

pkg/terraform/executor.go
```diff
@@ -138,6 +138,17 @@ func (ex *Executor) Apply() error {
 	return ex.Execute("apply", "-auto-approve")
 }
 
+// ApplyArgs is a wrapper function that runs
+// `terraform apply -auto-approve args...`.
+func (ex *Executor) ApplyArgs(args ...string) error {
+	ex.logger.Println("Applying part of the Terraform configuration. It might take a long time...")
```
invidian, line 144: I don't like that such basic function does the logging. It may also not be applying part of the Terraform configuration.

I'm not sure if we should add this function, for Terraform executor, having `Apply()`, `Destroy()` etc seems already too much for me.

If we decide to make it this way, maybe we could rename it to `ApplyAutoApprove(args []string)`?

docs/configuration-reference/platforms/packet.md
```diff
@@ -220,6 +220,8 @@ node_type = var.custom_default_worker_type
...
 | `worker_pool.taints`                  | Taints to assign to worker nodes.                                                                                                                                             | -               | false    |
+| `worker_pool.reservation_ids`         | Block with Packet hardware reservation IDs for worker nodes. Each key must have the format `worker-${index}` and the value is the reservation UUID. Can't be combined with `reservation_ids_default`. Example: `reservation_ids = { worker-0 = "<reservation_id>" }`.                                                | -               | false    |
+| `worker_pool.reservation_ids_default` | Default reservation ID for workers. The value`next-available` will choose any reservation that matches the pool's device type and facility. Can't be combined with `reservation_ids`.                                                                                                                                | -               | false    |
```
invidian, line 224:
```suggestion
| `worker_pool.reservation_ids_default` | Default reservation ID for workers. The value `next-available` will choose any reservation that matches the pool's device type and facility. Can't be combined with `reservation_ids`.                                                                                                                                | -               | false    |
```

docs/configuration-reference/platforms/packet.md
```diff
@@ -207,8 +207,8 @@ node_type = var.custom_default_worker_type
...
-| `reservation_ids_default`             | Default reservation ID for nodes not listed in the `reservation_ids`. The value`next-available` will choose any reservation that matches the pool's device type and facility. | ""              | false    |
+| `reservation_ids`                     | Block with Packet hardware reservation IDs for controller nodes. Each key must have the format `controller-${index}` and the value is the reservation UUID. Can't be combined with `reservation_ids_default`. Example: `reservation_ids = { controller-0 = "<reservation_id>" }`.                                    | -               | false    |
+| `reservation_ids_default`             | Default reservation ID for controllers. The value`next-available` will choose any reservation that matches the pool's device type and facility. Can't be combined with `reservation_ids`                                                                                                                             | -               | false    |
```
invidian, line 211:
```suggestion
| `reservation_ids_default`             | Default reservation ID for controllers. The value `next-available` will choose any reservation that matches the pool's device type and facility. Can't be combined with `reservation_ids`                                                                                                                             | -               | false    |
```

pkg/terraform/executor.go
rata: Would this solve your concern? Make it `ApplyTargets(args...string)` and it basically does a `terraform apply -auto-approve -target=args[0] -target=args[1] ...`?

I think while very reduced scoped, it seems better. It can't be misused, as it's quite clear when you want to use that instead of other functions, and would do just what we need (no more, no less).

What do you think?

docs/configuration-reference/platforms/packet.md
rata: Ups! Thanks!

docs/configuration-reference/platforms/packet.md
rata: Thanks!

rata:
@invidian (and other reviewers) I updated the PR description with some other thoughts I wanted to discuss. Please have a look :)

(github is so much worse than email... It's really complex to iterate. Sorry :-/)

pkg/terraform/executor.go
invidian: It seems to me, `ApplyTargets` is a bit better, as it operates on different level of abstraction (Terraform-scoped, not shell-scoped). And what about logging then?

rata: Logging stays the same, ApplyTarget() is applying part of your terraform configuration :)

And both are terraform scoped. **None** should ever be shell-scoped (in the sense of the command being run in a shell)

invidian: Can we maybe change the logging to mention which targets will be applied, instead of using "part of"?

pkg/platform/packet/packet_test.go
invidian: Yeah, I think we should use `HasErrors()` and this should be sufficient.

rata:
Closing in favor of: #299

PR #126, opened 2020-03-10
platform/packet: only allow calico networking

### LINKED ISSUE #124: Don't allow to use flannel on Packet
As then we cannot protect host ports. We don't really support it either, as we always deploy `calico-host-protection` on Packet, which breaks the installation.

iaguis:
We use calico for host protection on the Packet platform and we don't
want to disable host protection.

Closes #124


docs/configuration-reference/platforms/packet.md
```diff
@@ -205,7 +205,7 @@ node_type = var.custom_default_worker_type
...
 | `node_private_cidr`                   | Private IPv4 CIDR of the nodes used to allow inter-node traffic. Example "10.0.0.0/8"                                                                                         | -               | true     |
 | `enable_aggregation`                  | Enable the Kubernetes Aggregation Layer.                                                                                                                                      | true            | false    |
-| `networking`                          | CNI network plugin ("flannel" or "calico").                                                                                                                                   | "calico"        | false    |
+| `networking`                          | CNI network plugin (only "calico" supported on Packet).                                                                                                                       | "calico"        | false    |
```
invidian: Why can't we remove `networking` parameter completely?

iaguis: I was thinking on future networking that we might want to add so we don't have to re-add this later, but I'm not sure if that's a good enough reason...

invidian: I'm a bit concerned, that we still require user configuration block for it, even though it's the only valid option. We could leave it and make it a default, so user don't have to define it maybe?

iaguis: This is already the case, if you don't specify it in lokocfg it will take the terraform default which is calico.

invidian: Ah yes, I misread the code. I think it's fine then to leave it.

pkg/platform/packet/packet.go
```diff
@@ -101,6 +101,16 @@ func (c *config) LoadConfig(configBody *hcl.Body, evalContext *hcl.EvalContext)
 		}
 		return hcl.Diagnostics{err}
 	}
+
+	if c.Networking != "" && c.Networking != "calico" {
+		err := &hcl.Diagnostic{
+			Severity: hcl.DiagError,
+			Summary:  "Only calico networking is supported on Packet",
```
```suggestion
			Summary:  "Only Calico networking is supported on Packet",
```

as it's the project name 

or 

```suggestion
			Summary:  "Only 'calico' networking is supported on Packet",
```

to indicate, that this is the only valid _value_.

pkg/platform/packet/packet.go
```diff
@@ -101,6 +101,16 @@ func (c *config) LoadConfig(configBody *hcl.Body, evalContext *hcl.EvalContext)
...
+			Detail:   "We use calico for host protection on the Packet platform",
```
invidian: Ditto "calico".

iaguis:
We've decided to remove flannel. See #127. Closing.

PR #130, opened 2020-03-10
*: remove flannel and kube-router support

### LINKED ISSUE #127: Remove flannel
We don't really use it, it's not supported on Packet and we don't test it. Let's remove it and add it later if we see demand.

iaguis:
We don't really use them, they're not supported on Packet and we don't test
them. Let's remove them and add them later if we see demand.

This removes firewall rules for VXLAN for some platforms too. Note that
if we want to support the Calico's VXLAN encapsulation mode
(https://docs.projectcalico.org/getting-started/kubernetes/installation/config-options#switching-from-ip-in-ip-to-vxlan)
in the future we should enable those conditionally.

Closes #127


ipochi:
lgtm

PR #134, opened 2020-03-11
 *: change quiet flag to verbose and disable it by default 

### LINKED ISSUE #129: Hide terraform logs by default
We now show terraform logs when running lokoctl operations (e.g. refreshing state, logs of created/destroyed infrastructure...).

We should hide it by default by making `--quiet` the default and adding a `--verbose` flag.

iaguis:
So we don't show Terraform output by default.

Also, add more logging to show the execution phases and fix a bug where "terraform plan" was not showing the diff if `--verbose` is false, check commit messages for details.

Closes #129

### DIFF
--- a/cli/cmd/cluster-destroy.go
+++ b/cli/cmd/cluster-destroy.go
@@ -31,7 +31,7 @@ func init() {
 	clusterCmd.AddCommand(clusterDestroyCmd)
 	pf := clusterDestroyCmd.PersistentFlags()
 	pf.BoolVarP(&confirm, "confirm", "", false, "Destroy cluster without asking for confirmation")
-	pf.BoolVarP(&quiet, "quiet", "q", false, "Suppress the output from Terraform")
+	pf.BoolVarP(&verbose, "verbose", "v", false, "Show output from Terraform")
 }
 
 func runClusterDestroy(cmd *cobra.Command, args []string) {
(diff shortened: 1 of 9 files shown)


iaguis:
Noticed I didn't update the docs, will do after https://github.com/kinvolk/lokomotive/pull/136 is merged.

iaguis:
Also, when you create a cluster with this PR it stays for a long time in:

```
INFO[0001] Applying Terraform configuration              phase=infrastructure
```

And the user doesn't get any feedback that something is happening because we don't show terraform's logs. Maybe I can try to add a spinner or should I do this later?

invidian:
>And the user doesn't get any feedback that something is happening because we don't show terraform's logs. Maybe I can try to add a spinner or should I do this later?

I'd create an issue about it, to split install step of each platform into smaller `targets` and notify user which parts are created. I think that should give reasonable feedback then.

invidian:
@iaguis I'd also add a warning to not CTRL+C when this is running until we implement handling signals :smile:

iaguis:
> @iaguis I'd also add a warning to not CTRL+C when this is running until we implement handling signals

I added a sentence saying it might take a long time. I won't suggest not to CTRL+C since things should work if users wait a bit and it doesn't look great. Hopefully we can handle signals soon :)

PR #143, opened 2020-03-12
Small fixes to bootkube and kube-apiserver

invidian:
About removing pod anti affinities:
>With single controller node they are not needed, as they won't be
spreaded and with multiple, they will be installed as a DaemonSet, so
they are not needed either.


iaguis:
I seems CI didn't trigger for this PR :/

PR #144, opened 2020-03-12
Makefile: add lint-docker target 

invidian:


Makefile
```diff
@@ -59,7 +59,13 @@ test: run-unit-tests
 
 .PHONY: lint
 lint:
-	golangci-lint run --enable-all --disable=godox,gochecknoglobals --max-same-issues=0 --max-issues-per-linter=0 --build-tags aws,packet,e2e,disruptive-e2e --new-from-rev=$$(git merge-base $$(cat .git/resource/base_sha 2>/dev/null || echo "master") HEAD) --modules-download-mode=$(MOD) --timeout=5m --exclude-use-default=false ./...
+	git remote add linter https://github.com/kinvolk/lokomotive.git || true
```
invidian: I'm not sure if this is optimal, but that makes sure we always lint against upstream master branch.

rata: What we want almost always is "origin/master" at that is all.

It will fail in the cases that upstream is not origin? Sure. We can document it and be happy, IMHO

rata:
@invidian please check out this PR, that seems to fix (for me at least) all the errors I found to run the linter: https://github.com/kinvolk/lokomotive/pull/148

I think this makes sense in any case, though (the change to use `origin/master` is something I had locally and I'd push that or something along those lines) and running it in docker... not sure if it's worth, but no objections either :-)

Makefile
invidian: Well, if you fork the repository, usually `origin` points to the fork and as far as I know, not everyone keeps their fork master up to date, hence comparison with upstream.

invidian: I removed it, as it's causing troubles when remote is added via Docker (uid fun).

invidian:
@rata I changed it, so now it only adds option to run linter in Docker, which is at least some improvement. PTAL.

PR #146, opened 2020-03-12
packet: Use user-specified pool name for TF pool

rata:
This patches are extracted from https://github.com/kinvolk/lokomotive/pull/119. I found a bug in the code that I needed to change for that PR, so I fixed it.

That PR is not yet ready to be merged, so I'm including the fixes I found while writing it in this PR. Most of the commits are cosmetic (create a function for this and that) and make way more sense when PR #119 is merged, as those functions will not be so simple. To keep the backport simple, I decided to include them here, though

The fix is in patch: `packet: Use user-specified pool name for TF pool`, all the rest are cleanups or cosmetic changes. I c&p here the commit message from that patch, as it explains quite well what is doing and why it is important (let me know if I'm missing something!):

Before this commit, the pool name specified in lokoctl was being
completely ignored. The terraform code just iterated over the worker
pools, and the first one was named "pool-0", second "pool-1", etc.

This can create confusion for the user (an error on "pool-1" when the
name was "foo" is not meaningful, is difficult to know to which pools
servers belong). But, WAY more importantly, if the user reorders the
pools in the lokocfg file (or if for some reason that order is not
guaranteed by golang), it will create a mess in terraform.

This patch just makes that part sane: the terraform worker pools are
created with the user specified name. Therefore, errors in terraform are
meaningful to the user, the user can match servers hostnames in Packet
and can re-order pools without making terraform go nuts.

Also, this patch adds adds a check that all pool names are different

### DIFF
--- a/pkg/platform/packet/packet.go
+++ b/pkg/platform/packet/packet.go
@@ -138,41 +138,11 @@ func (c *config) Apply(ex *terraform.Executor) error {
 
 	c.AssetDir = assetDir
 
-	dnsProvider, err := dns.ParseDNS(&c.DNS)
-	if err != nil {
-		return errors.Wrap(err, "parsing DNS configuration failed")
-	}
-
 	if err := c.Initialize(ex); err != nil {
 		return err
 	}
 
-	// If the provider isn't manual, apply everything in a single step.
-	if dnsProvider != dns.DNSManual {
-		return ex.Apply()
-	}
-
-	arguments := []string{"apply", "-auto-approve"}
-
-	// Get DNS entries (it forces the creation of the controller nodes).
-	arguments = append(arguments, fmt.Sprintf("-target=module.packet-%s.null_resource.dns_entries", c.ClusterName))
-
-	// Add worker nodes to speed things up.
-	for index := range c.WorkerPools {
-		arguments = append(arguments, fmt.Sprintf("-target=module.worker-pool-%d.packet_device.nodes", index))
-	}
-
-	// Create controller and workers nodes.
-	if err := ex.Execute(arguments...); err != nil {
-		return errors.Wrap(err, "failed executing Terraform")
-	}
-
-	if err := dns.AskToConfigure(ex, &c.DNS); err != nil {
-		return errors.Wrap(err, "failed to configure DNS entries")
-	}
-
-	// Finish deployment.
-	return ex.Apply()
+	return c.terraformSmartApply(ex)
 }
 
 func (c *config) Destroy(ex *terraform.Executor) error {
@@ -237,6 +207,40 @@ func createTerraformConfigFile(cfg *config, terraformPath string) error {
 		return errors.Wrapf(err, "failed to write template to file: %q", path)
 	}
 	return nil
+}
+
+func (c *config) terraformSmartApply(ex *terraform.Executor) error {
+	dnsProvider, err := dns.ParseDNS(&c.DNS)
+	if err != nil {
+		return errors.Wrap(err, "parsing DNS configuration failed")
+	}
+
+	// If the provider isn't manual, apply everything in a single step.
+	if dnsProvider != dns.DNSManual {
+		return ex.Apply()
+	}
+
+	arguments := []string{"apply", "-auto-approve"}
+
+	// Get DNS entries (it forces the creation of the controller nodes).
+	arguments = append(arguments, fmt.Sprintf("-target=module.packet-%s.null_resource.dns_entries", c.ClusterName))
+
+	// Add worker nodes to speed things up.
+	for index := range c.WorkerPools {
+		arguments = append(arguments, fmt.Sprintf("-target=module.worker-pool-%d.packet_device.nodes", index))
+	}
+
+	// Create controller and workers nodes.
+	if err := ex.Execute(arguments...); err != nil {
+		return errors.Wrap(err, "failed executing Terraform")
+	}
+
+	if err := dns.AskToConfigure(ex, &c.DNS); err != nil {
+		return errors.Wrap(err, "failed to configure DNS entries")
+	}
+
+	// Finish deployment.
+	return ex.Apply()
 }
 
 func (c *config) GetExpectedNodes() int {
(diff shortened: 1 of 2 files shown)


pkg/platform/packet/packet.go
```diff
@@ -239,6 +209,40 @@ func createTerraformConfigFile(cfg *config, terraformPath string) error {
 	return nil
 }
 
+func (c *config) terraformSmartApply(ex *terraform.Executor) error {
+	dnsProvider, err := dns.ParseDNS(&c.DNS)
```
invidian: I would leave this function in the original place, as it's a "validation call" and not "action call" (like "ex.Apply" for example). This way, error returned from this function will always be a "runtime error".

invidian: It also reduces number of `return` calls to 4, which is what Code climate suggests as maximum.

pkg/platform/packet/packet.go
```diff
@@ -247,13 +248,31 @@ func (c *config) GetExpectedNodes() int {
...
-		err := &hcl.Diagnostic{
+		diagnostics = append(diagnostics, &hcl.Diagnostic{
 			Severity: hcl.DiagError,
 			Summary:  "At least one worker pool must be defined",
 			Detail:   "Make sure to define at least one worker pool block in your cluster block",
+		})
+	}
+
+	// Check that worker pools names are unique
+	dup := make(map[string]bool)
```
```suggestion
	dup := map[string]interface{}{}
```

Usually, when you only use map for getting unique keys, empty interface is used as a type, so you carry no information at all.

pkg/platform/packet/packet.go
```diff
@@ -247,13 +248,31 @@ func (c *config) GetExpectedNodes() int {
...
+	for _, w := range c.WorkerPools {
+		if !dup[w.Name] {
```
```suggestion
		if _, ok :=dup[w.Name]; !ok {
```

invidian: Hm, maybe boolean type actually make sense then, if this makes this condition easier. Seems "smart" though :smile:

pkg/platform/packet/packet.go
```diff
@@ -250,6 +250,15 @@ func (c *config) GetExpectedNodes() int {
 func (c *config) checkValidConfig() hcl.Diagnostics {
 	var diagnostics hcl.Diagnostics
 
+	diagnostics = append(diagnostics, c.checkNotEmptyWorkers()...)
+	diagnostics = append(diagnostics, c.checkWorkerPoolNamesUnique()...)
+
+	return diagnostics
+}
+
+func (c *config) checkNotEmptyWorkers() hcl.Diagnostics {
```
invidian: Can we please have some short comment for each function?

iaguis:
Since @rata is on vacation today I fixed the review comments that made sense to me. PTAL.

pkg/platform/packet/packet.go
```diff
@@ -248,3 +245,51 @@ func (c *config) GetExpectedNodes() int {
...
+func (c *config) checkValidConfig() hcl.Diagnostics {
+	var diagnostics hcl.Diagnostics
+
+	diagnostics = append(diagnostics, c.checkNotEmptyWorkers()...)
+	diagnostics = append(diagnostics, c.checkWorkerPoolNamesUnique()...)
+
+	return diagnostics
+}
+
+// Make sure there's at least one worker pool
```
```suggestion
// checkNotEmptyWorkers checks, if cluster has at least 1 node pool set.
```

All comments should end with dots, to resolve dispute if they should have a dot or not.

pkg/platform/packet/packet.go
```diff
@@ -248,3 +245,51 @@ func (c *config) GetExpectedNodes() int {
 
 	return c.ControllerCount + workers
 }
+
+// Check cluster config is valid
```
```suggestion
// checkValidConfig validates cluster configuration.
```

pkg/platform/packet/packet.go
```diff
@@ -248,3 +245,51 @@ func (c *config) GetExpectedNodes() int {
...
+			Severity: hcl.DiagError,
+			Summary:  "At least one worker pool must be defined",
+			Detail:   "Make sure to define at least one worker pool block in your cluster block",
+		})
+	}
+
+	return diagnostics
+}
+
+// Make sure there's no duplicate worker pool names
```
```suggestion
// checkWorkerPoolNamesUnique verifies, that all worker pool names are unique.
```

pkg/platform/packet/packet.go
```diff
@@ -239,6 +207,35 @@ func createTerraformConfigFile(cfg *config, terraformPath string) error {
 	return nil
 }
```
```suggestion
// terraformSmartApply applies cluster configuration.
```

pkg/platform/packet/packet.go
invidian: @iaguis when you're addressing the review comments, could you respond to them or ideally mark them as resolved please? That helps with consecutive reviews.

iaguis: Sure

pkg/platform/packet/packet.go
iaguis: I'll leave it as boolean

pkg/platform/packet/packet.go
iaguis: Done.

pkg/platform/packet/packet.go
iaguis: Done.

pkg/platform/packet/packet.go
iaguis: Done.

pkg/platform/packet/packet.go
iaguis: Done.

pkg/platform/packet/packet.go
```diff
@@ -248,3 +246,51 @@ func (c *config) GetExpectedNodes() int {
...
+
+	dup := make(map[string]bool)
+
+	for _, w := range c.WorkerPools {
+		if !dup[w.Name] {
+			dup[w.Name] = true
+			continue
+		}
+
+		// It is duplicated
```
```suggestion
		// It is duplicated.
```

rata:
@invidian @iaguis thanks!

PR #148, opened 2020-03-12
Makefile: Only run linter if code compiles

rata:
There are ton of cases where the linter can't run, with errors like:

	WARN [runner] Can't run linter goanalysis_metalinter: assign: failed prerequisites: inspect@github.com/kinvolk/lokomotive/pkg/platform/packet
	WARN [runner] Can't run linter unused: buildssa: analysis skipped: errors in package: [lokomotive/pkg/platform/packet/packet.go:271:20: cannot use arguments (variable of type []string) as string value in argument to ex.Execute]
	ERRO Running error: buildssa: analysis skipped: errors in package: [lokomotive/pkg/platform/packet/packet.go:271:20: cannot use arguments (variable of type []string) as string value in argument to ex.Execute]

The error is not crystal clear, but if you look closely it says there
were some errors in the package. While not very clear, that seems to
mean **compilation** errors.

I've faced this linter error several times in the last days and kept the
branches to later analyze (didn't find the reason immediately). I can
confirm that all the branches I kept where the linter failed to run with
similar errors, there were compilation errors and running "make lint"
(with this patch) makes it obvious.

Adding this dependency, the error is way clear, as it shows the
compilation error, like:

	pkg/platform/packet/packet.go:270:17: too many arguments in call to ex.Apply

This would have saved me quite some time, as this is way more clear.
And I sometimes trying to fix a linter error end up creating a syntax
error.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,7 +58,7 @@ build-slim:
 test: run-unit-tests
 
 .PHONY: lint
-lint:
+lint: build-slim
 	golangci-lint run --enable-all --disable=godox,gochecknoglobals --max-same-issues=0 --max-issues-per-linter=0 --build-tags aws,packet,e2e,disruptive-e2e --new-from-rev=$$(git merge-base $$(cat .git/resource/base_sha 2>/dev/null || echo "master") HEAD) --modules-download-mode=$(MOD) --timeout=5m --exclude-use-default=false ./...
 
 GOFORMAT_FILES := $(shell find . -name '*.go' | grep -v '^./vendor')
--- a/Makefile
+++ b/Makefile
@@ -58,7 +58,7 @@ build-slim:
 test: run-unit-tests
 
 .PHONY: lint
-lint: build-slim
+lint: build-slim build-test
 	golangci-lint run --enable-all --disable=godox,gochecknoglobals --max-same-issues=0 --max-issues-per-linter=0 --build-tags aws,packet,e2e,disruptive-e2e --new-from-rev=$$(git merge-base $$(cat .git/resource/base_sha 2>/dev/null || echo "master") HEAD) --modules-download-mode=$(MOD) --timeout=5m --exclude-use-default=false ./...
 
 GOFORMAT_FILES := $(shell find . -name '*.go' | grep -v '^./vendor')


rata:
@iaguis I bet the error you saw was this too.

Didn't check the linter doc, but I guess it only makes sense to run it when the code compiles :-D

Makefile
```diff
@@ -58,7 +58,7 @@ build-slim:
 test: run-unit-tests
 
 .PHONY: lint
-lint:
+lint: build-slim
```
invidian: you should also compile all tests before linting. This can be done by adding `build-test` target.

rata:
@invidian @iaguis thanks again! :)

Makefile
```diff
@@ -58,7 +58,7 @@ build-slim:
 test: run-unit-tests
 
 .PHONY: lint
-lint:
+lint: build-slim build-test
```
rata: @invidian @iaguis this `build-test` added is also running some tests:

```
testing: warning: no tests to run
PASS
coverage: 10.0% of statements
ok  	github.com/kinvolk/lokomotive/pkg/components/calico-hostendpoint-controller	0.035s	coverage: 10.0% of statements [no tests to run]
?   	github.com/kinvolk/lokomotive/pkg/components/cert-manager	[no test files]
testing: warning: no tests to run
PASS
coverage: 6.5% of statements
ok  	github.com/kinvolk/lokomotive/pkg/components/cluster-autoscaler	0.062s	coverage: 6.5% of statements [no tests to run]
testing: warning: no tests to run
PASS
...
```

I guess this was not the intention?

invidian: @rata are you sure it runs any tests? I don't think it should.

rata: Pasted the output when running `make lint` there... :)

invidian: For each package, you get `testing: warning: no tests to run`, meaning tests does not run. We also run it with `-v` and it does not show any test title.

Example output from running tests:
```
?   	github.com/kinvolk/lokomotive/pkg/components	[no test files]
=== RUN   TestRenderManifest
--- PASS: TestRenderManifest (0.00s)
=== RUN   TestEmptyConfig
--- PASS: TestEmptyConfig (0.00s)
=== RUN   TestRenderManifestBadConfig
--- PASS: TestRenderManifestBadConfig (0.00s)
PASS
coverage: 70.0% of statements
```

Also `make build-test  | grep Test` does not print anything.

Why then you think `build-test` runs some tests?

PR #150, opened 2020-03-13
Bump K8s version to 1.17.4

invidian:


invidian:
>Please change the README.md to the new version too.

Should we remove patch version of K8s shipped from README.md maybe? I think which major and minor version we ship is a feature, but patch version is just a maintenance task.

iaguis:
> Should we remove patch version of K8s shipped from README.md maybe? I think which major and minor version we ship is a feature, but patch version is just a maintenance task.

Sounds reasonable.

invidian:
I also updated Go dependencies. I think that should be part of the upgrade process too.

invidian:
>lgtm

Will wait for CI to pass before merging.

iaguis:
> Will wait for CI to pass before merging.

as always :)

PR #154, opened 2020-03-13
docs/packet: add ARM/x86 hybrid cluster instructions

iaguis:
Ported from
https://github.com/kinvolk/lokomotive-kubernetes/blob/master/docs/flatcar-linux/packet.md#arm-support-and-hybrid-clusters
and removing unnecessary details.

### DIFF
--- a/docs/configuration-reference/platforms/packet.md
+++ b/docs/configuration-reference/platforms/packet.md
@@ -8,6 +8,7 @@
 * [Argument reference](#argument-reference)
 * [Installing](#installing)
 * [Uninstalling](#uninstalling)
+* [ARM support and hybrid clusters](#arm-support-and-hybrid-clusters)
 
 ## Introduction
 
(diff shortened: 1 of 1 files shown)


docs/configuration-reference/platforms/packet.md
```diff
@@ -240,3 +241,37 @@ To destroy the Lokomotive cluster, execute the following command:
 lokoctl cluster destroy --confirm
 ```
 
+## ARM support and hybrid clusters
+
+Lokomotive and Flatcar currently include an alpha-quality preview for the
+Packet arm64 server types `c1.large.arm` and `c2.large.arm`. They can be used
```
pothos: I'm not sure if `c1.large.arm` still works, because we did not try it with Flatcar anymore for long time and last time I tried, it did not come up.

iaguis:
> LGTM, though we need should make sure, that all components which do not support ARM have selectors on amd64 nodes.

I added a note that components are not supported, but I agree we can make it nicer for users by doing that later.

docs/configuration-reference/platforms/packet.md
iaguis: I'll give it a try

iaguis: It doesn't seem to get past grub:

```
Booting `Flatcar default'

EFI stub: Booting Linux Kernel...
EFI stub: EFI_RNG_PROTOCOL unavailable, no randomness supplied
EFI stub: Using DTB from configuration table
EFI stub: Exiting boot services and installing virtual address map...
```

I'll remove it.

PR #169, opened 2020-03-17
ci: use spot instances for AWS testing

### LINKED ISSUE #149: Use spot instances for CI tests on AWS
We run CI quite often. On AWS, spot instances can be used to save costs. In most cases, this will be exactly like running with on-demand instances.

We will probably have to introduce some logic to detect if the test ends due to getting terminated and then restart the test in another region or with different instance types. But I would prose making the change without that logic to start with.

iaguis:
We're currently using t3.medium instances which cost $0.048 per hour.
Let's try to use spot instances with $0.02 as target price.

According to https://aws.amazon.com/ec2/spot/pricing/ the spot price for
a t3.medium is $0.0144 but even if we get a worse instance tests will
probably be ok.

We should revisit this when we enable single-node Lokomotive update
testing since that requires more memory.

Closes #149

### DIFF
--- a/ci/aws/aws-cluster.lokocfg.envsubst
+++ b/ci/aws/aws-cluster.lokocfg.envsubst
@@ -19,6 +19,7 @@ cluster "aws" {
     count       = 2
     ssh_pubkeys = ["$PUB_KEY"]
     disk_size   = 30
+    spot_price = "0.02"
     tags = {
       "deployment" = "ci"
     }


invidian:
Should we also enable it for controller nodes?

iaguis:
> Should we also enable it for controller nodes?

We probably should, but this is not ye supported by the terraform module.

PR #171, opened 2020-03-17
fix: Prefix cluster name to aws worker pool

ipochi:
- This commit fixes the condition where all AWS CI jobs have the same
  worker pool name, causing errors such as:

  ```
  error: Error creating LB Target Group: DuplicateTargetGroupName: A
  target group with the same name 'pool-ci-workers-http' exists, but
  with different settings
  ```

### DIFF
--- a/ci/aws/aws-cluster.lokocfg.envsubst
+++ b/ci/aws/aws-cluster.lokocfg.envsubst
@@ -15,7 +15,7 @@ cluster "aws" {
   os_channel   = var.os_channel
   ssh_pubkeys  = ["$PUB_KEY"]
 
-  worker_pool "pool-ci" {
+  worker_pool "$CLUSTER_ID-wp" {
     count       = 2
     ssh_pubkeys = ["$PUB_KEY"]
     disk_size   = 30


iaguis:
Heh:

```
Error: "name" cannot be longer than 32 characters
```

iaguis:
I think we might need to pass a smaller random string for worker pool names on the CI:

Currently, even without worker pool name:

```console
$ SUFFIX="$(printf '%b%b\n' "$(printf '\\%03o' "$((RANDOM % 26 + 97))")" "$(printf '\\%03o' $ "$((RANDOM % 26 + 97))")")"
$ CLUSTER_ID=l8eci$(date +%s)-$SUFFIX
$ echo -n $CLUSTER_ID-workers-https | wc -c
32
```

ipochi:
> I think we might need to pass a smaller random string for worker pool names on the CI.
> 
> Currently, even without worker pool name:
> 
> ```
> $ SUFFIX="$(printf '%b%b\n' "$(printf '\\%03o' "$((RANDOM % 26 + 97))")" "$(printf '\\%03o' $ "$((RANDOM % 26 + 97))")")"
> $ CLUSTER_ID=l8eci$(date +%s)-$SUFFIX
> $ echo -n $CLUSTER_ID-workers-https | wc -c
> 32
> ```

@iaguis https://github.com/kinvolk/lokomotive-ci/pull/102

Is this the correct appraoch ?

pothos:
If you tag all created resources with a tag for CI and information when it was created (afaik, not available for all resources), then you can stop having the date in the cluster prefix.
https://github.com/kinvolk/lokomotive-ci/pull/102 would make the workers invisible to the cleanup job.
So either shorten the date string or shorten the worker suffixes.

ipochi:
> If you tag all created resources with a tag for CI and information when it was created (afaik, not available for all resources), then you can stop having the date in the cluster prefix.
> [kinvolk/lokomotive-ci#102](https://github.com/kinvolk/lokomotive-ci/pull/102) would make the workers invisible to the cleanup job.
> So either shorten the date string (e.g., trim the seconds and later replace them with 00) or shorten the worker suffixes.

updated ci PR with removing the `l8eci` prefix to `ci`.

PR #180, opened 2020-03-17
pkg/aws: check cluster name and worker pool name length

### LINKED ISSUE #179: Validate max worker pool name length and cluster name length in AWs
As since #151, the maximum length is now limited, so using long cluster names or worker pool names results in following error:
```
module.aws-lokomotive-mateusz.module.workers.data.aws_ami.flatcar[0]: Refreshing state...

Error: "name" cannot be longer than 32 characters

  on ../lokomotive-kubernetes/aws/flatcar-linux/kubernetes/workers/ingress.tf line 26, in resource "aws_lb_target_group" "workers-https":
  26: resource "aws_lb_target_group" "workers-https" {


FATA[0010] error applying cluster: failed checking execution status: exit status 1  args="[]" command="lokoctl cluster apply"
```

Also, this affects `examples` configuration we provide.

iaguis:
There are tight restrictions on resource name length on AWS. Check
cluster name and worker pool name against those.

Fixes #179

### DIFF
--- a/pkg/platform/aws/aws.go
+++ b/pkg/platform/aws/aws.go
@@ -226,6 +226,36 @@ func (c *config) checkValidConfig() hcl.Diagnostics {
 
 	diagnostics = append(diagnostics, c.checkNotEmptyWorkers()...)
 	diagnostics = append(diagnostics, c.checkWorkerPoolNamesUnique()...)
+	diagnostics = append(diagnostics, c.checkNameSizes()...)
+
+	return diagnostics
+}
+
+// checkNameSizes checks the size of names since AWS has a limit of 32
+// characters on resources.
+func (c *config) checkNameSizes() hcl.Diagnostics {
+	var diagnostics hcl.Diagnostics
+
+	maxAWSResourceName := 32
+	maxNameLen := maxAWSResourceName - len("-workers-https") // This is the longest resource suffix.
+
+	if len(c.ClusterName) > maxNameLen {
+		diagnostics = append(diagnostics, &hcl.Diagnostic{
+			Severity: hcl.DiagError,
+			Summary:  "Cluster name too long",
+			Detail:   fmt.Sprintf("Maximum lenth is %d", maxNameLen),
+		})
+	}
+
+	for _, wp := range c.WorkerPools {
+		if len(wp.Name) > maxNameLen {
+			diagnostics = append(diagnostics, &hcl.Diagnostic{
+				Severity: hcl.DiagError,
+				Summary:  "Worker pool name too long",
+				Detail:   fmt.Sprintf("Maximum lenth is %d", maxNameLen),
+			})
+		}
+	}
 
 	return diagnostics
 }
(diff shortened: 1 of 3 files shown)


pkg/platform/aws/aws.go
```diff
@@ -226,6 +226,35 @@ func (c *config) checkValidConfig() hcl.Diagnostics {
...
+
+	return diagnostics
+}
+
+// checkNameSizes checks the size of names since AWS has a limit of 32
+// characters on resources.
+func (c *config) checkNameSizes() hcl.Diagnostics {
+	var diagnostics hcl.Diagnostics
+
+	maxNameLen := 18 // 32 - len("-workers-https"). This is the longest resource suffix.
```
invidian: Can we have this calculation in place rather than in the comment?

pkg/platform/aws/aws.go
```diff
@@ -226,6 +226,35 @@ func (c *config) checkValidConfig() hcl.Diagnostics {
...
+
+	if len(c.ClusterName) > maxNameLen {
+		diagnostics = append(diagnostics, &hcl.Diagnostic{
+			Severity: hcl.DiagError,
+			Summary:  "Cluster name too long",
+			Detail:   "Maximum lenth is 18",
```
invidian: should use `fmt.Sprintf` to include the max length variable.

pkg/platform/aws/aws.go
```diff
@@ -226,6 +226,35 @@ func (c *config) checkValidConfig() hcl.Diagnostics {
...
+		})
+	}
+
+	for _, wp := range c.WorkerPools {
+		if len(wp.Name) > maxNameLen {
+			diagnostics = append(diagnostics, &hcl.Diagnostic{
+				Severity: hcl.DiagError,
+				Summary:  "Worker pool name too long",
+				Detail:   "Maximum lenth is 18",
```
invidian: should use `fmt.Sprintf` to include the max length variable.

iaguis:
> Does worker pool name combines with cluster name in any place? If not, is it possible to have 2 worker pools with the same name on 2 different clusters? If so, should we also add a check for it's maximum length?

I don't think it does, I've tested with both cluster name and worker pool name being 18 characters and it worked fine.

I think having two worker pools with the same name results in errors like the one solved in https://github.com/kinvolk/lokomotive/pull/171

pkg/platform/aws/aws.go
iaguis: Sure, I'll have to tame the linter though.

pkg/platform/aws/aws.go
iaguis: Right, thanks for the catch.

pkg/platform/aws/aws.go
iaguis: Right, thanks for the catch.

invidian:
>I think having two worker pools with the same name results in errors like the one solved in #171

I wonder if we should document/report that :thinking:

iaguis:
> I wonder if we should document/report that thinking

I created https://github.com/kinvolk/lokomotive/pull/186

ipochi:
> > I think having two worker pools with the same name results in errors like the one solved in #171
> 
> I wonder if we should document/report that

In the same cluster , if there are two worker pools with same name, then there an appropriate error message provided.

PR #191, opened 2020-03-18
docs/configuration-reference: add httpbin reference 

iaguis:
Also, change wording in component documentation.

### DIFF
--- a/docs/configuration-reference/components/httpbin.md
+++ b/docs/configuration-reference/components/httpbin.md
@@ -1,0 +1,55 @@
+# httpbin configuration reference for Lokomotive
+
+## Contents
+
+* [Introduction](#introduction)
+* [Prerequisites](#prerequisites)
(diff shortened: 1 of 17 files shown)


docs/configuration-reference/components/httpbin.md
```diff
@@ -0,0 +1,55 @@
...
+```bash
+lokoctl component apply httpbin
+```
+
+## Destroying
+
+To destroy the component:
+
+```bash
+lokoctl component render-manifest httpbin | kubectl delete -f -
```
invidian: This is not ideal, as it leaves helm release as a garbage at the moment. We could say to do `helm uninstall`, not sure if we discussed that before.

iaguis: We did at some point, but I don't think it's a blocker.

PR #195, opened 2020-03-19
Rephrase command help

johananl:
This PR improves the phrasing of command help in lokoctl.

### DIFF
--- a/cli/cmd/cluster-apply.go
+++ b/cli/cmd/cluster-apply.go
@@ -34,8 +34,11 @@ var (
 
 var clusterApplyCmd = &cobra.Command{
 	Use:   "apply",
-	Short: "Apply configuration changes to a Lokomotive cluster with components",
-	Run:   runClusterApply,
+	Short: "Deploy or update a cluster",
+	Long: `Deploy or update a cluster.
+Deploys a cluster if it isn't deployed, otherwise updates it.
+Unless explicitly skipped, components listed in the configuration are applied as well.`,
+	Run: runClusterApply,
 }
 
 func init() {
--- a/cli/cmd/component-apply.go
+++ b/cli/cmd/component-apply.go
@@ -26,10 +26,11 @@ import (
 )
 
 var componentApplyCmd = &cobra.Command{
-	Use: "apply",
-	Short: `Apply a component configuration. If not present it will install it.
-If ran with no arguments it will apply all components mentioned in the
-configuration.`,
+	Use:   "apply",
+	Short: "Deploy or update a component",
+	Long: `Deploy or update a component.
+Deploys a component if not yet present, otherwise updates it.
+When run with no arguments, all components listed in the configuration are applied.`,
 	Run: runApply,
 }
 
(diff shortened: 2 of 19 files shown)


johananl:
Right, forgot we have the reference now. I guess we can enforce this in the CI.

cli/cmd/cluster-apply.go
```diff
@@ -34,8 +34,10 @@ var (
 
 var clusterApplyCmd = &cobra.Command{
 	Use:   "apply",
-	Short: "Apply configuration changes to a Lokomotive cluster with components",
-	Run:   runClusterApply,
+	Short: "Deploy or update a Lokomotive cluster",
+	Long: `Deploy or update a Lokomotive cluster.
+Unless explicitly skipped, Lokomotive components listed in the configuration are applied as well.`,
+	Run: runClusterApply,
```
joaquimrocha: nitpick: The commit title only mentions the short description was changed.

cli/cmd/component-apply.go
```diff
@@ -26,10 +26,10 @@ import (
 )
 
 var componentApplyCmd = &cobra.Command{
-	Use: "apply",
-	Short: `Apply a component configuration. If not present it will install it.
-If ran with no arguments it will apply all components mentioned in the
-configuration.`,
+	Use:   "apply",
+	Short: "Deploy or update a Lokomotive component",
+	Long: `Idempotently deploy a Lokomotive component.
```
joaquimrocha: I'd rather use a more clear description (i.e. I don't love the word "idempotently").
I think the short description already makes it clear, but if it's needed to be reenforced in the long one, maybe something like:
`Installs a component if it's not yet present, otherwise it updates it (applies the new confirguration).`

(The line below is fine)

cli/cmd/component-apply.go
```diff
@@ -26,10 +26,10 @@ import (
 )
 
 var componentApplyCmd = &cobra.Command{
-	Use: "apply",
```
joaquimrocha: (Commenting here because GH doesn't allow comments on the commit's description)

Typo in `Fix tens[e]` (also `ran` is not the future, but you don't even have to name the tenses :) ).

johananl: Oops. Fixed the typo :-)

Ran is past tense, and "If ran" is wrong because the intention here is "if the command **is run**", i.e. "if someone runs the command" but in passive form.

The "future -> present" is another fix, e.g. "will apply" -> "applies".

johananl: I've amended the commit message to make it clearer that I'm referring to two fixes.

cli/cmd/cluster-apply.go
johananl: Thanks, fixed.

cli/cmd/component-apply.go
johananl: I've rephrased the long description of both `component apply` and `cluster apply` since they are similar in nature.

johananl:
>Also, I am not sure if we should use Lokomotive components all the time, or even Lokomotive cluster in the help text.
It should be clear by the lokoctl description that the context for the commands is Lokomotive.

Yes, I've thought about this. I can remove all the superfluous `Lokomotive`s. My main reason for saying "Lokomotive components" is that the word "components" has too many meanings.

johananl:
@joaquimrocha I've removed the superfluous "Lokomotive" mentions from the help text. PTAL.

johananl:
Rebased. Need a re-approval :-)

Issue #181, opened 2020-03-17
Reduce the amount of tasks handled by Terraform


invidian:
Because of https://github.com/kinvolk/lokomotive-kubernetes origins, we had to implement many features via Terraform to keep the compatibility. Now that our main UX is `lokoctl`, we should be able to move more things to the Go code itself. We could perhaps move:
- all charts unpacking code
- generating manifests for static controlplane
- generate YAML values for controlplane components charts

This will make adding new features much easier, as we will avoid going trough so many layers of code to get things in place and perhaps make those pieces easier to test.

We could also consider moving PKI generation, currently done in bootkube module to Go, if that would make getting certificates rotation feature easier. Do note, that the certificates are currently stored in Terraform state, so we would either still have to store them there and find some other place for it.

Ideally I think, Terraform should only be responsible for cloud-related tasks.

PR #201, opened 2020-03-19
calico-host-protection: fix node name of HostEndpoint objects

invidian:
This commit fixes a regression introduced in
91a7bc2e2a62092780e8b69306ceb6130f104403, which was setting wrong value
for 'node' property of HostEndpoint objects, which was making Calico
policies not applicable for nodes, as the node name has been added with
-bond0 prefix.


assets/lokomotive-kubernetes/packet/flatcar-linux/kubernetes/calico-host-protection/templates/host-endpoints.yaml
```diff
@@ -10,5 +10,5 @@ metadata:
     {{- end }}
 spec:
   interfaceName: {{ .interfaceName }}
-  node: {{ .name }}
+  node: {{ .nodeName }}
```
rata: Can we create an issue to add tests for yamls generation too? (Not sure if helm has some built-in support to make testing easier, but we can see when tackling the issue)

assets/lokomotive-kubernetes/packet/flatcar-linux/kubernetes/calico-host-protection.yaml.tmpl
```diff
@@ -10,6 +10,7 @@ hostEndpoints:
 %{ for h in host_endpoints ~}
 - name: ${h.name}
   interfaceName: ${h.interface_name}
+  nodeName: ${h.node_name}
```
rata: I don't follow why we have this duplication here. Is this just for the controllers?

assets/lokomotive-kubernetes/packet/flatcar-linux/kubernetes/calico-host-protection/templates/host-endpoints.yaml
```diff
@@ -10,5 +10,5 @@ metadata:
     {{- end }}
 spec:
```
rata: Not sure why we have this file in the first place, either. The calico-hostendpoint-controller creates this for workers, right?

assets/lokomotive-kubernetes/packet/flatcar-linux/kubernetes/calico-host-protection/templates/host-endpoints.yaml
invidian: What kind of tests?

assets/lokomotive-kubernetes/packet/flatcar-linux/kubernetes/calico-host-protection/templates/host-endpoints.yaml
invidian: It was there for a long time. Perhaps it can be removed now so controller will take care of everything?

assets/lokomotive-kubernetes/packet/flatcar-linux/kubernetes/calico-host-protection.yaml.tmpl
invidian: This is Terraform template which generates Helm values.yaml file for the chart.  What do you mean by duplication?

assets/lokomotive-kubernetes/packet/flatcar-linux/kubernetes/calico-host-protection/templates/host-endpoints.yaml
rata: Test that generated yamls have the expected values

assets/lokomotive-kubernetes/packet/flatcar-linux/kubernetes/calico-host-protection/templates/host-endpoints.yaml
rata: For workers it was removed for a long time. I think I follow the code (sorry, next time will see it locally) and is only for controllers, so :ok_hand:

assets/lokomotive-kubernetes/packet/flatcar-linux/kubernetes/calico-host-protection/templates/host-endpoints.yaml
invidian: I'm not familiar with any test framework for Helm and I'm not sure if it's worth investing time into that. IMO functional tests are more important first. E.g. to see if all nodes has associated HostEndpoint objects.

I was hoping to get this PR merged quickly, as it fixes security issue and then implement regression test for it.

assets/lokomotive-kubernetes/packet/flatcar-linux/kubernetes/calico-host-protection/templates/host-endpoints.yaml
invidian: Do you think controller could also create the objects for controller nodes? We have boostrap firewall policies, so I think it should be fine?

assets/lokomotive-kubernetes/packet/flatcar-linux/kubernetes/calico-host-protection.yaml.tmpl
rata: This file calico-host-protection.yaml.tmpl seems like a template (given the name) and seems very similar to the yaml in the chart.

I see now that this is the "vules.yaml" file for the chart to use, but wasn't obvious from the name before to me.

I'd very much like to see how to simplify this flow in the future

assets/lokomotive-kubernetes/packet/flatcar-linux/kubernetes/calico-host-protection/templates/host-endpoints.yaml
rata: Sure, that is why I said the PR looks good :)

assets/lokomotive-kubernetes/packet/flatcar-linux/kubernetes/calico-host-protection/templates/host-endpoints.yaml
rata: I don't think I'd change who creates what, just wanted to make sure this was only for controllers.

I'd rather wait on that and see how to use the new coming feature in calico to replace the controller (if possible, if not maybe create another PR for calico adding what we miss :))

assets/lokomotive-kubernetes/packet/flatcar-linux/kubernetes/calico-host-protection.yaml.tmpl
invidian: https://github.com/kinvolk/lokomotive/issues/181 is one suggestion of how to simplify it

invidian:
>LGTM.

Thanks.

Issue #137, opened 2020-03-11
Calico should override bootstrap iptables rules


invidian:
Currently, we add bootstrap itpables rules in Packet here: https://github.com/kinvolk/lokomotive/blob/master/assets/lokomotive-kubernetes/packet/flatcar-linux/kubernetes/cl/controller.yaml.tmpl#L267, which apply to all interfaces.

Then we create Calico rules here: https://github.com/kinvolk/lokomotive/blob/master/assets/lokomotive-kubernetes/packet/flatcar-linux/kubernetes/calico-host-protection/templates/host-protection.yaml#L37.

However, modify Calico rules is not sufficient to open some ports, as bootstrap iptables rules are still enforced. This is confusing (e.g. see #120).

I think we should investigate that and make sure, that Calico rules always take precedence over bootstrap rules.

rata:
This seems very weird, because we rely on calico rules to open port 80 and 443, for example. Are you saying it does work for these ports but not for others? Or calico rules don't really open ports at all?

invidian:
It looks like I'm able to reach `kubelet` on port 10250 on controller node too by default :man_facepalming:

invidian:
Same with `kube-proxy` on port `10256`.

invidian:
This is now fixed by #201, but we still need to have some tests to prevent this from happening in the future.

invidian:
Well, #201 fixes it only partially. The bootstrap rules are still not overridden by Calico and as they are broader, they still have an effect. E.g.:
```
 1325  314K ACCEPT     tcp  --  *      *       0.0.0.0/0            0.0.0.0/0            tcp dpt:22
```

bootstrap rule is broader than following GNP:
```
apiVersion: crd.projectcalico.org/v1
kind: GlobalNetworkPolicy
metadata:
  name: ssh
spec:
  applyOnForward: true
  ingress:
  - action: Allow
    destination:
      ports:
      - 22
    protocol: TCP
    source:
      nets:
      - 1.1.1.1/16
  order: 0
  preDNAT: true
  selector: has(host-endpoint)
```

rata:
@invidian that should be a regression too. I know for a fact that the calico ssh rules take precedence over the bootstrap iptables rules before the helm refactors. We are either testing incorrectly or regressed, quite sure :)

rata:
> Well, #201 fixes it only partially. The bootstrap rules are still not overridden by Calico and as they are broader, they still have an effect. E.g.:

[...]

PR #202, opened 2020-03-20
Packet: improve bootstrap iptables rules

invidian:
This PR further locks down bootstrap iptables rules. The disadvantage though, which we need to consider is, that if user updates some of those CIDRs and rely on them, then Calico rules are flushed for whatever reason, the original rules will be applied (from node creation time), unless the node is replaced during the process. We need to make sure, that this is the behavior we want.

Alternatively, changing CIDRs would require node upgrade rollout we plan to implement. 

Refs #137 #8


assets/lokomotive-kubernetes/packet/flatcar-linux/kubernetes/cl/controller.yaml.tmpl
```diff
@@ -266,11 +266,13 @@ storage:
           :FORWARD ACCEPT [0:0]
           :OUTPUT ACCEPT [0:0]
           -A INPUT -i lo -j ACCEPT
-          -A INPUT -p tcp --dport 22 -j ACCEPT
-          -A INPUT -p tcp --dport 179 -j ACCEPT
-          -A INPUT -p tcp --dport 2379 -j ACCEPT
-          -A INPUT -p tcp --dport 2380 -j ACCEPT
-          -A INPUT -p tcp --dport 2381 -j ACCEPT
+          %{ for management_cidr in management_cidrs ~}
+          -A INPUT -s ${management_cidr} -p tcp --dport 22 -j ACCEPT
```
rata, line 270: Grr, I'd say we disable ssh (and it will be open when calico starts) but bootkube will fail :-/

assets/lokomotive-kubernetes/packet/flatcar-linux/kubernetes/controllers.tf
```diff
@@ -62,36 +62,21 @@ resource "packet_device" "controllers" {
...
-  content = data.template_file.controller-install[count.index].rendered
-}
-
-data "template_file" "controller-install" {
-  count    = var.controller_count
-  template = file("${path.module}/cl/controller-install.yaml.tmpl")
-
-  vars = {
+  count = var.controller_count
+  content = templatefile("${path.module}/cl/controller-install.yaml.tmpl", {
```
rata, line 66: why is this changed? The linked issue doesn't say anything either. Is it a simplification at the terraform layer with no effect on the generated config?

assets/lokomotive-kubernetes/packet/flatcar-linux/kubernetes/calico-host-protection.yaml.tmpl
```diff
@@ -3,7 +3,7 @@ clusterCIDRs:
 - ${cidr}
 %{ endfor ~}
 managementCIDRs:
-%{ for cidr in management_cidrs ~}
+%{~ for cidr in management_cidrs }
```
rata, line 6: Was it broken before? How? Or is it a cosmetic change with no effect to the managementCIDRs effectively used? Can you please clarify? :)

assets/lokomotive-kubernetes/packet/flatcar-linux/kubernetes/workers/workers.tf
```diff
@@ -20,22 +20,16 @@ resource "packet_device" "nodes" {
...
-data "ct_config" "install-ignitions" {
-  content = data.template_file.install.rendered
-}
-
 # These configs are used for the fist boot, to run flatcar-install
-data "template_file" "install" {
-  template = file("${path.module}/cl/install.yaml.tmpl")
-
-  vars = {
+data "ct_config" "install-ignitions" {
```
rata, line 24: Don't see the need to split both commits in workers and controllers, if they are basically doing the same thing. IMHO, it is a little harder to review just checking if the same was done in the other commit.

assets/lokomotive-kubernetes/packet/flatcar-linux/kubernetes/workers/cl/worker.yaml.tmpl
```diff
@@ -315,10 +315,12 @@ storage:
           :FORWARD ACCEPT [0:0]
           :OUTPUT ACCEPT [0:0]
           -A INPUT -i lo -j ACCEPT
-          -A INPUT -p tcp --dport 22 -j ACCEPT
-          -A INPUT -p tcp --dport 179 -j ACCEPT
-          -A INPUT -p tcp --dport 10250 -j ACCEPT
-          -A INPUT -p tcp --dport 10256 -j ACCEPT
+          %{ for management_cidr in management_cidrs ~}
+          -A INPUT -s ${management_cidr} -p tcp --dport 22 -j ACCEPT
+          %{~ endfor }
```
rata, line 320: I don't think ssh is needed for workers before they are running calico (out of band console can be used if needed). What do you think?

assets/lokomotive-kubernetes/packet/flatcar-linux/kubernetes/cl/controller.yaml.tmpl
```diff
@@ -266,11 +266,13 @@ storage:
...
+          %{~ endfor }
+          -A INPUT -s ${node_private_cidr} -p tcp --dport 179 -j ACCEPT
+          -A INPUT -s ${node_private_cidr} -p tcp --dport 2379 -j ACCEPT
+          -A INPUT -s ${node_private_cidr} -p tcp --dport 2380 -j ACCEPT
+          -A INPUT -s ${node_private_cidr} -p tcp --dport 2381 -j ACCEPT
```
rata, line 275: Not sure if we should use the `node_private_cidr` or just `10.0.0.0/8`. I like the simplicity of just using `10.0.0.0/8`, but... no strong opinion.

What do you think?

assets/lokomotive-kubernetes/packet/flatcar-linux/kubernetes/controllers.tf
invidian: with tf 0.12, you no longer need `template_file` baked by 3rd party provider, as Terraform provides built-in `templatefile`, so it's recommended ot use this one.

rata: Oh, okay. I see in [the docs](https://www.terraform.io/docs/providers/template/d/file.html) that it says so. Linking and explaining in the commit would be nice :)

invidian: I'll add it, though I'm not fan of duplicating this information, as there will be more commits in the future relating to the same problem. We use GitHub, we should make use of it :smile:

assets/lokomotive-kubernetes/packet/flatcar-linux/kubernetes/workers/cl/worker.yaml.tmpl
invidian: It seems reasonable, though opinionated, so it should be documented. If calico fails to start for some reason, it makes it more difficult to debug. What do you think @kinvolk/team-cloud?

assets/lokomotive-kubernetes/packet/flatcar-linux/kubernetes/workers/workers.tf
invidian: True, will squash.

assets/lokomotive-kubernetes/packet/flatcar-linux/kubernetes/cl/controller.yaml.tmpl
invidian: We actually copy secrets over SSH to all controller nodes, not only to the bootstrap one :/

assets/lokomotive-kubernetes/packet/flatcar-linux/kubernetes/cl/controller.yaml.tmpl
invidian: `node_private_cidr` could default to `10.0.0.0/8` maybe?

assets/lokomotive-kubernetes/packet/flatcar-linux/kubernetes/workers/cl/worker.yaml.tmpl
invidian: Hm, maybe port 22 should also be allowed from `node_private_cidr`?

assets/lokomotive-kubernetes/packet/flatcar-linux/kubernetes/controllers.tf
rata: Don't know what you mean. The linked issue is not clear to me that upstream recommends not using it anymore.

It's easier to review if you have the context you need. If you prefer to link to the issue only, instead of adding some lines, fine (is very easy to make it easier for reviewers by including those lines, but if you really don't want to do it, don't.). But let's make the issue clear on that, at least.

assets/lokomotive-kubernetes/packet/flatcar-linux/kubernetes/cl/controller.yaml.tmpl
rata: Now that I think about it, this has been discussed in other issues, I think. The problem is that these rules can't be changed... so, it's complicated. I think both are fine for me (`10.0.0.0/8` vs `node_private_cidr`). Do whatever you consider best :)

assets/lokomotive-kubernetes/packet/flatcar-linux/kubernetes/cl/controller.yaml.tmpl
rata: Yeah, exactly :-/

Totally out of scope, but I'd like to play with adding those to ignition compressed (it seems ignition supports compression, not sure if packet will get in the way or if we are using something older, etc.)

rata: But yeah, nothing to do here with this, we need it for now :-/

assets/lokomotive-kubernetes/packet/flatcar-linux/kubernetes/workers/workers.tf
rata: Thanks!

assets/lokomotive-kubernetes/packet/flatcar-linux/kubernetes/calico-host-protection.yaml.tmpl
johananl: In this specific case there is no difference between `{~ }` and `{ ~}` AFAICT. The former deletes the newline on the preceding line and the latter deletes the newline on the same line. It *would* matter if the following line was indented, for example.

assets/lokomotive-kubernetes/packet/flatcar-linux/kubernetes/workers/cl/worker.yaml.tmpl
johananl: - IMO we want to be able to SSH into workers at all times, and especially during failures.
- I don't think we want to allow SSH from `node_private_cidr`. This increases the nodes' attack surface (very theoretically) but mainly I don't find it useful. My guess is you're suggesting this to allow using a bastion host (right?), in which case I would include the bastion's `/32` address in `management_cidrs` because it's a management CIDR.

assets/lokomotive-kubernetes/packet/flatcar-linux/kubernetes/calico-host-protection.yaml.tmpl
invidian: The difference shows up, when there is no `management_cidrs` specified, then:
```yaml
managementCIDRs:


```
is produced, instead of:
```yaml
managementCIDRs:
```

Not a big difference, but I spotted it as a unnecessary new line when testing the template. Can drop this patch as well actually. 

Also added explanation in the commit message.

assets/lokomotive-kubernetes/packet/flatcar-linux/kubernetes/controllers.tf
invidian: Added a bit more context to both issue and commit message :)

assets/lokomotive-kubernetes/packet/flatcar-linux/kubernetes/workers/workers.tf
invidian: Done

assets/lokomotive-kubernetes/packet/flatcar-linux/kubernetes/cl/controller.yaml.tmpl
invidian: Decided to use `10.0.0.0/8`. Seems like a sane bootstrap value.

invidian:
Addressed all review feedback, please have a look again.

invidian:
>LGTM. However, I'd add a comment saying why we use 10.0.0.0/8 as that is a trck not obvious for most readers not very involved with Packet networking internals or any new comer.

Make sense, done.

assets/lokomotive-kubernetes/packet/flatcar-linux/kubernetes/calico-host-protection.yaml.tmpl
rata: Thanks!

assets/lokomotive-kubernetes/packet/flatcar-linux/kubernetes/controllers.tf
rata: Thanks!

PR #207, opened 2020-03-23
components/external-dns: expose owner_id

iaguis:
We were setting it to the AWS DNS Zone ID and that means that several
clusters using the same DNS Zone will interact badly with each other.

This exposes it to the user so they can choose a reasonable ID.

### DIFF
--- a/ci/aws/aws-cluster.lokocfg.envsubst
+++ b/ci/aws/aws-cluster.lokocfg.envsubst
@@ -40,7 +40,8 @@ component "cert-manager" {
 }
 
 component "external-dns" {
-  policy = "sync"
+  policy   = "sync"
+  owner_id = "$CLUSTER_ID"
   aws {
     zone_id               = "$AWS_DNS_ZONE_ID"
     aws_access_key_id     = "$AWS_ACCESS_KEY_ID"
--- a/ci/packet/packet-cluster.lokocfg.envsubst
+++ b/ci/packet/packet-cluster.lokocfg.envsubst
@@ -47,7 +47,8 @@ component "cert-manager" {
 }
 
 component "external-dns"{
-  policy = "sync"
+  policy   = "sync"
+  owner_id = "$CLUSTER_ID"
   aws {
     zone_id = "$AWS_DNS_ZONE_ID"
     aws_access_key_id = "$AWS_ACCESS_KEY_ID"
--- a/docs/configuration-reference/components/external-dns.md
+++ b/docs/configuration-reference/components/external-dns.md
@@ -59,17 +59,18 @@ Table of all the arguments accepted by the component.
 
 Example:
 
-| Argument                    | Description                                                                                                       | Default        | Required |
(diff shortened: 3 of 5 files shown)


ci/aws/aws-cluster.lokocfg.envsubst
```diff
@@ -41,6 +41,7 @@ component "cert-manager" {
 
 component "external-dns" {
   policy = "sync"
+  owner_id = "$CLUSTER_ID"
```
invidian: Parameters should be aligned.

ci/packet/packet-cluster.lokocfg.envsubst
```diff
@@ -48,6 +48,7 @@ component "cert-manager" {
 
 component "external-dns"{
   policy = "sync"
+  owner_id = "$CLUSTER_ID"
```
invidian: Ditto alignment.

docs/configuration-reference/components/external-dns.md
```diff
@@ -59,17 +59,18 @@ Table of all the arguments accepted by the component.
...
+|-----------------------------|-----------------------------------------------------------------------------------------------------------------------------------|:--------------:|:--------:|
+| `sources`                   | Kubernetes resources type to be observed for new DNS entries by ExternalDNS.                                                      | ["service"]    | false    |
+| `namespace`                 | Namespace to install ExternalDNS.                                                                                                 | "external-dns" | false    |
+| `policy`                    | Modify how DNS records are sychronized between sources and providers (options: sync, upsert-only).                                | "upsert-only"  | false    |
+| `metrics`                   | Enable metrics collection by Prometheus. Needs [Prometheus Operator component](prometheus-operator.md) installed.                 | false          | false    |
+| `owner_id`                  | A name that identifies this instace of ExternalDNS. Set it to a unique value that doesn't change for the lifetime of the cluster. | -              | true     |
```
invidian: I think we should mention, that this should be unique across the zone ID.

Alternatively, we could make it optional, so it only needs to be set if you have more than one cluster sharing the zone ID. That would be error-prone though.

iaguis: This is supposed to be platform-agnostic, but we only support AWS for now so I'll add it.

ci/aws/aws-cluster.lokocfg.envsubst
iaguis: Done.

ci/packet/packet-cluster.lokocfg.envsubst
iaguis: Done.

docs/configuration-reference/components/external-dns.md
invidian: We could say across DNS zone.

invidian: Ah, this is what you did, cool :)

PR #209, opened 2020-03-23
test/components/util: don't return errors when taking testing.T as arg

invidian:
As this is an anti-pattern when writing tests in Go. When helper
function takes testing.T as an argument and errors occurs, t.Fatalf or
t.Errorf should be used instead of returning error.


invidian:
BTW, majority of the tests could be further refactored into single large matrix tests for Deployments, DaemonSets etc. Maybe that would also cover the per-platform compatibility testing we struggle in #140?

PR #230, opened 2020-03-30
docs: clarify values for ssh_pubkeys

iaguis:
They should be the full public key and this is not obvious.

Also, show a more descriptive value for the quickstart examples.


docs/configuration-reference/platforms/aws.md
```diff
@@ -178,7 +178,7 @@ worker_pool "my-worker-pool" {
...
 | `dns_zone`                    | Route 53 DNS Zone.                                                                                                                                                                         | -               | true     |
 | `dns_zone_id`                 | Route 53 DNS Zone ID.                                                                                                                                                                      | -               | true     |
-| `ssh_pubkeys`                 | SSH public keys for user `core`.                                                                                                                                                           | -               | true     |
+| `ssh_pubkeys`                 | SSH public keys for user `core`. This is the full public key as shown by `ssh-add -L`.                                                                                                     | -               | true     |
```
johananl: This assumes the user users ssh-agent. I sometimes encounter tech folks who don't know about ssh-agent or wouldn't use it for some other reason. Also, assuming most users will have the key pair as local files, I would consider using the word "contents" to make it very clear they have to paste the contents of the public key file rather than a reference to the file path.

iaguis: Updated.

docs/configuration-reference/platforms/packet.md
```diff
@@ -194,7 +194,7 @@ node_type = var.custom_default_worker_type
...
 | `facility`                            | Packet facility to use for deploying the cluster.                                                                                                                             | -               | false    |
 | `project_id`                          | Packet project ID.                                                                                                                                                            | -               | true     |
-| `ssh_pubkeys`                         | SSH public keys for user `core`.                                                                                                                                              | -               | true     |
+| `ssh_pubkeys`                         | SSH public keys for user `core`. This is the contents of the keys.                                                                                                            | -               | true     |
```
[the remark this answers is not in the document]
iaguis: That sounds better. I'll use this.

PR #233, opened 2020-03-30
docs/quickstarts: add reference to usage docs and PSP note

iaguis:
By default the cluster will refuse to run most images because they run
as root and that's disallowed by the PodSecurityPolicy. Mention this in
the quickstart pointing to the relevant documentation.

Also, link to the Kubernetes Basics documentation so inexperienced users
can learn about Kubernetes usage.

Reported-by: @jpetazzo

### DIFF
--- a/docs/concepts/securing-lokomotive-cluster.md
+++ b/docs/concepts/securing-lokomotive-cluster.md
@@ -33,6 +33,9 @@ default PSPs for general purpose application usage:
     Allowed to workloads in `kube-system` namespace only. This PSP does not restrict any workloads
     from any required permissions.
 
+Many projects provide their own Pod Security Policies tailored to their needs which can be used
+when deploying if the policies provided by Lokomotive are too strict.
+
 ## Global network policy (for Packet platform only)
 
 Lokomotive installs Calico’s
--- a/docs/quickstarts/aws.md
+++ b/docs/quickstarts/aws.md
@@ -163,6 +163,19 @@ export KUBECONFIG=./lokomotive-assets/cluster-assets/auth/kubeconfig
 kubectl get nodes
 ```
 
+## Using the cluster
+
+At this point you have access to the Kubernetes cluster and can use it!
+If you don't have Kubernetes experience you can check out the [Kubernetes
+Basics official
+documentation](https://kubernetes.io/docs/tutorials/kubernetes-basics/deploy-app/deploy-intro/)
+to learn about its usage.
+
+**Note**: Lokomotive sets up a pretty restrictive Pod Security Policy that
+disallows running containers as root by default, check the [Pod Security Policy
+documentation](../concepts/securing-lokomotive-cluster.md#cluster-wide-pod-security-policy)
+for more details.
+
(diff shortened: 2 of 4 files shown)


docs/quickstarts/aws.md
```diff
@@ -163,6 +163,19 @@ export KUBECONFIG=./lokomotive-assets/cluster-assets/auth/kubeconfig
 kubectl get nodes
 ```
 
+## Using the cluster
```
invidian: Should we have some general quick start guide and leave platform specific documents for only platform specific information to avoid such duplication? I feel, that it could quickly get out of sync.

docs/quickstarts/aws.md
```diff
@@ -163,6 +163,19 @@ export KUBECONFIG=./lokomotive-assets/cluster-assets/auth/kubeconfig
...
+At this point you have access to the Kubernetes cluster and can use it!
+If you don't have Kubernetes experience you can check out the [Kubernetes
+Basics official
+documentation](https://kubernetes.io/docs/tutorials/kubernetes-basics/deploy-app/deploy-intro/)
+to learn about its usage.
+
+**Note**: Lokomotive sets up a pretty restrictive Pod Security Policy that
+disallows running containers as root by default, check the [Pod Security Policy
+documentation](../concepts/securing-lokomotive-cluster.md#cluster-wide-pod-security-policy)
+for more details.
```
invidian: We could mention, that many projects provide their own PSPs, which could be used, when deploying to increase the compatibility. e.g https://github.com/helm/charts/blob/master/stable/prometheus-operator/values.yaml#L72

docs/quickstarts/aws.md
iaguis: Yes, can you create an issue about that?

invidian: Done: #234.

docs/quickstarts/aws.md
iaguis: I added a note in the PSP documentation since adding it here sounded weird to me.

docs/concepts/securing-lokomotive-cluster.md
```diff
@@ -33,6 +33,9 @@ default PSPs for general purpose application usage:
     Allowed to workloads in `kube-system` namespace only. This PSP does not restrict any workloads
     from any required permissions.
 
+Many projects provide their own Pod Security Policies tailored to their needs which should be used
```
invidian: From security perspective, I'm not sure if saying `should` here is what we want. The policies we provide are secure and may work just fine for properly hardened workloads.

```suggestion
Many projects provide their own Pod Security Policies tailored to their needs which can be used
```

docs/concepts/securing-lokomotive-cluster.md
```diff
@@ -33,6 +33,9 @@ default PSPs for general purpose application usage:
...
+when deploying.
```
invidian:
```suggestion
when deploying, if provided policies are too strict.
```

PR #281, opened 2020-04-08
Update terraform-provider-ct to v0.5.0 and mention it in the docs

### LINKED ISSUE #223: Update terraform-provider-ct version
[v0.5.0 was released](https://github.com/poseidon/terraform-provider-ct/releases/tag/v0.5.0), we should update it.

In the docs we say terraform-provider-ct should be installed but we don't mention any version, we should consider mentioning it to avoid user problems when a new version is released and we haven't updated yet.

Reported-by: @jpetazzo

invidian:
Closes #223


invidian:
CI passes on the test pipeline with updated providers: https://yard.lokomotive-k8s.net/teams/main/pipelines/lokomotive-oss-aws-mateusz-testing/jobs/test-lokomotive-aws-deployment/builds/2.

docs/quickstarts/aws.md
```diff
@@ -27,7 +27,7 @@ By the end of this guide, you'll have a production-ready Kubernetes cluster runn
 * Basic understanding of Kubernetes concepts.
 * AWS account and IAM credentials.
 * AWS Route53 DNS Zone (registered Domain Name or delegated subdomain).
-* Terraform v0.12.x and [terraform-provider-ct](https://github.com/poseidon/terraform-provider-ct)
+* Terraform v0.12.x and [terraform-provider-ct v0.5.0](https://github.com/poseidon/terraform-provider-ct)
```
johananl, line 30: Nit: I would move the `v0.5.0` out of the hyperlink.

ci/set-terraform-version
```diff
@@ -1,2 +0,0 @@
-#!/bin/sh
-ln -fs ~/.local/bin/terraform12 ~/.local/bin/terraform
```
johananl, line 2: Why is this no longer needed? Would be nice to know the rationale.
Also, the CI code seems to still reference this script.

johananl: Actually looks like CI is failing because of this change.

invidian: As mentioned in the comment, CI changes needs to be applied first, so this PR passes the checks. I'll add some comment about removal here too, as it's done in CI code.

invidian: Done.

docs/quickstarts/aws.md
invidian: Done.

Issue #295, opened 2020-04-15
Document the control plane bootstrap process


johananl:
Lokomotive uses a self-hosted control plane based on Bootkube + Helm. The sequence of events which eventually results in an operational self-hosted k8s control plane is rather complex and contains multiple potential failure points. Therefore, it would be very useful to have document (possibly under the "concepts" category) which describes the various "layers" which comprise the control plane as well as the sequence of steps which take place from the moment a controller machine is created until the self-hosted control plane is converged.

I suggest we include the following in the document:

- A general explanation about what a self-hosted control plane is.
- A diagram showing the Lokomotive control plane architecture.
- A step-by-step explanation of a successful control plane bootstrap process.
- A troubleshooting section which explains what to look at in case the control plane on a cluster can't converge.
- Any "gotchas", e.g. special things to note during cluster failures, cluster upgrades etc.

invidian:
Note, if #297 gets merged, we should document it too.

PR #297, opened 2020-04-16
kube-apiserver: fix TLS handshake errors on Packet

invidian:

Refs #278


invidian:
Note: the documentation for that will be added as part of #295, as currently we don't have a good place to document it.

assets/lokomotive-kubernetes/bootkube/resources/charts/kube-apiserver/templates/kube-apiserver.yaml
```diff
@@ -239,8 +239,7 @@ spec:
 
           {{ end }}
           backend kube-apiserver
-            option httpchk GET /healthz HTTP/1.1\r\nHost:\ kube-apiserver\r\nAuthorization:\ Bearer\ {{ template "token" . }}
-            server 1 $ADDRESS:{{ template "port" . }} verify none check check-ssl
+            server 1 $ADDRESS:{{ template "port" . }}
```
iaguis, line 242: should we squash this with the previous commit? can we explain it a bit more (for example, why don't we have the `check` keyword now?)?

assets/lokomotive-kubernetes/bootkube/resources/charts/kube-apiserver/templates/kube-apiserver.yaml
```diff
@@ -221,9 +239,11 @@ spec:
 
           {{ end }}
           backend kube-apiserver
-            server 1 $ADDRESS:{{ template "port" . }} check
+            option httpchk GET /healthz HTTP/1.1\r\nHost:\ kube-apiserver\r\nAuthorization:\ Bearer\ {{ template "token" . }}
+            server 1 $ADDRESS:{{ template "port" . }} verify none check check-ssl
           EOF
           cat /run/kube-apiserver/haproxy.cfg
+          echo "{{ template "token" . }},nobody,nobody" > /run/kube-apiserver/token-auth-file
```
iaguis, line 245: I was trying to find info on this "nobody" user but couldn't find anything. Do you have a reference here? Or is it just a non-existent user and that means they don't have any privileges?

assets/lokomotive-kubernetes/bootkube/resources/charts/kube-apiserver/templates/kube-apiserver.yaml
```diff
@@ -14,6 +14,12 @@
 6443
 {{- end -}}
 {{ end }}
+{{- define "token" -}}
+{{ include (print $.Template.BasePath "/kube-apiserver-secret.yaml") . | sha256sum }}
```
iaguis, line 18: I assume this is the rendered file, right?

assets/lokomotive-kubernetes/bootkube/resources/charts/kube-apiserver/templates/kube-apiserver.yaml
invidian: The explanation is in the commit:

>As we always have only one backend, it doesn't make sense to perform
those checks.

Simply, HAProxy with checks can remove unhealthy backend from the list of backends, so the requests does not fail. In our case, we only ever have one backend, so we don't need to have health checks, as HAProxy won't return any cached or more specific errors either.

What I think we should have is the liveness probes for both containers, so they are restarted if they stop responding. But I think this is the topic for another issue.

I can actually swap the order of the commits, so we don't add it and then remove it back.

assets/lokomotive-kubernetes/bootkube/resources/charts/kube-apiserver/templates/kube-apiserver.yaml
invidian: It's just an arbitrary user name, which do not have any RBAC associated with. I wouldn't say it's "non-existent", as K8s users are not really stateful. If it has valid API token/cert, it appears in the system :smile: 

If you have better suggestions, I'm happy to adopt.

assets/lokomotive-kubernetes/bootkube/resources/charts/kube-apiserver/templates/kube-apiserver.yaml
invidian: Yes, this is already a template with values included. The pattern is from https://helm.sh/docs/howto/charts_tips_and_tricks/ "https://helm.sh/docs/howto/charts_tips_and_tricks/" section.

assets/lokomotive-kubernetes/bootkube/resources/charts/kube-apiserver/templates/kube-apiserver.yaml
iaguis: that's fine then

assets/lokomotive-kubernetes/bootkube/resources/charts/kube-apiserver/templates/kube-apiserver.yaml
iaguis: cool

assets/lokomotive-kubernetes/bootkube/resources/charts/kube-apiserver/templates/kube-apiserver.yaml
iaguis: ok, a verbose explanation like this comment would be better than the on in the commit :)

assets/lokomotive-kubernetes/bootkube/resources/charts/kube-apiserver/templates/kube-apiserver.yaml
invidian: I'll add an inline comment about it for the time being.

assets/lokomotive-kubernetes/bootkube/resources/charts/kube-apiserver/templates/kube-apiserver.yaml
invidian: Ditto inline comment.

assets/lokomotive-kubernetes/bootkube/resources/charts/kube-apiserver/templates/kube-apiserver.yaml
invidian: Added.

iaguis:
CI is broken.

PR #299, opened 2020-04-16
 packet: improve hardware reservations support using terraform dependencies 

rata:
This PR supersedes #119 and adds support for hardware reservations in Packet by adding auto-generated (i.e generated automatically by lokomotive) explicit terraform dependencies.

The need to use explicit dependencies on terraform (`depends_on`) is explained in the code itself, but you can have a look at these two links for more info (also linked in the code):
  https://github.com/terraform-providers/terraform-provider-packet/issues/176
  https://github.com/terraform-providers/terraform-provider-packet/pull/208

To guarantee that nodes with specific hw reservation UUID are created first, terraform dependencies are used and computed automatically in lokomotive. In contrast, PR #119 to achieve the same, did the following: `terraform apply -target=<nodes_to_create_first>` and then finished with `terraform apply` to create the rest of the resources.

The reasoning to change and rely on terraform for dependencies is simple: terraform already handles dependencies, handling them in different places is confusing and may hit several unexpected issues (as if running `terraform apply -target=<nodes_to_create_first>` we don't expect some resources to be created, but due to dependencies at the terraform layer they might be created if needed).

As explained in #119, I first started with the `terraform -target` approach because it was already used for DNS provider manual in packet, seemed simple and supporting all possible combinations of `reservation_ids` and `reservation_ids_default` seemed difficult otherwise. While working on #119, I decided that to support all combinations was not sane, and little by little, started to reduce the different combinations accepted. Basically, it was reduced and now `reservation_ids` and `reservation_ids_default` are mutually exclusive now. With that simplification in mind, going the terraform dependencies route seems simple. This PR just goes down that route, with this simplification.

There are some TODOs in the code, for things that I want to fully understand before merging, but other than that, this is pretty much functional.

## Pending things:
- [x] Add tests to new functions in this iteration and review other tests, just in case
- [x] Address "TODO" currently in the code 
- [ ] Create issue about plumbing to test terraform code (e.g. dependencies between devices, like we added here). Or should this be included here?


assets/lokomotive-kubernetes/packet/flatcar-linux/kubernetes/workers/workers.tf
```diff
@@ -67,6 +77,11 @@ data "ct_config" "ignitions" {
       setup_raid_hdd        = var.setup_raid_hdd
       setup_raid_ssd        = var.setup_raid_ssd
       setup_raid_ssd_fs     = var.setup_raid_ssd_fs
+
+      // This variable is not really used in the template, it is just to create
+      // a dependency from packet_device.nodes (that uses this a user_data input
+      // param) to the resources listed on var.depends_on_hack
+      //depends_on_hack = join(", ", var.depends_on_hack)
```
invidian, line 84: Everything commented out?

pkg/platform/packet/packet.go
```diff
@@ -76,6 +88,7 @@ type config struct {
 	ReservationIDs           map[string]string `hcl:"reservation_ids,optional"`
 	ReservationIDsDefault    string            `hcl:"reservation_ids_default,optional"`
 	CertsValidityPeriodHours int               `hcl:"certs_validity_period_hours,optional"`
+	NodesDependsOn           string            // Not exposed to the user
```
invidian, line 91: I wonder if this picks up the "default" label, the same as YAML and JSON, where you need to do explicitly `json:"-"` to skip it from marshaling/unmarshaling.

pkg/platform/packet/packet.go
```diff
@@ -237,6 +255,79 @@ func (c *config) terraformSmartApply(ex *terraform.Executor, dnsProvider dns.DNS
...
+//
+//
+// The explicit terraform dependency is needed to guarantees that nodes using
+// hardware reservation "next-available" won't use reservation IDS that another
+// worker pool may specify with a specific UUID, and thus fail to create the
+// node. This race condition is best explained here, if you want more info:
+// https://github.com/terraform-providers/terraform-provider-packet/issues/176
+// https://github.com/terraform-providers/terraform-provider-packet/pull/208
+func (c *config) terraformAddDependencies() {
+	// Nodes with specific hw reservation IDs
```
invidian, line 279: Comments should end with dots.
```suggestion
	// Nodes with specific hw reservation IDs.
```

pkg/platform/packet/packet.go
```diff
@@ -237,6 +255,79 @@ func (c *config) terraformSmartApply(ex *terraform.Executor, dnsProvider dns.DNS
...
+// 	c.NodesDependsOn = "[ module.worker-example.worker_nodes_ids ]"
+//
+//
+// The explicit terraform dependency is needed to guarantees that nodes using
+// hardware reservation "next-available" won't use reservation IDS that another
+// worker pool may specify with a specific UUID, and thus fail to create the
+// node. This race condition is best explained here, if you want more info:
+// https://github.com/terraform-providers/terraform-provider-packet/issues/176
+// https://github.com/terraform-providers/terraform-provider-packet/pull/208
+func (c *config) terraformAddDependencies() {
```
invidian, line 278: It would be nice to have some tests to validate the output of it.

pkg/platform/packet/packet.go
```diff
@@ -33,21 +35,31 @@ import (
...
+	NodeType              string            `hcl:"node_type,optional"`
+	Labels                string            `hcl:"labels,optional"`
+	Taints                string            `hcl:"taints,optional"`
+	ReservationIDs        map[string]string `hcl:"reservation_ids,optional"`
+	ReservationIDsDefault string            `hcl:"reservation_ids_default,optional"`
+	SetupRaid             bool              `hcl:"setup_raid,optional"`
+	SetupRaidHDD          bool              `hcl:"setup_raid_hdd,optional"`
+	SetupRaidSSD          bool              `hcl:"setup_raid_ssd,optional"`
+	SetupRaidSSDFS        bool              `hcl:"setup_raid_ssd_fs,optional"`
+	NodesDependsOn        string            // Not exposed to the user
```
invidian, line 62: I would make it `[]string` and then render it nicely in the template.

pkg/platform/packet/packet.go
```diff
@@ -237,6 +255,79 @@ func (c *config) terraformSmartApply(ex *terraform.Executor, dnsProvider dns.DNS
...
+func poolTarget(name, resource string) string {
+	return fmt.Sprintf("module.worker-%v.%v", name, resource)
+}
+
+// clusterTarget returns a string that can be used as "-target" argument to terraform.
+// For example:
+//	// target will be "module.packet-clusterName.ex".
+//	target := clusterTarget("clusterName", "ex")
+func clusterTarget(name, resource string) string {
+	return fmt.Sprintf("module.packet-%v.%v", name, resource)
```
invidian, line 328: We should get rid of that and simply name controller module `module.controller` or something, as it doesn't need to be unique in `lokoctl`.

invidian: I should perhaps create an issue about it.

assets/lokomotive-kubernetes/packet/flatcar-linux/kubernetes/workers/workers.tf
rata: Left over :see_no_evil:

pkg/platform/packet/packet.go
rata: Good point, thanks!

I was in the fence on that, using json marshall or this. But I think the current approach is the one I dislike the most, will try json or rendering (json doesn't seem so appealing because it will be in the struct itself, instead in other fields as all [other parts of the code do](https://github.com/kinvolk/lokomotive/blob/master/pkg/platform/packet/packet.go#L194-L196))

pkg/platform/packet/packet.go
rata: I think that even if we do that (probably we should), having a function to go from a cluster in go to a terraform target seems reasonable. It seems boilerplate code, otherwise (although not too much, so... I doubt a little bit) :)

assets/lokomotive-kubernetes/packet/flatcar-linux/kubernetes/workers/workers.tf
rata: Fixed locally, will be included in the next push :)

pkg/platform/packet/packet.go
rata: Not sure I follow. What is the issue?

pkg/platform/packet/packet.go
rata: Done :)

pkg/platform/packet/packet.go
rata: So, IMHO it is fine like this. If you think it doesn't need to be addressed here, please open an issue and tag me so we can discuss it there.

Re-open the discussion if this is not addressed (sorry, I think it is, but I might be missing something :))

pkg/platform/packet/packet.go
rata: Fixed, thanks!

pkg/platform/packet/packet.go
rata: Makes sense! Created a list of pending things in the PR description :)

rata: Marking as solved, to not double-track things :)

pkg/platform/packet/packet.go
invidian:
```go
package main

import (
  "encoding/json"
  "testing"
)

type F struct {
  Foo string
}

func TestMarshal(t *testing.T) {
  f := F{
    Foo: "bar",
  }

  b, err := json.Marshal(f)
  if err != nil {
    t.Fatalf("failed to encode: %v", err)
  }

  t.Logf("result: %+v", string(b))
}
```

I mean does something like this work for HCL. No major issue here, just a wondering if we should maybe add `hcl:"-"` to not unmarshalable struct fields.

pkg/platform/packet/packet.go
invidian: Thanks for the effort!

pkg/platform/packet/packet.go
invidian: Yeah, no worries. Created an issue for that: https://github.com/kinvolk/lokomotive/issues/308.

assets/lokomotive-kubernetes/packet/flatcar-linux/kubernetes/controllers.tf
```diff
@@ -59,6 +59,16 @@ resource "packet_device" "controllers" {
   ipxe_script_url = var.ipxe_script_url
   always_pxe      = false
   tags            = var.tags
+
+  # TODO: review this before merging
+  # depends_on needs a static list expression, so we can't do
+  #   depends_on = var.nodes_depends_on
+  # However, if we add "[]" to the var, it seems to be a static list expression
```
invidian, line 66: `depends_on` must always take a list of objects as far as I know. I guess pointing it to the list will make Terraform wait until all values inside it are "known". I guess this is OK.

pkg/platform/packet/packet.go
```diff
@@ -237,6 +255,100 @@ func (c *config) terraformSmartApply(ex *terraform.Executor, dnsProvider dns.DNS
...
+// worker pool may specify with a specific UUID, and thus fail to create the
+// node. This race condition is best explained here, if you want more info:
+// https://github.com/terraform-providers/terraform-provider-packet/issues/176
+// https://github.com/terraform-providers/terraform-provider-packet/pull/208
+func (c *config) terraformAddDependencies() {
+	// Nodes with specific hw reservation IDs.
+	nodesWithRes := make([]string, 0)
+
+	// Note that dependencies expressed in terraform are using the module
+	// ouput "controller_nodes_ids" for controllers and "worker_nodes_ids"
```
invidian, line 283:
```suggestion
	// output "controller_nodes_ids" for controllers and "worker_nodes_ids"
```

invidian, line 283: Hm, could we unify that and just have output parameter `device_ids` ?

pkg/platform/packet/packet.go
```diff
@@ -237,6 +255,100 @@ func (c *config) terraformSmartApply(ex *terraform.Executor, dnsProvider dns.DNS
...
+	// 	"nodes_depends_on" defined at ...:
+	//	all list elements must have the same type.
+	//
+	// Therefore, using the output of the resources ids, this issue is
+	// solved: all elements of the list (no matter if the dependency is on
+	// workers, controller or both) will always have the same type and work
+	// correctly, they are just resources ids (strings).
+	// Also, it makes nodes depend on nodes, that is the strict dependency
+	// that we really have, instead of depending in the whole module. So, it
+	// allows terraform to handle paralelization, and we only add
```
invidian, line 303:
```suggestion
	// allows Terraform to handle parallelization, and we only add
```

pkg/platform/packet/packet.go
```diff
@@ -237,6 +255,100 @@ func (c *config) terraformSmartApply(ex *terraform.Executor, dnsProvider dns.DNS
...
+	// Note that dependencies expressed in terraform are using the module
+	// ouput "controller_nodes_ids" for controllers and "worker_nodes_ids"
+	// for workers. And it is very important to keep it this way.
+	//
+	// If we modify it to depend only on the module, for example (just
+	// "module.packet" instead of "module.packet.controller_nodes_ids") it
+	// seems to work fine. However, it breaks if the dependency later
+	// becomes on the controller and another worker pool (e.g.
+	// [ "module.packet-cluster", "module.worker-1"]) as the resources aren't of
+	// the same *type*. In that case, terraform throws this error:
```
invidian, line 291:
```suggestion
	// the same *type*. In that case, Terraform throws this error:
```

pkg/platform/packet/packet.go
```diff
@@ -237,6 +255,100 @@ func (c *config) terraformSmartApply(ex *terraform.Executor, dnsProvider dns.DNS
...
+// hardware reservation "next-available" won't use reservation IDS that another
+// worker pool may specify with a specific UUID, and thus fail to create the
+// node. This race condition is best explained here, if you want more info:
+// https://github.com/terraform-providers/terraform-provider-packet/issues/176
+// https://github.com/terraform-providers/terraform-provider-packet/pull/208
+func (c *config) terraformAddDependencies() {
+	// Nodes with specific hw reservation IDs.
+	nodesWithRes := make([]string, 0)
+
+	// Note that dependencies expressed in terraform are using the module
```
invidian, line 282:
```suggestion
	// Note that dependencies expressed in Terraform are using the module
```

pkg/platform/packet/packet.go
```diff
@@ -237,6 +255,100 @@ func (c *config) terraformSmartApply(ex *terraform.Executor, dnsProvider dns.DNS
...
+// terraform targets needed to be created first. For example:
+//
+//	// Suppose worker pool "example" is the only to use a specific hw
+//	// reservation IDs. IOW, the controllers (c.NodesDependsOn) depends on
+//	// worker pool "example" to be created first.
+//	// Then, after calling this function, the attribute will be:
+// 	c.NodesDependsOn = "[ module.worker-example.worker_nodes_ids ]"
+//
+//
+// The explicit terraform dependency is needed to guarantees that nodes using
```
invidian, line 272:
```suggestion
// The explicit Terraform dependency is needed to guarantees that nodes using
```

pkg/platform/packet/packet.go
```diff
@@ -237,6 +255,100 @@ func (c *config) terraformSmartApply(ex *terraform.Executor, dnsProvider dns.DNS
 	return ex.Apply()
 }
 
+// terraformAddDependencies adds explicit dependencies to cluster nodes so nodes
+// with a specific hw reservation UUID are created before nodes that don't have
+// a specific hw reservation UUID.
+// The function modifies c.NodesDependsOn and c.workerPools[].NodesDependsOn,
+// assigning to them a string, formatted as a terraform list, of all worker
+// terraform targets needed to be created first. For example:
```
invidian, line 263:
```suggestion
// Terraform targets needed to be created first. For example:
```

pkg/platform/packet/packet.go
```diff
@@ -237,6 +255,100 @@ func (c *config) terraformSmartApply(ex *terraform.Executor, dnsProvider dns.DNS
 	return ex.Apply()
 }
 
+// terraformAddDependencies adds explicit dependencies to cluster nodes so nodes
+// with a specific hw reservation UUID are created before nodes that don't have
+// a specific hw reservation UUID.
+// The function modifies c.NodesDependsOn and c.workerPools[].NodesDependsOn,
+// assigning to them a string, formatted as a terraform list, of all worker
```
invidian, line 262:
```suggestion
// assigning to them a string, formatted as a Terraform list, of all worker
```

pkg/platform/packet/packet.go
```diff
@@ -237,6 +255,100 @@ func (c *config) terraformSmartApply(ex *terraform.Executor, dnsProvider dns.DNS
...
+	// solved: all elements of the list (no matter if the dependency is on
+	// workers, controller or both) will always have the same type and work
+	// correctly, they are just resources ids (strings).
+	// Also, it makes nodes depend on nodes, that is the strict dependency
+	// that we really have, instead of depending in the whole module. So, it
+	// allows terraform to handle paralelization, and we only add
+	// fine-grained dependencies.
+
+	if len(c.ReservationIDs) > 0 {
+		// Use a dummy tf output to wait on controllers nodes
```
invidian, line 307:
```suggestion
		// Use a dummy tf output to wait on controllers nodes.
```

pkg/platform/packet/packet.go
```diff
@@ -237,6 +255,100 @@ func (c *config) terraformSmartApply(ex *terraform.Executor, dnsProvider dns.DNS
...
+		tfTarget := clusterTarget(c.ClusterName, "controller_nodes_ids")
+		nodesWithRes = append(nodesWithRes, tfTarget)
+	}
+
+	for _, w := range c.WorkerPools {
+		if len(w.ReservationIDs) > 0 {
+			// Use a dummy tf output to wait on workers nodes
```
invidian, line 314:
```suggestion
			// Use a dummy tf output to wait on workers nodes.
```

assets/lokomotive-kubernetes/packet/flatcar-linux/kubernetes/controllers.tf
rata: Yes, but that doesn't explain the tricky part: `var.nodes_depends_on` is already a list. So, actually, this is a list of lists, just so terraform thinks it is a static list. What makes doubt here is that I strongly prefer to not use undocumented features (that may break in new releases)

pkg/platform/packet/packet.go
rata: Fixed, thanks!

rata: I went that way because workers already use the `worker_` prefix for its outputs (`worker_nodes_hostname` and friends), for consistency.

IMHO, it's the same. The string is shorter but clear (that is good), but won't allow to remove if branches anyways, so... I'm okay either way.

If we are going to use the same, though, I think I prefer `nodes_ids`, as it is even shorter, is a little bit clearer it refers to server instances and it's more consistent with `nodes_depends_on` naming. What do you think?

pkg/platform/packet/packet.go
rata: Thanks, fixed!

pkg/platform/packet/packet.go
rata: Fixed, thanks!

pkg/platform/packet/packet.go
rata: Fixed, thanks

pkg/platform/packet/packet.go
rata: Fixed! :)

pkg/platform/packet/packet.go
rata: Fixed!

pkg/platform/packet/packet.go
rata: Fixed, thanks!

pkg/platform/packet/packet.go
rata: Fixed, thanks!

pkg/platform/packet/packet.go
rata: Fixed

pkg/platform/packet/packet.go
rata: But why is this relevant? Do you think we may use it? Or it would be nice if the struct is marshall-able for some reason?

pkg/platform/packet/packet.go
invidian: I choose `device_ids`, because Packet resource is called `packet_device` and has `id` attribute, so `device_ids` seem natural. `node_ids` would also be okay I guess. I don't really like `nodes_ids` personally.

assets/lokomotive-kubernetes/packet/flatcar-linux/kubernetes/controllers.tf
rata: @invidian Do you understand why it workarounds the limitation and if it is a documented terraform feature or not?

pkg/platform/packet/packet.go
invidian: I posted incomplete example. The same works for `Marshal` and `Unmarshal`. Anyway..

pkg/platform/packet/packet.go
```diff
@@ -237,6 +255,102 @@ func (c *config) terraformSmartApply(ex *terraform.Executor, dnsProvider dns.DNS
...
+// a specific hw reservation UUID.
+// The function modifies c.NodesDependsOn and c.workerPools[].NodesDependsOn,
+// assigning to them a string, formatted as a Terraform list, of all worker
+// Terraform targets needed to be created first. For example:
+//
+//	// Suppose worker pool "example" is the only to use a specific hw
+//	// reservation IDs. IOW, the controllers (c.NodesDependsOn) depends on
+//	// worker pool "example" to be created first.
+//	// Then, after calling this function, the attribute will be:
+// 	c.NodesDependsOn = "[ module.worker-example.worker_nodes_ids ]"
```
invidian, line 269: This comment is no longer valid, as we changed the type to slice.

assets/lokomotive-kubernetes/packet/flatcar-linux/kubernetes/controllers.tf
invidian: I think this is a better reference link https://discuss.hashicorp.com/t/tips-howto-implement-module-depends-on-emulation/2305/2 and it seems to me, that this is working as designed (though not explicitly documented).

Honestly, I would call it a workaround. This is just how thing works in Terraform.

rata: Saw that too, will maybe update the link to that. Thanks!

I'll sleep on it (and maybe look a little bit terraform code to see the error, not sure) and probably just keep it. Thanks, it helps me to decide to keep it, really :)

pkg/platform/packet/packet.go
rata: Ouch! Sorry, I didn't ask for another review (just pushed some of the changes) because I knew some things might not be accurate and wanted to have a look myself. But thanks for catching it, I'm not sure if I would catch it (I usually need fresh eyes to catch those things :D)

assets/lokomotive-kubernetes/packet/flatcar-linux/kubernetes/controllers.tf
rata: Simplified the comment and kept it. Thanks =)

pkg/platform/packet/packet.go
rata: Okay, switched to `device_ids`.

Just curious, what is it that you don't like about the plural form (devices_ids or nodes_ids)?

pkg/platform/packet/packet.go
rata: Solved this and several other outdated comments :)

rata:
This is ready for review again, @invidian. PTAL :)

When it is ready to merge, I will create the issue mentioned in the PR

rata:
@invidian thanks for review and the patience while I tried different approaches! I think, however, time was well invested and the end result is quite better =)

pkg/platform/packet/packet.go
rata: Not sure what possible issue do you refer to, but hcl is loaded just fine :)

PR #334, opened 2020-04-23
Makefile: lint: Fallback to "origin/master"

rata:
The ".git/resource/base_sha" file exists in the concourse CI only. When
running it locally, the code fallbacks to "master". However, there is no
guarantee that the master branch in a local coputer is up to date.

This patch just changes the fallback to be "origin/master", as that has
more chances of being up to date. For example, I usually work in
branches and rarely update my master branch.

Using "origin/master" will also continue to work on users that update
the local master branch, as when they do so, they update the
"origin/master" reference too, usually.

The downside is that it assumes that the remote is named "origin". This
is a reasonable requirement for local development, however, as many
other project have them. If you want to name it differently, you can run
the linter manually too.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,7 +63,7 @@ test: run-unit-tests
 lint: build-slim build-test
 	# Note: Make sure that you run `git config diff.noprefix false` in this repo
 	# See this issue for more details: https://github.com/golangci/golangci-lint/issues/948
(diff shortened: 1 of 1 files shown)


rata:
Yeah, I agree (as mentioned in the commit). But it's very easy to fix and only affects you if you run the linter locally, and while we have the `--new-from-rev` flag, that should be removed because the whole codebase should be clean in the future. I don't really worry about it :)

PR #348, opened 2020-04-26
[RFC] Common platform config

schu:
This is an attempt to address #123 with minimal changes, i.e. a patch to share configuration attributes between platform packages with the goal of reducing duplicated code and make maintenance easier in the future.

The example demonstrates this with a single attribute only,

```
cluster_name = "{{.Config.ClusterName}}"
```

Moving all common configuration parameters to the shared struct and removing them from each individual platform package should make a difference in lines of codes.

Note: untested (modulo taking a quick look at the rendered terraform file) and there's probably room for improvement. But you get the idea.


johananl:
Thanks @schu! AFAIK there is actually ongoing work around this by @ipochi, so we could definitely consider the approach you're proposing.

rata:
I think the approach seems reasonable to me (I suggested embedding too in a design doc as something to explore :smile:). I haven't looked at the @ipochi PR yet :)

Maybe in another PR, but we can unify the things at the terraform layer too (like have a `variables_common.tf` and `variables.tf` on each platform, and the `_common.tf` one can be a symlink to file shared with variables for all platforms; the `variables.tf` the variables that are specific to the platform). I don't see why would we keep the duplication there, IMHO it makes sense to remove it and will make easier to reason in go code also what is common and what is not. It will also help to avoid issues in the future regarding feature parity.

ipochi:
Also sharing it here

https://github.com/kinvolk/lokomotive/issues/123#issuecomment-623386012

PR #353, opened 2020-04-27
kube-apiserver: render token-auth-file into secret

### LINKED ISSUE #347: Error: invalid authentication config: open /run/kube-apiserver/token-auth-file: no such file or directory
lokoctl version `v0.1.0-169-gef479dc9` (HEAD of master today).

Following the quickstart guide for AWS doesn't lead to a working cluster; the apiserver is not running:

```
[...]
Flag --insecure-port has been deprecated, This flag will be removed in a future version.
I0425 12:13:32.553216       1 server.go:656] external host was not specified, using 10.0.14.30
I0425 12:13:32.553634       1 server.go:153] Version: v1.18.2
Error: invalid authentication config: open /run/kube-apiserver/token-auth-file: no such file or directory
Usage:
  kube-apiserver [flags]
[...]
```

Possibly a regression from c0d340374a4bc0a92cf1a00ba3adb489d9b85ce9

/cc @invidian

invidian:
This PR fixes the regression added in c0d3403, by moving generation
of the token file from init container (which is only rendered for single
controller setups) to the kube-apiserver secret, which is available
regardless of number of controller nodes.

Closes #347

Also fixes some rendering nit of authorization header with unnecessary new line.


invidian:
>I haven't looked at the code yet, but IMHO that is a requirement to merge that and have at least a chance to not break it.

I created #363 so we can discuss that and IMO this is sufficient to merge this PR, as we should keep master branch stable.

rata:
@invidian added comment there. I don't think there is anything to discuss: if we want to support something, we want to test it. If it's CI, unit test, etc. we can discuss. But surely we want to test it.

If adding the test is not so easy for some reason (need more plumbing, need to release ASAP, etc.), we can do it in a different PR. If it is easy, we **really** should add it here. It's a good practice to fix regressions adding test for them.

assets/lokomotive-kubernetes/bootkube/resources/charts/kube-apiserver/templates/kube-apiserver.yaml
```diff
@@ -20,7 +20,7 @@
 {{- define "token" -}}
 {{ include (print $.Template.BasePath "/kube-apiserver-secret.yaml") . | sha256sum }}
 {{- end -}}
-{{- define "authHeader" }}
+{{- define "authHeader" -}}
```
rata, line 17: Can you please elaborate on the commit? Removing the trailing blank spaces why fixes it?

assets/lokomotive-kubernetes/bootkube/resources/charts/kube-apiserver/templates/kube-apiserver-secret.yaml
```diff
@@ -14,4 +8,23 @@ data:
...
+{{- end -}}
+# Value of "token" is composed by injecting all values into kube-apiserver-secret.yaml template and
+# then calculating sha256 sum of it, so it will be different for each cluster and additionally will be
+# changed every time certificate or secret template changes.
+{{- define "token" -}}
+{{ include "secrets" . | sha256sum }}
+{{- end }}
+{{- define "token-auth-file" -}}
+{{ template "token" . }},nobody,nobody
+{{- end }}
```
rata, line 21: Do you happen to have a link handy for these helm things?

assets/lokomotive-kubernetes/bootkube/resources/charts/kube-apiserver/templates/kube-apiserver.yaml
```diff
@@ -107,7 +101,7 @@ spec:
           --service-cluster-ip-range={{ .Values.apiserver.serviceCIDR }} \
           --tls-cert-file=/etc/kubernetes/secrets/apiserver.crt \
           --tls-private-key-file=/etc/kubernetes/secrets/apiserver.key \
-          --token-auth-file=/run/kube-apiserver/token-auth-file \
+          --token-auth-file=/etc/kubernetes/secrets/token-auth-file \
```
rata, line 104: I don't see where the secret is mounted... Is not mounted here? I see only this: https://github.com/kinvolk/lokomotive/pull/353/commits/dd41c95ae611e8e0cd819b225209af0181545b4d#diff-12be5db389a0fefe0038d10211a890b4R242-R249 and don't see this mounted.

Why is that the new path?

invidian: The secret is moved from the runtime directory to the secret object, which is always mounted.

From here: https://github.com/kinvolk/lokomotive/pull/353/files#diff-12be5db389a0fefe0038d10211a890b4L250

To here: https://github.com/kinvolk/lokomotive/pull/353/files#diff-0b161dbfa83ace6b4646c7465803f021R29

invidian: >Why is that the new path?

This is the new path, because this is where rest of the secrets is mounted.

assets/lokomotive-kubernetes/bootkube/resources/charts/kube-apiserver/templates/kube-apiserver-secret.yaml
invidian: https://helm.sh/docs/chart_template_guide/named_templates/

assets/lokomotive-kubernetes/bootkube/resources/charts/kube-apiserver/templates/kube-apiserver.yaml
invidian: Without this patch, the template content looks like following:
```

...
```

So when you inject it into the YAML, it looks like:
```yaml
foo: "
...
```

Instead of:
```yaml
foo: "..."
```

I'm not sure how to inject such details into commit messages. It's hard to describe, but easy to see if you look at the manifests. I should perhaps do a diff on rendered file :thinking:

invidian:
> If adding the test is not so easy for some reason (need more plumbing, need to release ASAP, etc.), we can do it in a different PR. If it is easy, we really should add it here. It's a good practice to fix regressions adding test for them at the same time.

It requires new pipeline with new configuration to be set up, I'd rather not do that in the scope of this. I wonder what other think about it.

assets/lokomotive-kubernetes/bootkube/resources/charts/kube-apiserver/templates/kube-apiserver.yaml
rata: You can just put this in the commit message, IMHO :-)

assets/lokomotive-kubernetes/bootkube/resources/charts/kube-apiserver/templates/kube-apiserver-secret.yaml
rata: Thanks!

assets/lokomotive-kubernetes/bootkube/resources/charts/kube-apiserver/templates/kube-apiserver-secret.yaml
```diff
@@ -14,4 +8,23 @@ data:
...
+  aggregation-client.key: "{{ .Values.apiserver.aggregationClientKey }}"
+{{- end -}}
+# Value of "token" is composed by injecting all values into kube-apiserver-secret.yaml template and
+# then calculating sha256 sum of it, so it will be different for each cluster and additionally will be
+# changed every time certificate or secret template changes.
+{{- define "token" -}}
+{{ include "secrets" . | sha256sum }}
+{{- end }}
+{{- define "token-auth-file" -}}
+{{ template "token" . }},nobody,nobody
```
rata, line 20: I still don't follow who uses the `,nobody,nobody` nor why it is needed now :-/

assets/lokomotive-kubernetes/bootkube/resources/charts/kube-apiserver/templates/kube-apiserver.yaml
rata: Ohh, okay, doing a local checkout of the branch I can find where it is. Thanks!

assets/lokomotive-kubernetes/bootkube/resources/charts/kube-apiserver/templates/kube-apiserver-secret.yaml
invidian: It was there before... This is the syntax of the token auth file: https://kubernetes.io/docs/reference/access-authn-authz/authentication/#static-token-file

assets/lokomotive-kubernetes/bootkube/resources/charts/kube-apiserver/templates/kube-apiserver.yaml
invidian: I added my comment to commit message, not sure if it's very readable, but it's something I guess :smile:

PR #354, opened 2020-04-27
Fix: Rendering of fields in AWS template.

ipochi:
- fix rendering error target_groups in worker pool.
- fix rendering error in clc_snippets in worker pool.
- fix rendering error in ssh_pub_keys in worker pool.

### DIFF
--- a/pkg/platform/aws/template.go
+++ b/pkg/platform/aws/template.go
@@ -118,7 +118,7 @@ module "worker-pool-{{ $index }}" {
   cluster_domain_suffix = "{{ $.Config.ClusterDomainSuffix }}"
   {{- end }}
 
-  ssh_keys              = {{ index (index $.WorkerpoolCfg $index) "ssh_pub_keys" }}
+  ssh_keys              = {{ (index $.WorkerpoolCfg $index "ssh_pub_keys") }}
   name                  = "{{ $pool.Name }}"
   worker_count          = "{{ $pool.Count}}"
   os_name               = "flatcar"
@@ -149,12 +149,13 @@ module "worker-pool-{{ $index }}" {
   {{- if $pool.SpotPrice }}
   spot_price            = "{{ $pool.SpotPrice }}"
   {{- end }}
-  {{- if $pool.TargetGroups }}
-  target_groups         = "{{ index (index $.WorkerpoolCfg $index) "target_groups" }}"
+
+	{{- if $pool.TargetGroups }}
+  target_groups         = {{ (index $.WorkerpoolCfg $index "target_groups") }}
   {{- end }}
 
-  {{- if ne (index (index $.WorkerpoolCfg $index) "clc_snippets") "null" }}
-  clc_snippets          = {{ index (index $.WorkerpoolCfg $index) "clc snippets" }}
+  {{- if $pool.CLCSnippets }}
+  clc_snippets          = {{ (index $.WorkerpoolCfg $index "clc_snippets") }}
   {{- end }}
 
   {{- if $pool.Tags }}


pkg/platform/aws/template.go
```diff
@@ -149,12 +149,13 @@ module "worker-pool-{{ $index }}" {
   {{- if $pool.SpotPrice }}
   spot_price            = "{{ $pool.SpotPrice }}"
   {{- end }}
-  {{- if $pool.TargetGroups }}
-  target_groups         = "{{ index (index $.WorkerpoolCfg $index) "target_groups" }}"
+
+	{{- if $pool.TargetGroups }}
```
invidian, line 153: Indentation?

ipochi: Rendering was not done.

```{{- if ne (index (index $.WorkerpoolCfg $index) "clc_snippets") "null" }}``` was not correct as for a worker pool that didnt have clc_snippets. 

the rendered cluster.tf was `clc_snippets = `

File: pkg/platform/aws/template.go [hunk 2]
```diff
@@ -118,7 +118,7 @@ module "worker-pool-{{ $index }}" {
   cluster_domain_suffix = "{{ $.Config.ClusterDomainSuffix }}"
   {{- end }}
 
-  ssh_keys              = {{ index (index $.WorkerpoolCfg $index) "ssh_pub_keys" }}
+  ssh_keys              = {{ (index $.WorkerpoolCfg $index "ssh_pub_keys") }}
```
rata, line 121: Why not use `$pool.SSHPubKeys`. I mean:
```suggestion
  ssh_keys              = {{ $pool.SSHPubKeys }}
```

pkg/platform/aws/template.go
```diff
@@ -149,12 +149,13 @@ module "worker-pool-{{ $index }}" {
...
-  target_groups         = "{{ index (index $.WorkerpoolCfg $index) "target_groups" }}"
+
+	{{- if $pool.TargetGroups }}
+  target_groups         = {{ (index $.WorkerpoolCfg $index "target_groups") }}
   {{- end }}
 
-  {{- if ne (index (index $.WorkerpoolCfg $index) "clc_snippets") "null" }}
-  clc_snippets          = {{ index (index $.WorkerpoolCfg $index) "clc snippets" }}
+  {{- if $pool.CLCSnippets }}
+  clc_snippets          = {{ (index $.WorkerpoolCfg $index "clc_snippets") }}
```
rata, line 158: Idem other comment, why this can just be `$pool.CLCSnippets`?

pkg/platform/aws/template.go
rata: This can't work because that is a slice, we woud need like:
```suggestion
    ssh_keys = [
      {{- range $ke := $pool.SSHPubKeys }}
      {{ $key }} = "{{ $value }}"
      {{- end }}
    ]
```

pkg/platform/aws/template.go
rata: Same can be done here

ipochi:
> Oh, I see why we need to do the ugly index trick now. Commented with a more readable alternative that works now, and can remove the need for the json.Marshall all over the place, that only complicates the template rendering later :-/
> 
> However, this is yet another reason why we should do something like [I commented on the other PR](https://github.com/kinvolk/lokomotive/pull/355#discussion_r416812854) for this. The mentioned suggestion shows how to do it.
> 
> However, I'm okay if you want a quick fix for the syntax now, as long as we improve it to make it readable in the other PR (specially if this helps you to avoid conflicts, etc.). If we are not doing it in the [other PR](https://github.com/kinvolk/lokomotive/pull/355), I'd request the changes here. That needs to be improved, specially if it's broken and we are fixing it :)

@rata I'll address your concerns regarding this in the other PR #355

Issue #349, opened 2020-04-26
AWS: ingress not working by default


schu:
Worker pool and NLB target group configuration doesn't match in a default Lokomotive setup on AWS and ingress doesn't work out of the box.

There are two problems:

* The NLB target groups for port 80 and 443 are empty after installation. (`<clustername>-workers-http`/`<clustername>-workers-https`). The worker pool target groups (`<poolname>-workers-http`/`<poolname>-workers-https`) on the other hand are not configured with the NLB.

* The NLB target group's health check is configured to check target nodes for health on port `:10254/healthz`, but that port is not actually used / listend on anymore.

Manually adding the worker nodes to the NLB target groups and changing the health check configuration to theck port `80`/`443` (the target ports) instead is a workaround.

johananl:
Fixed in https://github.com/kinvolk/lokomotive/pull/367.

Issue #366, opened 2020-04-27
Handle AWS ELB target group names > 32 characters


johananl:
ELB target groups are limited to 32 characters on the AWS API. Since we interpolate the cluster name and worker pool name into the target group, long cluster/pool names may exceed this limit.

At the moment there is no check in place for this limit - we fail the deployment with a not-so-nice Terraform error. Ideally we should find a way to never hit this limit. An alternative would be to validate the config and fail fast if necessary.

johananl:
Handled in https://github.com/kinvolk/lokomotive/pull/367. We simply let Terraform generate a random target group name and put the meaningful info (cluster + worker pool name) in AWS tags.

invidian:
See also #282.

johananl:
#367 is merged. Closing.

PR #367, opened 2020-04-27
Fix AWS worker pool handling

johananl:
## General

Fixes https://github.com/kinvolk/lokomotive/issues/349.

This PR fixes the handling of worker pools on AWS. Following is a summary of the changes:

- There used to be an "invisible" worker pool of size 0 that was created with every cluster. This pool is now removed.
- NLB listeners are now created per worker pool and are correctly wired to the right target groups. Before this change there used to be one "global" set of listeners which were wired to the empty worker pool, practically blackholing all ingress traffic to the cluster.
- NLB listener ports for HTTP/HTTPS are now configurable with a default of `80/443`, respectively. This is required in order to support an arbitrary number of worker pools with ingress using the same NLB.
- Target group names have been cleaned. These are limited to 32 characters by the AWS API so we want to keep them short. Related issue: https://github.com/kinvolk/lokomotive/issues/366

## Testing

To test this PR, deploy a cluster with a config similar to the following

```hcl
cluster "aws" {
  asset_dir        = "./assets"
  cluster_name     = "my-cluster"
  controller_count = 1
  dns_zone         = "example.com"
  dns_zone_id      = "xxxxxxxx"
  ssh_pubkeys      = ["ssh-rsa ..."]

  worker_pool "pool-1" {
    instance_type = "t3a.small"
    count         = 2
    ssh_pubkeys   = ["ssh-rsa ..."]
  }
}

component "contour" {
  service_type = "NodePort"
}
component "cert-manager" {
  email = "me@example.com"
}
component "httpbin" {
  ingress_host = "httpbin.example.com"
}
```

httpbin should be available at https://httpbin.example.com.


johananl:
We're hitting the size limit of the ELB target group name (32 chars).

johananl:
> We're hitting the size limit of the ELB target group name (32 chars).

Fixed.

assets/lokomotive-kubernetes/aws/flatcar-linux/kubernetes/workers/variables.tf
```diff
@@ -1,4 +1,9 @@
-variable "name" {
```
invidian, line 1: If we don't care about Terraform API, we could leave it like it is and combine the name one level higher.

assets/lokomotive-kubernetes/aws/flatcar-linux/kubernetes/workers/ingress.tf
```diff
@@ -1,7 +1,7 @@
 # Target groups of instances for use with load balancers
 
 resource "aws_lb_target_group" "workers-http" {
-  name        = "${var.name}-workers-http"
+  name        = "${var.cluster_name}-${var.pool_name}-workers-http"
```
invidian: Maybe it would make sense to remove `-workers` from it, so we save some characters? `-workers-http` is already 13 characters, + one `-`, it's 14. This leaves us only 17 characters for cluster name + pool name, not much in my opinion.

assets/lokomotive-kubernetes/aws/flatcar-linux/kubernetes/nlb.tf
```diff
@@ -36,30 +36,6 @@ resource "aws_lb_listener" "apiserver-https" {
   }
 }
 
-# Forward HTTP ingress traffic to workers
```
invidian, line 39: There is a typo in commit message: `protocl`.

docs/configuration-reference/platforms/aws.md
```diff
@@ -208,6 +214,8 @@ worker_pool "my-worker-pool" {
 | `worker_pool.disk_iops`       | IOPS of the EBS volume (e.g 100).                                                                                                                                                          | 0               | false    |
 | `worker_pool.spot_price`      | Spot price in USD for autoscaling group spot instances. Leave as empty string for autoscaling group to use on-demand instances. Switching in-place from spot to on-demand is not possible. | ""              | false    |
 | `worker_pool.target_groups`   | Additional target group ARNs to which worker instances should be added.                                                                                                                    | []              | false    |
+| `worker_pool.lb_http_port`    | Port the load balancer should listen on for HTTP connections                                                                                                                               | 80              | false    |
```
invidian, line 217:
```suggestion
| `worker_pool.lb_http_port`    | Port the load balancer should listen on for HTTP connections.                                                                                                                            | 80              | false    |
```

docs/configuration-reference/platforms/aws.md
```diff
@@ -208,6 +214,8 @@ worker_pool "my-worker-pool" {
...
+| `worker_pool.lb_https_port`    | Port the load balancer should listen on for HTTPS connections                                                                                                                             | 443             | false    |
```
invidian, line 218:
```suggestion
| `worker_pool.lb_https_port`    | Port the load balancer should listen on for HTTPS connections.                                                                                                                       | 443             | false    |
```

ci/aws/aws-cluster.lokocfg.envsubst
```diff
@@ -15,7 +15,7 @@ cluster "aws" {
   os_channel   = var.os_channel
   ssh_pubkeys  = ["$PUB_KEY"]
 
-  worker_pool "$CLUSTER_ID-wp" {
+  worker_pool "wp" {
```
surajssd, line 18: Won't this cause conflict because there will be multiple CI piplelines creating same worker pool ?

surajssd:
> `httpbin` should be available at https://httpbin.example.com.

@johananl I was manually testing this PR, where should `httpbin.example.com` be pointing to in Route53? This is a manual step right?

johananl:
> > `httpbin` should be available at https://httpbin.example.com.
> 
> @johananl I was manually testing this PR, where should `httpbin.example.com` be pointing to in Route53? This is a manual step right?

Lokomotive creates a wildcard record which routes `*.example.com` to the NLB. You shouldn't need to touch Route 53 manually. My testing instructions should work exactly as written.

ci/aws/aws-cluster.lokocfg.envsubst
johananl: No, because the worker pool name gets prefixed with the cluster name.

assets/lokomotive-kubernetes/aws/flatcar-linux/kubernetes/nlb.tf
johananl: Thanks, fixed.

assets/lokomotive-kubernetes/aws/flatcar-linux/kubernetes/workers/variables.tf
johananl: We use `cluster_name` to tag the ELB target groups we create inside the `workers` module. Doing what you're suggesting would make us lose this information.

docs/configuration-reference/platforms/aws.md
johananl: Thanks, fixed.

docs/configuration-reference/platforms/aws.md
johananl: Thanks, fixed.

invidian:
>Lokomotive creates a wildcard record which routes *.example.com to the NLB. You shouldn't need to touch Route 53 manually. My testing instructions should work exactly as written.

@johananl if this should work fully automatically, perhaps we should add e2e test for it? Which just does the HTTP request which should succeed?

johananl:
> > Lokomotive creates a wildcard record which routes *.example.com to the NLB. You shouldn't need to touch Route 53 manually. My testing instructions should work exactly as written.
> 
> @johananl if this should work fully automatically, perhaps we should add e2e test for it? Which just does the HTTP request which should succeed?

I can look into that, although I wanted to keep the scope of this PR small and merge it ASAP as things are currently broken.

surajssd:
> I can look into that, although I wanted to keep the scope of this PR small and merge it ASAP as things are currently broken.

This works for me. Only one difference from deploying on packet is that on packet we need to create a manual entry in route53 and here user is free to choose whatever URL they want.

On AWS we need to make sure that your the URL is `<app-name>.<cluster-name>.<dns-zone>`.

johananl:
>Only one difference from deploying on packet is that on packet we need to create a manual entry in route53 and here user is free to choose whatever URL they want.

Opened https://github.com/kinvolk/lokomotive/issues/370 to track this.

>On AWS we need to make sure that your the URL is `<app-name>.<cluster-name>.<dns-zone>.`

Opened https://github.com/kinvolk/lokomotive/issues/371 to track this.

johananl:
I've added an experimental ingress integration test. Let's see if it passes CI (I'm not sure how the ordering/dependencies of the e2e tests work).

test/ingress/aws/aws_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+	defer func() {
+		if err := resp.Body.Close(); err != nil {
+			t.Logf("closing HTTP response body: %v", err)
+		}
+	}()
+
+	if resp.StatusCode != http.StatusOK {
+		t.Fatalf("got an error HTTP response: %d", resp.StatusCode)
+	}
```
surajssd, line 52: This might fail for various reasons like the pod is not yet started, TLS cert not generated, etc. I would recommend adding some retry logic here.

Using `wait.PollImmediate` is the easiest way to add retry. And we use it everywhere in the code. 

https://github.com/kinvolk/lokomotive/blob/2f92fbe8789dcb49d707a233fe703f017be7721d/test/components/util/util.go#L133

surajssd:
@johananl 

> (I'm not sure how the ordering/dependencies of the e2e tests work).

The ordering won't be an issue here because you are invoking those tests after components have been deployed.

test/ingress/aws/aws_test.go
johananl: Yes, I've considered this of course. I wanted to optimistically start with the simplest code possible and add complexity as things break. But fair enough, I'll add some retry logic.

johananl: Well, this failed on the first CI run, so :laughing: :shrug:

johananl:
Added a simple integration test for ingress traffic.

johananl:
Any remaining concerns?

johananl:
I've just discovered traffic seems to work although there is a problem with the NLB target groups. I'm checking this.

johananl:
I've replaced exposing Envoy using `hostPort` with `NodePort` on AWS (and left the existing `LoadBalancer` on Packet, which works fine without the `hostPort` setting.

I've tested the change manually on both AWS and Packet and things LGTM.

invidian:
Rebasing needed.

johananl:
Failed CI builds are due to Packet provisioning errors. I've tested the latest status manually on Packet and AWS and things seem to work fine.

pkg/components/contour/component.go
```diff
@@ -28,7 +28,11 @@ import (
 	"github.com/kinvolk/lokomotive/pkg/util/walkers"
 )
 
-const name = "contour"
+const (
+	name                    = "contour"
+	serviceTypeNodePort     = "NodePort"
```
invidian, line 33: Should we use https://godoc.org/k8s.io/api/core/v1#ServiceType instead?

pkg/components/contour/component.go
```diff
@@ -80,6 +83,16 @@ func (c *component) RenderManifests() (map[string]string, error) {
 		return nil, errors.Wrap(err, "failed to walk assets")
 	}
 
+	// Set default service type to 'LoadBalancer'.
+	if c.ServiceType == "" {
```
invidian, line 87: Why not in `newComponent`, so it gets overwritten by HCL if the field is set?

pkg/components/contour/component.go
johananl: This entails adding a dependency, which seems more complex to me.
What is you rationale for using the constants from the k8s package?

pkg/components/contour/component.go
johananl: Didn't realize we could do that. Fixed.
NOTE: This kind of patterns we can (and probably should) enforce for all components :shrug:

pkg/components/contour/component.go
invidian: It's less duplication to me and it better reflects what the property represents. Our project depend on K8s anyway, so it's not adding a dependency to the project. But I don't insist on that.

johananl: Right, it's adding a dependency to the package.

johananl:
> As far as I understand, on AWS we will now prefer NodePort type over LoadBalancer. This is because we don't have support for creating load balancers using CCM on AWS right now, correct?

It's part that, and part me wanting to reuse the existing NLB as much as possible. NLBs are highly scalable and very fast as they are only doing layer3/4-based routing, so in theory there shouldn't be a problem to use one NLB for both management access and ingress traffic, assuming the ports don't collide.

>Once we add that, NLB-related code can be removed from Terraform, correct?

Well, it *could* be removed, but I don't see why we would want to remove the NLB.

test/ingress/aws/aws_test.go
```diff
@@ -0,0 +1,69 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// +build aws
+// +build e2e
+
+package aws
```
rata, line 18: Sorry for the silly question, but I fail to see where this is used. Maybe in the ci project will be used once it is merged? No idea :D

assets/lokomotive-kubernetes/aws/flatcar-linux/kubernetes/variables.tf
```diff
@@ -29,12 +29,6 @@ variable "controller_count" {
   description = "Number of controllers (i.e. masters)"
 }
 
-variable "worker_count" {
```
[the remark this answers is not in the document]
johananl: If by "we" you mean "Johannes" - yes, I can open an issue for that :-)

test/ingress/aws/aws_test.go
```diff
@@ -0,0 +1,69 @@
...
+
+	i, err := client.ExtensionsV1beta1().Ingresses("httpbin").Get("httpbin", metav1.GetOptions{})
+	if err != nil {
+		t.Fatalf("getting httpbin ingress: %v", err)
+	}
+
+	h := i.Spec.Rules[0].Host
+
+	err = retryutil.Retry(retryIntervalSeconds*time.Second, maxRetries, func() (bool, error) {
+		resp, err := http.Get(fmt.Sprintf("https://%s/get", h))
```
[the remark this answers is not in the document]
johananl: My rationale was to cover as much critical parts as I can with this simple test. I can definitely split it into an HTTP and an HTTPS test if that makes more sense to you.

To emphasize, increasing test coverage is not the scope of this PR. I've added this test following @invidian's request and would like to avoid delaying the merge further if we can since **things are currently broken on master**.

test/ingress/aws/aws_test.go
johananl: What do you mean by "this"?

test/ingress/aws/aws_test.go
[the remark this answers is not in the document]
rata: I'm totally fine with this test and the approach in general. e2e test don't need to be very specific, we can debug de failure when introduced :)

test/ingress/aws/aws_test.go
rata: Exaclty that. Great, then, this is even added to the ci! Awesome! :) (sorry, not familiar with the CI still :see_no_evil: )

ipochi:
@rata @johananl 

I assume you are talking about `LoadConfig()` in `pkg/config/config.go`

In PR #388 , `LoadConfig` has been refactored `LoadHCLFiles` and `ParseHCLFiles` and unit tests have been added for the same in `config_test.go`
https://github.com/kinvolk/lokomotive/blob/imran/refactor-1/pkg/config/config.go#L95-#L231

Regarding Validation of the configuration: 

I have this idea 

The code `pkg/config/config.go` deals with loading and parsing of the HCL configuration without the knowledge of the underlying structure.

I have in mind to provide `LokomotiveConfig` struct that acts a central place for the entire Lokomotive configuration and a method of this struct would validate the configuration. 

Each interface(Platform, Component and Backend) would add a method `Validate()` to the interface.
`LoadConfig` in each of the three interfaces should only load the config or be removed and any other logic (validation) should be refactored into Validate().

Something like this: 
```
type LokomotiveConfig struct {
  // Backend
  Backend backend.Backend
  Platform platform.Platform
  Components []components.Component
}

func (lc *LokomotiveConfig) Validate() hcl.Diagnostics {
 platform.Validate()
 backend.Validate()
 components.Validate()
}
```

Furthermore the goal is   to remove the Lokomotive management code from `cli/cmd` with the below struct implementing the `Manager` interface.

```
type Manager interface {
  Apply()
  Destroy()
  ApplyComponents()
  DestroyComponents()
  Health()
}
```
``` 
type lokomotive struct {
  Platform platform.Platform
  Config *LokomotiveConfig
  Executor *terraform.Executor
  Logger *logrus.Entry
} 
```

Please note this is part of the refactoring that I am trying to do and I have it in one of the PRs.

assets/lokomotive-kubernetes/aws/flatcar-linux/kubernetes/variables.tf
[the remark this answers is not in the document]
johananl: https://github.com/kinvolk/lokomotive/issues/396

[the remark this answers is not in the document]
rata: Thanks!

rata:
@ipochi thanks! But no, I don't mean that function. I mean this: https://github.com/kinvolk/lokomotive/pull/367/files#diff-7a1275af7081a8b0912aeae9a1d2f21cL62 (that is here `pkg/components/contour/component.go`). cc @johananl in case it wasn't clear, I meant that function :)

Regarding the changes that you mention, can we see them in the PR you mention and consider them out of scope for this PR? Or am I missing something?

pkg/components/contour/component.go
```diff
@@ -80,6 +85,11 @@ func (c *component) RenderManifests() (map[string]string, error) {
 		return nil, errors.Wrap(err, "failed to walk assets")
 	}
 
+	// Validate service type.
+	if c.ServiceType != serviceTypeNodePort && c.ServiceType != serviceTypeLoadBalancer {
+		return nil, fmt.Errorf("unknown service type %q", c.ServiceType)
```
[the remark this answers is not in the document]
rata: As mentioned in the comments, I meant moving it to [this function](https://github.com/kinvolk/lokomotive/pull/367/files#diff-7a1275af7081a8b0912aeae9a1d2f21cL62)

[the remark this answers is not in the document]
johananl: Fixed. Thanks for the pointer - it looks much better now.
I've also added a test and refactored the existing tests a bit to make them cleaner.

test/ingress/aws/aws_test.go
```diff
@@ -0,0 +1,69 @@
...
+
+const (
+	retryIntervalSeconds = 5
+	maxRetries           = 60
+)
+
+func TestAWSIngress(t *testing.T) {
+	client := testutil.CreateKubeClient(t)
+
+	i, err := client.ExtensionsV1beta1().Ingresses("httpbin").Get("httpbin", metav1.GetOptions{})
```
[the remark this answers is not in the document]
johananl: Fixed.

johananl:
Thanks @rata and @ipochi. I've understood the intention and fixed accordingly.

pkg/components/contour/component_test.go
```diff
@@ -22,41 +22,72 @@ import (
...
+	}
+
+	for _, tc := range tests {
+		b, d := util.GetComponentBody(tc.hcl, name)
+		if d != nil {
+			t.Fatalf("Error getting component body: %v", d)
+		}
+
+		c := newComponent()
+		d = c.LoadConfig(b, &hcl.EvalContext{})
```
rata, line 74: nit: Can't you just use `nil` instead of `&hcl.EvalContext{}`? Checking the docs, it [seems nil will do the same](https://pkg.go.dev/github.com/hashicorp/hcl2/gohcl?tab=doc#DecodeBody):

>  The given EvalContext is used to resolve any variables or functions in expressions encountered while decoding. This may be nil to require only constant values, for simple applications that do not support variables or functions.

I think both are fine, though. This is just nit-picking and my personal taste, no big deal if you prefer it this way for some reason.

johananl:
Fixed all concerns. CI is green. Need another LGTM before I can merge.

pkg/components/contour/component_test.go
```diff
@@ -22,41 +22,72 @@ import (
...
+
+		if !tc.wantErr && d.HasErrors() {
+			t.Fatalf("Valid config should not return error, got: %s", d)
```
rata, line 77: Why `Fatalf` instead of `Errorf`? There is no point if one error is found in one test case, to run the other test cases too? (With error, it will fail but continue to run the others)

IMHO, it is useful to run the other testcases too. In that case, just `Errorf()` seems better, or if you want to abort execution of the current test too, `Errorf()` and `continue` (to continue with the next item in the loop). What do you think?

iaguis:
This looks very close to being merged. Can you rebase and address the small comments?

pkg/components/contour/component_test.go
invidian: @rata what I think should happen is that each case should run in `t.Run()` scope, so then use of `t.Fatal` is legitimate.

pkg/components/contour/component.go
```diff
@@ -38,38 +42,50 @@ func init() {
...
+
 	if configBody == nil {
 		return hcl.Diagnostics{
 			components.HCLDiagConfigBodyNil,
 		}
 	}
+
 	if err := gohcl.DecodeBody(*configBody, evalContext, c); err != nil {
-		return err
+		diagnostics = append(diagnostics, err...)
```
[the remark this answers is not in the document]
johananl: Makes sense, but I've copied this logic as-is from existing components. We need to fix this also in the `cluster-autoscaler` and `velero` components:

https://github.com/kinvolk/lokomotive/blob/ede5d2ad41b83a09c1e857dccabf924acba4bb7c/pkg/components/cluster-autoscaler/component.go#L232-L234

https://github.com/kinvolk/lokomotive/blob/ede5d2ad41b83a09c1e857dccabf924acba4bb7c/pkg/components/velero/velero.go#L131-L133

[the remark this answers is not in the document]
johananl: Opened https://github.com/kinvolk/lokomotive/issues/413 to track fixing the other components.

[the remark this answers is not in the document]
johananl: Actually `gohcl.DecodeBody` returns an `hcl.Diagnostics`, not an error. The returned value should be checked using the `HasErrors()` method, not compared to nil.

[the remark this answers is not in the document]
johananl: I've fixed the error check and added a return.

Passing a malformed HCL does fail as expected, however in order to add a test for this we would have to refactor the way we do component testing, which is way out of scope for this PR. The reason is - when adding a unit test with malformed HCL, `util.GetComponentBody()` fails before we have a chance to check whether we wanted to get an error or not.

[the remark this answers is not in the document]
johananl: @rata can we treat this as resolved? I want to remind that we have a broken state on master and I don't want to delay merging this any further.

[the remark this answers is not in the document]
johananl: Actually the return I've added doesn't do anything because we error out somewhere in `gohcl.DecodeBody()`. My error checking code never gets executed.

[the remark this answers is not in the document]
johananl: I'm completely confused now. If anybody knows the proper way to handle this - please let me know. I am currently reading through the parsing logic in the HCL library :disappointed:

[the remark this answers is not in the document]
invidian: @johananl if you try passing valid HCL buth with wrong structure (e.g. non existent attribute), then I think you should be able to trigger it. We have some tests, which handles that. E.g. https://github.com/kinvolk/lokomotive/blob/master/pkg/components/external-dns/component_test.go#L94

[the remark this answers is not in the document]
rata: @johananl Sure, mark as resolved when you consider it solved :)

pkg/components/contour/component_test.go
rata: I don't think it makes sense to use `t.Parallel()`, it has some side-effects and this probably takes some ms to run.

In any case, this is nit-picking IMHO. I'm fine with this being as it is too.

rata:
@johananl This LGTM, as I said in the previous reviews, I don't think I need to review again, the changes that might receive this PR are very small.
 
Regarding the open comment, as I mentioned, mark it solved when you consider it solved (AFAICS you found problems to test your solution, I think whatever you consider is reasonable at this point, is fine). Regarding the other comments, they are nit-picking so whatever you decide (even leave it as it is), is fine by me :)

If you need, of course, just ping me for a LGTM :)

pkg/components/contour/component.go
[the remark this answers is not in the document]
johananl: Thanks @invidian, that was indeed the case. I feel like we need better guidelines (and in my case - understanding) regarding how to deal with HCL errors. This still feels rather messy to me.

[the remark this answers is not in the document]
invidian: It is messy indeed. @ipochi refactioring work actually encapsulates all HCL code really nicely.

pkg/components/contour/component_test.go
johananl: Good point @rata. This matters because I've switched to table-driven tests, which means all test cases for Contour are executed in a single function. To clarify, `t.Fatal` only stops any additional tests for the Contour component. Subsequent test **functions** continue executing as expected.

johananl: Changed to `t.Errorf`.

pkg/components/contour/component_test.go
johananl: This, too, was copied from an existing component (cluster-autoscaler). Feel free to open an issue to handle other occurrences.

johananl:
All feedback was handled. Please check again.

assets/components/contour/values.yaml
```diff
@@ -10,6 +10,7 @@ contour:
 envoy:
   image: docker.io/envoyproxy/envoy
   tag: v1.13.1
+  serviceType:
```
rata, line 13: just curious (helm ignorant here): is this empty to make it mandatory or something?

johananl: Kind of. It's empty because we're setting the default in the component's Go code and I didn't want to set the default a second time in the values file. Regarding the Helm mechanism, AFAIK you can leave a key in `values.yaml` without a value and it's valid. If no value is provided at run time (via any of the supported Helm mechanisms), I guess you'll get an error when rendering the chart.

iaguis:
I've rebased and fixed conflicts.

assets/components/contour/values.yaml
rata: Oh, okay. Thanks for the explanation! :)

PR #390, opened 2020-05-05
ci/aws: test HA control plane on the CI

iaguis:
Since this is the production recommended setup let's test it on every
PR.

We have some coverage for the non-HA control plane because we test it in
the Flatcar Edge pipeline.

### DIFF
--- a/ci/aws/aws-cluster.lokocfg.envsubst
+++ b/ci/aws/aws-cluster.lokocfg.envsubst
@@ -8,12 +8,13 @@ variable "os_channel" {
 }
 
 cluster "aws" {
-  asset_dir    = pathexpand(var.asset_dir)
-  cluster_name = "$CLUSTER_ID"
-  dns_zone     = "$AWS_DNS_ZONE"
-  dns_zone_id  = "$AWS_DNS_ZONE_ID"
-  os_channel   = var.os_channel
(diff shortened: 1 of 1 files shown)


iaguis:
This should fail due to #347

ci/aws/aws-cluster.lokocfg.envsubst
```diff
@@ -8,12 +8,13 @@ variable "os_channel" {
...
 cluster "aws" {
-  asset_dir    = pathexpand(var.asset_dir)
-  cluster_name = "$CLUSTER_ID"
-  dns_zone     = "$AWS_DNS_ZONE"
-  dns_zone_id  = "$AWS_DNS_ZONE_ID"
-  os_channel   = var.os_channel
-  ssh_pubkeys  = ["$PUB_KEY"]
+  asset_dir        = pathexpand(var.asset_dir)
+  cluster_name     = "$CLUSTER_ID"
+  controller_count = 3
```
invidian, line 13: We could also use just 2.

iaguis: right, that should be enough, I'll change.

invidian:
Thanks for the PR. I pulled it into #347 to see if it works, as we discussed.

iaguis:
Let's close this since it's included in #347 now.

PR #409, opened 2020-05-11
bootkube: use host's /etc/hosts entries

### LINKED ISSUE #408: Change bootkube rkt container to use host's /etc/hosts entries
Currently, bare metal environment requires resolvable DNS names for API server and etcd servers. For user without running DNS server locally or for testing, it would be nice to be able have some workaround for it. One would be to populate `/etc/hosts` of nodes with required entries.

This is almost possible, using CLC snippets for controller nodes, except the bootkube, which runs in `rkt` container and have it's own `/etc/hosts` file, so health check on static kube-apiserver pod never succeeds. To make it work, `rkt` parameter `--hosts-entry=host` can be added.

I suggest we add `--hosts-entry=host` to bootkube, as potential breakage impact is minimal and this will save as from exposing general `bootkube_rkt_extra_args` in the Terraform module/lokocfg.

Note: this issue also affects upcoming Tinkerbell platform.

invidian:
Currently, bare metal environment requires resolvable DNS names for API
server and etcd servers. For user without running DNS server locally or
for testing, it would be nice to be able have some workaround for it.
One would be to populate /etc/hosts of nodes with required entries.

This is almost possible, using CLC snippets for controller nodes, except
the bootkube, which runs in rkt container and have it's own /etc/hosts
file, so health check on static kube-apiserver pod never succeeds. To
make it work, rkt parameter --hosts-entry=host needs to be added.

I suggest we add --hosts-entry=host to bootkube, as potential breakage
impact is minimal and this will save as from exposing general
bootkube_rkt_extra_args in the Terraform module/lokocfg.

By the issue, it is not required to be applied on all platforms, but
again, the breakage chance seems minimal and generally we should be
heading towards having quite unified configurations across all the
platforms, so this commit adds it to all of them.

Closes #408


assets/lokomotive-kubernetes/aws/flatcar-linux/kubernetes/cl/controller.yaml.tmpl
```diff
@@ -157,6 +157,7 @@ storage:
             quay.io/kinvolk/bootkube:v0.14.0-helm-amd64 \
             --net=host \
             --dns=host \
+            --hosts-entry=host \
```
johananl, line 160: Have we verified this doesn't have side effects? Also, I believe you're proposing to add this flag to all platforms for the sake of uniformity, however does it make sense to do that on platforms which don't need this functionality? I think it might be better to only add the flag where it's needed, to minimize potential "surprises" in the future.

invidian: >Have we verified this doesn't have side effects?

I cannot come up with any possible side effects of that beside benefits. This potentially allows to also use CLC snippets for /etc/hosts for names resolving for other platforms too (if using only Terraform code, which I know we don't support).

iaguis:
The bootkube container is running with host networking so it makes sense it also uses the host's `/etc/hosts`. So I'm fine with this change, can you rebase?

invidian:
>The bootkube container is running with host networking so it makes sense it also uses the host's /etc/hosts. So I'm fine with this change, can you rebase?

Rebased. Please have a look.

invidian:
Thanks @iaguis. Will merge once CI passes.

PR #436, opened 2020-05-18
Fix self-hosted kubelet on bare metal platform

### LINKED ISSUE #376: Self-hosted kubelet does not work on bare-metal, missing /etc/kubernetes/kubelet.config
I got this when trying to add Tinkerbell platform support for Lokomotive, based on `bare-metal` platform:
```
0 ✓ (106ms) 16:01:20 invidian@dellxps15mateusz ~/data/workspaces/tinkerbell $ klo kubelet-794qt
+ kubectl logs -f kubelet-794qt
./hyperkube kubelet --node-ip=192.168.50.10  --anonymous-auth=false --authentication-token-webhook --authorization-mode=Webhook --client-ca-file=/etc/kubernetes/ca.crt --cluster_dns=10.3.0.10 --cluster_domain=cluster.local --cni-conf-dir=/etc/kubernetes/cni/net.d --config=/etc/kubernetes/kubelet.config --kubeconfig=/etc/kubernetes/kubeconfig --lock-file=/var/run/lock/kubelet.lock --network-plugin=cni --pod-manifest-path=/etc/kubernetes/manifests --read-only-port=0 --volume-plugin-dir=/var/lib/kubelet/volumeplugins --node-labels=node.kubernetes.io/master,node.kubernetes.io/controller=true --register-with-taints=node-role.kubernetes.io/master=:NoSchedule --address=192.168.50.10

Flag --anonymous-auth has been deprecated, This parameter should be set via the config file specified by the Kubelet's --config flag. See https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/ for more information.
Flag --authentication-token-webhook has been deprecated, This parameter should be set via the config file specified by the Kubelet's --config flag. See https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/ for more information.
Flag --authorization-mode has been deprecated, This parameter should be set via the config file specified by the Kubelet's --config flag. See https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/ for more information.
Flag --client-ca-file has been deprecated, This parameter should be set via the config file specified by the Kubelet's --config flag. See https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/ for more information.
Flag --cluster-dns has been deprecated, This parameter should be set via the config file specified by the Kubelet's --config flag. See https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/ for more information.
Flag --cluster-domain has been deprecated, This parameter should be set via the config file specified by the Kubelet's --config flag. See https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/ for more information.
Flag --pod-manifest-path has been deprecated, This parameter should be set via the config file specified by the Kubelet's --config flag. See https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/ for more information.
Flag --read-only-port has been deprecated, This parameter should be set via the config file specified by the Kubelet's --config flag. See https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/ for more information.
Flag --address has been deprecated, This parameter should be set via the config file specified by the Kubelet's --config flag. See https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/ for more information.
F0428 13:57:45.805795   19635 server.go:199] failed to load Kubelet config file /etc/kubernetes/kubelet.config, error failed to read kubelet config file "/etc/kubernetes/kubelet.config", error: open /etc/kubernetes/kubelet.config: no such file or directory
```

Looking at the source code, it seems we don't create `kubelet.config` file on bare-metal platform. We don't test self-hosted kubelet there either.
```
0 ✓ (68.0ms) 18:43:55 invidian@dellxps15mateusz ~/repos/kinvolk/lokomotive (invidian/tinkerbell-platform)*$ git grep kubelet.config
assets/lokomotive-kubernetes/aws/flatcar-linux/kubernetes/cl/controller.yaml.tmpl:          --config=/etc/kubernetes/kubelet.config \
assets/lokomotive-kubernetes/aws/flatcar-linux/kubernetes/cl/controller.yaml.tmpl:          cat <<EOF >/etc/kubernetes/kubelet.config
assets/lokomotive-kubernetes/aws/flatcar-linux/kubernetes/cl/controller.yaml.tmpl:          apiVersion: kubelet.config.k8s.io/v1beta1
assets/lokomotive-kubernetes/aws/flatcar-linux/kubernetes/workers/cl/worker.yaml.tmpl:          --config=/etc/kubernetes/kubelet.config \
assets/lokomotive-kubernetes/aws/flatcar-linux/kubernetes/workers/cl/worker.yaml.tmpl:          cat <<EOF >/etc/kubernetes/kubelet.config
assets/lokomotive-kubernetes/aws/flatcar-linux/kubernetes/workers/cl/worker.yaml.tmpl:          apiVersion: kubelet.config.k8s.io/v1beta1
assets/lokomotive-kubernetes/bootkube/resources/charts/kubelet/templates/kubelet-ds.yaml:          --config=/etc/kubernetes/kubelet.config \
assets/lokomotive-kubernetes/kvm-libvirt/flatcar-linux/kubernetes/cl/controller.yaml.tmpl:          --config=/etc/kubernetes/kubelet.config \
assets/lokomotive-kubernetes/kvm-libvirt/flatcar-linux/kubernetes/cl/controller.yaml.tmpl:          cat <<EOF >/etc/kubernetes/kubelet.config
assets/lokomotive-kubernetes/kvm-libvirt/flatcar-linux/kubernetes/cl/controller.yaml.tmpl:          apiVersion: kubelet.config.k8s.io/v1beta1
assets/lokomotive-kubernetes/kvm-libvirt/flatcar-linux/kubernetes/workers/cl/worker.yaml.tmpl:          --config=/etc/kubernetes/kubelet.config \
assets/lokomotive-kubernetes/kvm-libvirt/flatcar-linux/kubernetes/workers/cl/worker.yaml.tmpl:          cat <<EOF >/etc/kubernetes/kubelet.config
assets/lokomotive-kubernetes/kvm-libvirt/flatcar-linux/kubernetes/workers/cl/worker.yaml.tmpl:          apiVersion: kubelet.config.k8s.io/v1beta1
assets/lokomotive-kubernetes/packet/flatcar-linux/kubernetes/cl/controller.yaml.tmpl:          --config=/etc/kubernetes/kubelet.config \
assets/lokomotive-kubernetes/packet/flatcar-linux/kubernetes/cl/controller.yaml.tmpl:          cat <<EOF >/etc/kubernetes/kubelet.config
assets/lokomotive-kubernetes/packet/flatcar-linux/kubernetes/cl/controller.yaml.tmpl:          apiVersion: kubelet.config.k8s.io/v1beta1
assets/lokomotive-kubernetes/packet/flatcar-linux/kubernetes/workers/cl/worker.yaml.tmpl:            --config=/etc/kubernetes/kubelet.config \
assets/lokomotive-kubernetes/packet/flatcar-linux/kubernetes/workers/cl/worker.yaml.tmpl:          cat <<EOF >/etc/kubernetes/kubelet.config
assets/lokomotive-kubernetes/packet/flatcar-linux/kubernetes/workers/cl/worker.yaml.tmpl:          apiVersion: kubelet.config.k8s.io/v1beta1
pkg/components/rook/manifests.go:  # kubelet directory path, if kubelet configured to use other than /var/lib/kubelet path.
```

invidian:
This PR fixes crashing self-hosted kubelet DaemonSet on bare metal and enables the tests for it on this platform, to make sure that it is actually fixed.

The PR use newly introduced Terraform modules from PR #392, to avoid the differences between the Ignition configuration across platforms, which caused this exact bug.

Closes #376


assets/lokomotive-kubernetes/bare-metal/flatcar-linux/kubernetes/cl/controller.yaml.tmpl
```diff
@@ -171,6 +173,19 @@ storage:
             --net=host \
             --dns=host \
             --exec=/bootkube -- start --asset-dir=/assets "$@"
+    - path: /etc/kubernetes/configure-kubelet-cgroup-driver
```
iaguis, line 176: You can mention in the commit message that we do this for all the other platforms.

assets/lokomotive-kubernetes/bare-metal/flatcar-linux/kubernetes/cl/controller.yaml.tmpl
```diff
@@ -95,6 +96,7 @@ systemd:
           --cluster_dns=${cluster_dns_service_ip} \
           --cluster_domain=${cluster_domain_suffix} \
           --cni-conf-dir=/etc/kubernetes/cni/net.d \
+          --config=/etc/kubernetes/kubelet.config \
```
iaguis, line 99: So this is the root cause of the self-hosted kubelet not working on bare metal?

invidian:
>So in the end we're not using the new Terraform modules here?

No, because it changes 10 other things, it seems too much of a burden for me. I created the issues for all those things instead. You can see those changes here though: https://github.com/kinvolk/lokomotive/compare/invidian/bare-metal-fix-self-hosted-kubelet-backup?expand=1

assets/lokomotive-kubernetes/bare-metal/flatcar-linux/kubernetes/cl/controller.yaml.tmpl
invidian: The root cause is that we don't generate this file using systemd service, so self-hosted kubelet which use this flag cannot find the configuration file and crashes.

assets/lokomotive-kubernetes/bare-metal/flatcar-linux/kubernetes/cl/controller.yaml.tmpl
invidian: Improved the commit message. PTAL.

invidian:
I had to update generated assets. @iaguis should I merge it or wait for 2nd approval?

iaguis:
I'd like to have 2 reviews here since this is not a trivial change.

Also, CI is failing.

invidian:
>Also, CI is failing.

Interesting, some changes must have gone lost while rebasing :(

invidian:
CI is passing again.

Makefile
```diff
@@ -6,7 +6,7 @@ VERSION :=
 MOD ?= vendor
 DOCS_DIR ?= docs/cli
 
-ALL_BUILD_TAGS := "aws,packet,aks,e2e,disruptivee2e,poste2e"
+ALL_BUILD_TAGS := "aws,packet,aks,e2e,baremetal,disruptivee2e,poste2e"
```
rata, line 9: This being split with previous commit makes the previous commit (1st commit in the series) just do half of the things (add tags but not compile it when it should). I don't see the point to split it.

Am I missing something?

pkg/platform/baremetal/template.go
```diff
@@ -47,6 +47,14 @@ module "bare-metal-{{.ClusterName}}" {
   worker_names       = {{.WorkerNames}}
   worker_macs        = {{.WorkerMacs}}
   worker_domains     = {{.WorkerDomains}}
+
+  {{- if .Labels}}
+  labels = [
+  {{- range .Labels}}
+    "{{.}}",
```
rata, line 54: Can you please assign a name instead? This seems quite cryptic without any reason, IMHO.

assets/lokomotive-kubernetes/bare-metal/flatcar-linux/kubernetes/cl/worker.yaml.tmpl
```diff
@@ -92,7 +92,7 @@ storage:
           KUBELET_IMAGE_URL=docker://k8s.gcr.io/hyperkube
           KUBELET_IMAGE_TAG=v1.18.3
           KUBELET_IMAGE_ARGS="--exec=/usr/local/bin/kubelet"
-          NODE_LABELS="node.kubernetes.io/node"
+          NODE_LABELS="${kubelet_labels}"
```
rata, line 95: Why change it this way? On Packet, for example, we use:

```
          NODE_LABELS="node.kubernetes.io/node,${node_labels}"
```

Why not do the same?

assets/lokomotive-kubernetes/bare-metal/flatcar-linux/kubernetes/profiles.tf
```diff
@@ -224,6 +224,7 @@ data "template_file" "worker-configs" {
     cluster_dns_service_ip = module.bootkube.cluster_dns_service_ip
     cluster_domain_suffix  = var.cluster_domain_suffix
     ssh_keys               = jsonencode(var.ssh_keys)
+    kubelet_labels         = join(",", concat(["node.kubernetes.io/node"], var.labels))
```
rata, line 227: And we can remove the `join()` and `concat()` if we do that on on worker.yaml.tmpl and we use labels as a string (as we do on packet).

Why not do the same as on Packet, for example?

assets/lokomotive-kubernetes/bare-metal/flatcar-linux/kubernetes/variables.tf
```diff
@@ -59,6 +59,12 @@ variable "clc_snippets" {
   default     = {}
 }
 
+variable "labels" {
+  type        = list(string)
+  description = "List of labels for worker nodes."
+  default     = []
+}
+
```
rata, line 67: Idem, why not use the same format on all platforms? Packet is using string (instead of list), this will just create more work when we want to unify the behaviour for no good reason. Or is there a reason?

ci/baremetal/baremetal-cluster.lokocfg.envsubst
```diff
@@ -31,4 +31,9 @@ cluster "bare-metal" {
     "node2",
     "node3",
   ]
+
+  labels = [
+    "testing.io=yes",
+    "roleofnode=testing",
+  ]
```
rata, line 38: Squash this with enabling the test

assets/lokomotive-kubernetes/bare-metal/flatcar-linux/kubernetes/variables.tf
invidian: We take list of labels, we shouldn't treat a list as a string. We should be unifying the behavior towards correct formats, not the opposite IMO.

assets/lokomotive-kubernetes/bare-metal/flatcar-linux/kubernetes/cl/worker.yaml.tmpl
invidian: IMO the template should not make an assumption what labels will be used. This is aiming towards https://github.com/kinvolk/lokomotive/issues/449. The less data we hardcode in the template, the better also.

pkg/platform/baremetal/template.go
invidian: OK

Makefile
invidian: The scope of the change is different. 1st commit introduces the tests. 2nd commit enables them, as it touches different area of code. It allows to disable the test without removing the code and vice versa in the future. For me it seems like a reasonable split.

assets/lokomotive-kubernetes/bare-metal/flatcar-linux/kubernetes/profiles.tf
invidian: Hmm, ideally `join()` should be called inside the function, but that does not work, because we use `template_file` instead of `templatefile()`. Let me try fixing it.

ci/baremetal/baremetal-cluster.lokocfg.envsubst
invidian: Actually, I'm not sure if I should squash.

assets/lokomotive-kubernetes/bare-metal/flatcar-linux/kubernetes/cl/worker.yaml.tmpl
rata: There are other paths towards #449, code-duplication might be possible to remove with symlinks, for example.

I'd even argue this commit goes against #449: if the templates are all different, it's harder to remove duplication in the future. This commit is making templates across platforms unconsistentn and different.

assets/lokomotive-kubernetes/bare-metal/flatcar-linux/kubernetes/variables.tf
rata: Well, this series is not doing so. It is just creating inconsistencies. I'm fine if you prefer to have it this way and create an issue to change packet to this format with the "tech-debt" label.

assets/lokomotive-kubernetes/bare-metal/flatcar-linux/kubernetes/profiles.tf
invidian: I pushed the commit to change it to use `template_file` + `null_resource`.

iaguis:
Needs rebase.

invidian:
Rebased.

assets/lokomotive-kubernetes/bare-metal/flatcar-linux/kubernetes/variables.tf
invidian: Created https://github.com/kinvolk/lokomotive/issues/489.

invidian:
Tests are passing again.

assets/lokomotive-kubernetes/bare-metal/flatcar-linux/kubernetes/variables.tf
rata: Cool, thanks!

assets/lokomotive-kubernetes/bare-metal/flatcar-linux/kubernetes/profiles.tf
rata: That seems overly complicated for no good reason, IMHO. It seems more readable as it is here.

Or isn't it easier to move it to `tempaltefile()` instead to add a null resource? Would that solve the join thing?

invidian: Hm, I messed something up. I thought the output of this template is used in more than one place, but this is the case only for `container-linux-install-configs`, which I decided not to touch in this PR.

I removed the `null_resource` and I use `templatefile()` directly.

Also, for future reference, if we need to use `count.index` with `templatefile`, `for` loop with `locals` will do just fine.

invidian:
I'd change it to `map[string]string`, as this is also how you define labels in Kubernetes.

ci/baremetal/baremetal-cluster.lokocfg.envsubst
```diff
@@ -35,4 +35,9 @@ cluster "bare-metal" {
   # Acts as a smoke test to check if API server is functional after addition
   # of extra flags.
   oidc {}
+
+  labels = {
```
iaguis, line 39: Should we add a simple test that checks labels are there (without any reboots)?

invidian: `TestSelfHostedKubeletLabels` will already test that, though we could perhaps have one more test to just seek for those labels on the nodes. I'm not sure if we should implement here though, what do you think?

iaguis: I was thinking that since that we start skipping that test here we should add a simple labels test to not lose coverage. Does that make sense?

invidian: Ah yes, I checked the code on the wrong branch. let me implement it then.

test/components/kubernetes/kubelet_labels_test.go
```diff
@@ -0,0 +1,41 @@
...
+	client := testutil.CreateKubeClient(t)
+
+	nodes, err := client.CoreV1().Nodes().List(metav1.ListOptions{
+		LabelSelector: "testing.io=yes,roleofnode=testing",
+	})
+	if err != nil {
+		t.Errorf("could not list nodes: %v", err)
+	}
+
+	if len(nodes.Items) == 0 {
```
iaguis, line 38: Although unlikely, this will pass if only one worker has the labels when there are several of them. Maybe we can get all nodes with label `node.kubernetes.io/node=` and make sure they're != 0 and they all contain the labels we want?

Feel free to ignore if that doesn't make sense to you.

invidian: Hm, I can imagine we could have a case where we have more than one worker pool in the CI configuration, then this won't really work. And since we don't have access to the configuration from this test, we cannot really tell how many nodes we should have. IMO it's good enough for the time being.

pkg/platform/baremetal/baremetal_internal_test.go
```diff
@@ -0,0 +1,27 @@
+package baremetal
```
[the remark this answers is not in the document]
invidian: Done, thanks.

PR #479, opened 2020-05-25
Makefile: update golangci-lint to 1.27.0

invidian:
It is latest version available now.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -67,7 +67,7 @@ lint: build-slim build-test
 
 .PHONY: lint-docker
 lint-docker:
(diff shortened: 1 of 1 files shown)


invidian:
>Also, would this docker thing mean the CI will use it? Or how would the CI be updated?

The CI is updated separately, though it should be with sync with this PR. This PR is only for local runs.

invidian:
>LGTM if you test that it works with all the optiins we use today (as this PR is not testing that, IIUC).

It does, I use latest version locally.

PR #483, opened 2020-05-26
Handle OS interrupts in lokoctl to fix leaking terraform process.

### LINKED ISSUE #329: Leaking Terraform processes when terminating a cluster operation
When we Ctrl+C out of a `lokoctl cluster apply` operation (and likely `destroy`, too), the lokoctl process terminates without killing its child processes. As a result, depending on the exact stage in which Ctrl+C was pressed, one or more Terraform-related processes are left behind as orphans.

Steps to reproduce:

1. Run `lokoctl cluster apply -v` against any `.lokocfg` file.
1. Wait for Terraform to start creating infrastructure, then hit Ctrl+C.
1. Check for orphan processes: `ps -ef | grep terraform`.

Here is what I'm getting:

```
johannes 3061095    1421  0 13:52 pts/3    00:00:01 /usr/bin/terraform internal-plugin provisioner file
johannes 3061106 3061095  0 13:52 pts/3    00:00:00 /usr/bin/terraform internal-plugin provisioner file
johannes 3061107    1421  0 13:52 pts/3    00:00:00 /tmp/lk/assets/terraform/.terraform/plugins/linux_amd64/terraform-provider-template_v2.1.2_x4
johannes 3061125    1421  0 13:52 pts/3    00:00:01 /usr/bin/terraform internal-plugin provisioner remote-exec
johannes 3061137 3061125  0 13:52 pts/3    00:00:00 /usr/bin/terraform internal-plugin provisioner remote-exec
johannes 3061165    1421  0 13:52 pts/3    00:00:02 /tmp/lk/assets/terraform/.terraform/plugins/linux_amd64/terraform-provider-tls_v2.1.1_x4
johannes 3061179    1421  0 13:52 pts/3    00:00:00 /tmp/lk/assets/terraform/.terraform/plugins/linux_amd64/terraform-provider-local_v1.4.0_x4
johannes 3061195    1421  0 13:52 pts/3    00:00:00 /home/johannes/.terraform.d/plugins/terraform-provider-ct_v0.4.0
johannes 3061212    1421  0 13:52 pts/3    00:00:00 /tmp/lk/assets/terraform/.terraform/plugins/linux_amd64/terraform-provider-packet_v2.7.5_x4
johannes 3061228    1421  0 13:52 pts/3    00:00:00 /tmp/lk/assets/terraform/.terraform/plugins/linux_amd64/terraform-provider-null_v2.1.2_x4
```

The orphan processes stay in memory indefinitely, though thankfully it doesn't seem like they keep executing Terraform operations in this state.

We should most likely implement a proper cleanup procedure which we always run in response to a SIGINT.

ipochi:
Fixes #329 

Registers a Signal handler for terraform process and components installation/deletion.

Before running the `lokoctl cluster apply`
```
ps ef | grep terraform | wc -l
0
```

```
module.packet-liger.data.ct_config.controller-install-ignitions[0]: Refreshing state...
module.worker-madrid.data.ct_config.install-ignitions: Refreshing state...
module.worker-madrid.packet_device.nodes[0]: Creating...
module.packet-liger.packet_device.controllers[0]: Creating...
module.worker-madrid.packet_device.nodes[0]: Still creating... [10s elapsed]
module.packet-liger.packet_device.controllers[0]: Still creating... [10s elapsed]
^CInterrupt received.
Please wait for Terraform to exit or data loss may occur.
Gracefully shutting down...
Stopping operation...
```
After Ctrl + C
```
ps ef | grep terraform | wc -l
0
```

### DIFF
--- a/pkg/terraform/executor.go
+++ b/pkg/terraform/executor.go
@@ -296,6 +296,8 @@ func (ex *Executor) ExecuteAsync(args ...string) (int, chan struct{}, error) {
 	cmd.Stdout = wPipe
 	cmd.Stderr = wPipe
 
+	// Initialize the signal handler.
+	h := signalHandler(ex.logger)
 	// Start Terraform.
 	err := cmd.Start()
 	if err != nil {
@@ -319,6 +321,10 @@ func (ex *Executor) ExecuteAsync(args ...string) (int, chan struct{}, error) {
 			ioutil.WriteFile(ex.failPath(cmd.Process.Pid), []byte(err.Error()), 0660)
 		}
 
+		// Once the process is finished whether successfully or terminated by an
+		// interrupt, we stop listening for the interrupt and close the channel.
+		h.stop()
+
 		// Close descriptors.
 		wPipe.Close()
 		logFile.Close()
@@ -330,8 +336,15 @@ func (ex *Executor) ExecuteAsync(args ...string) (int, chan struct{}, error) {
 
 // ExecuteSync is like Execute, but synchronous.
 func (ex *Executor) ExecuteSync(args ...string) ([]byte, error) {
+	// Initialize the signal handler.
+	h := signalHandler(ex.logger)
+
 	cmd := ex.generateCommand(args...)
-	return cmd.Output()
+	output, err := cmd.Output()
+
+	h.stop()
+
+	return output, err
 }
 
 // Plan runs 'terraform plan'.
--- a/pkg/terraform/signals.go
+++ b/pkg/terraform/signals.go
@@ -1,0 +1,60 @@
+// Copyright 2020 The Lokomotive Authors
+// Copyright 2017 CoreOS, Inc
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package terraform
+
+import (
+	"os"
+	"os/signal"
+
+	"github.com/sirupsen/logrus"
+)
+
+// handler defines signal handler.
+type handler struct {
+	logger *logrus.Entry
+	// signalCh stores signals received from the system.
+	signalCh chan os.Signal
+}
+
+// signalHandler creates a new Handler.
+func signalHandler(logger *logrus.Entry) *handler {
+	// Different signals that we are listening to. For the time being putting
+	// only os.Interrupt, making a slice so that we can easily add more.
+	signals := []os.Signal{os.Interrupt}
+
+	signalCh := make(chan os.Signal, 1)
+	signal.Notify(signalCh, signals...)
+
+	// Goroutine that prints the message to the user in case the --verbose
+	// option is not used.
+	go func() {
+		for range signalCh {
+			logger.Warnln("Interrupt received, please wait for Terraform to terminate.")
+		}
+	}()
+
+	return &handler{
+		signalCh: signalCh,
+		logger:   logger,
+	}
+}
+
+func (h *handler) stop() {
+	// Stop listening for interrupts on the channel
+	signal.Stop(h.signalCh)
+	// Close the channel.
+	close(h.signalCh)
+}
--- a/pkg/terraform/doc.go
+++ b/pkg/terraform/doc.go
@@ -1,0 +1,18 @@
+// Copyright 2020 The Lokomotive Authors
+// Copyright 2017 CoreOS, Inc
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package terraform handles the execution of terraform processes to bring
+// up the infrastructure for the Lokomotive cluster.
+package terraform


pkg/signals/signals.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+}
+
+// Register starts listening for the os interrupts.
+func (h *Handler) Register() {
+	go func() {
+		defer signal.Stop(h.signalCh)
+
+		for s := range h.signalCh {
+			if s == os.Interrupt {
```
invidian, line 48: Should we use switch/case or `s != os.Interrupt` to remove the indentation?

pkg/signals/signals.go
```diff
@@ -0,0 +1,89 @@
...
+				if h.signalType == 0 {
```
invidian, line 49: Again, we could switch/case on `h.signalType`.

pkg/signals/signals.go
```diff
@@ -0,0 +1,89 @@
...
+					h.signalType = signalAbort
+
+					log.Println("SIGINT Received, waiting for the graceful shutdown")
```
invidian, line 52: We should use `logrus` here I think, not `log`. Maybe we can take it as an argument for `NewSignalHandler()` and store in `Handler`?

pkg/signals/signals.go
```diff
@@ -0,0 +1,89 @@
...
+
+					continue
+				}
+
+				log.Println("Second interrupt received, exiting immediately.")
+
+				h.signalType = signalTerminate
```
invidian, line 59: Is this needed if we call `os.Exit()` right after?

pkg/signals/signals.go
```diff
@@ -0,0 +1,89 @@
...
+			}
+		}
+	}()
+}
+
+// RegisterCmd starts listening for the os interrupts for the provided Cmd.
+// and passes the interrupt to the running process.
+func (h *Handler) RegisterCmd(cmd *exec.Cmd) {
+	go func() {
+		defer signal.Stop(h.signalCh)
```
invidian, line 71: If I register multiple cmds, is it fine to stop the channel multiple times?

pkg/signals/signals.go
```diff
@@ -0,0 +1,89 @@
...
+
+		for s := range h.signalCh {
+			if s == os.Interrupt {
```
invidian, line 74: Again, please revert the conditional, to keep the happy path not deeply indented. 

```go
if s != os.Interrupt || cmd.Process == nil {
  continue
}
```

pkg/signals/signals.go
```diff
@@ -0,0 +1,89 @@
...
+				os.Exit(1) //nolint:gomnd
+			}
+		}
+	}()
+}
+
+// RegisterCmd starts listening for the os interrupts for the provided Cmd.
+// and passes the interrupt to the running process.
+func (h *Handler) RegisterCmd(cmd *exec.Cmd) {
+	go func() {
```
invidian, line 70: Do we need a goroutine for each cmd? Maybe it would be easier to collect them and once we receive the signal, iterate?

pkg/signals/signals.go
```diff
@@ -0,0 +1,89 @@
...
+	signalCh := make(chan os.Signal, 1)
+	signal.Notify(signalCh, signals...)
+
+	return &Handler{
+		signalCh: signalCh,
+	}
+}
+
+// Register starts listening for the os interrupts.
+func (h *Handler) Register() {
```
invidian, line 43: What if I call `Register()` twice, does everything still work just fine?

pkg/components/util/install.go
```diff
@@ -26,10 +26,15 @@ import (
...
 	"github.com/kinvolk/lokomotive/pkg/components"
 	"github.com/kinvolk/lokomotive/pkg/k8sutil"
+	"github.com/kinvolk/lokomotive/pkg/signals"
 )
 
 // InstallComponent installs given component using given kubeconfig.
 func InstallComponent(name string, c components.Component, kubeconfig string) error {
+	// Get a new instance of SignalHandler and register to listen for os interrupts.
+	h := signals.NewSignalHandler()
+	h.Register()
```
invidian, line 36: It seems we **will** register it multiple times. What if I interrupt the process after first component has been installed? Also, what's the point of interrupting, if it does not abort anything? I think we should print the correct error message, that the installing won't be interrupted if one hits CTRL+C.

ipochi:
After a call with @invidian we found out that `terraform apply` handles the interrupts but `terraform init` doesn't and this behavior is reflected the same in `lokoctl` as well.

An upstream issue has been filed for the same.

https://github.com/hashicorp/terraform/issues/25118

@invidian @iaguis please re-review.

pkg/terraform/executor.go
```diff
@@ -106,7 +106,6 @@ func NewExecutor(conf Config) (*Executor, error) {
 	ex.logger = log.WithFields(log.Fields{
 		"phase": "infrastructure",
 	})
-
```
iaguis, line 109: This has nothing to do with the commit, please put in a separate commit if you think it improves readability.

pkg/terraform/executor.go
```diff
@@ -296,14 +295,16 @@ func (ex *Executor) ExecuteAsync(args ...string) (int, chan struct{}, error) {
 	cmd.Stdout = wPipe
 	cmd.Stderr = wPipe
 
+	// Initialize the signal handler.
+	h := signalHandler(ex.logger)
+	// Register the signal handler to handle os interrupts.
```
iaguis, line 300: This comment seems out of place.

pkg/terraform/signals.go
```diff
@@ -1,3 +1,18 @@
+// Copyright 2020 The Lokomotive Authors
```
iaguis, line 1: This is more than what the commit msg says. Please add a mention of it in the commit message or put it in a different commit.

ipochi: Yes, this should have been part of the first commit. Done.

ipochi:
> I have some nits and a general comment.
> 
> From my testing it seems when I interrupt lokoctl at, say, `terraform init` this doesn't stop lokoctl. I'm aware terraform init doesn't handle interrupts but I would expect lokoctl to stop after `terraform init` is finished instead of continuing with the next steps (`terraform plan` and so on). This works fine if I do two interrupts for example when `terraform plan` is running, I guess because the process ends with exit status 1 and we propagate the error.
> 
> Should we add a channel to send interrupt information from the signal handler to the `Executor` so we can `os.Exit()` or error out after the command is finished if we get an interrupt?

It would become quite messy, besides there would be two branches where it would be handle the differently if the current process is `terraform version/init` vs `terraform plan/apply/refresh`

Nevertheless, I am opening up a follow up issue that tracks this limitation and and will also link the upstream issue in the follw up issue.

ipochi:
Follow up issue created #555

PR #487, opened 2020-05-27
Makefile: Disable goerr113 linter check

johananl:
We aren't aligned on a single approach regarding adding info to
Go errors. Since we seem to be using `fmt.Errorf` more than other
approaches (e.g. wrapping), we should disable this check for now.


johananl:
Merging with a single approver since it's a trivial change which is currently blocking feature work.

PR #488, opened 2020-05-27
Makefile: error on bad git settings when linting 

invidian:
To avoid misbehavior between running lint locally and on CI, make lint
target fail if user has invalid git configuration and suggest how to
resolve that.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -61,8 +61,7 @@ test: run-unit-tests
 
 .PHONY: lint
 lint: build-slim build-test
-	# Note: Make sure that you run `git config diff.noprefix false` in this repo
-	# See this issue for more details: https://github.com/golangci/golangci-lint/issues/948
+	@if [ "$$(git config --get diff.noprefix)" = "true" ]; then printf "\n\ngolangci-lint has a bug and can't run with the current git configuration: 'diff.noprefix' is set to 'true'. To override this setting for this repository, run the following command:\n\n'git config diff.noprefix false'\n\nFor more details, see https://github.com/golangci/golangci-lint/issues/948.\n\n\n"; exit 1; fi
 	golangci-lint run --enable-all --disable=godox,gochecknoglobals,goerr113 --max-same-issues=0 --max-issues-per-linter=0 --build-tags $(ALL_BUILD_TAGS) --new-from-rev=$$(git merge-base $$(cat .git/resource/base_sha 2>/dev/null || echo "origin/master") HEAD) --modules-download-mode=$(MOD) --timeout=5m --exclude-use-default=false ./...
 
 .PHONY: lint-docker


Makefile
```diff
@@ -60,10 +60,12 @@ build-slim:
 test: run-unit-tests
 
 .PHONY: lint
+lint: diffnoprefix := $(shell sh -c "git config --get --show-origin diff.noprefix | awk '/file:\.git\/config\s+\S+/ {print \$$2}'")
 lint: build-slim build-test
-	# Note: Make sure that you run `git config diff.noprefix false` in this repo
-	# See this issue for more details: https://github.com/golangci/golangci-lint/issues/948
+	# To workaround: https://github.com/golangci/golangci-lint/issues/948
```
johananl, line 65: Either "To work around" or "A workaround for" :-)

Makefile
```diff
@@ -60,10 +60,12 @@ build-slim:
...
+	git config --replace-all diff.noprefix false
 	golangci-lint run --enable-all --disable=godox,gochecknoglobals,goerr113 --max-same-issues=0 --max-issues-per-linter=0 --build-tags $(ALL_BUILD_TAGS) --new-from-rev=$$(git merge-base $$(cat .git/resource/base_sha 2>/dev/null || echo "origin/master") HEAD) --modules-download-mode=$(MOD) --timeout=5m --exclude-use-default=false ./...
+	if [ -z "$(diffnoprefix)" ]; then git config --unset diff.noprefix; else git config --replace-all diff.noprefix $(diffnoprefix); fi
```
johananl, line 68: This isn't global, right? I'd be careful with doing laptop-level changes on behalf of users...

invidian: No, it's only touching `.git/config`. For it being global, you would have to use `--global`. It only touches local config and preserves the value, if you have something set there. 

The only "controversial" part is `--replace-all`, if you have multiple values set for the same key, it will wipe them and just set the effective one, but that seem rare enough case to me to accept it.

Makefile
invidian: Fixed, thanks.

Makefile
```diff
@@ -60,10 +60,12 @@ build-slim:
 test: run-unit-tests
 
 .PHONY: lint
+lint: diffnoprefix := $(shell sh -c "git config --get --show-origin diff.noprefix | awk '/file:\.git\/config\s+\S+/ {print \$$2}'")
 lint: build-slim build-test
-	# Note: Make sure that you run `git config diff.noprefix false` in this repo
-	# See this issue for more details: https://github.com/golangci/golangci-lint/issues/948
+	# A workaround for https://github.com/golangci/golangci-lint/issues/948.
+	git config --replace-all diff.noprefix false
```
rata, line 66: Why not just run `git config diff.noprefix` and if it is false, fail with a clear error (like "run XXX due to bug YYY")?

Then, IIUC, we have the best of both worlds:
 * Not modify git config for the user (specially the replace-all flag that we can't properly restore)
 * A user without that properly set will easily notice


I looked at env vars for git [here][link] and it seems there is no easy way to set it, if not that would be the way to proceed (`ENV=asd cmd`) IMHO. But doesn't seem possible with this option (I had a quick look, so I might be missing something).

[link]: https://git-scm.com/book/en/v2/Git-Internals-Environment-Variables

invidian: With this approach, you can use `diff.noprefix = true` for this repository and still have properly functional linter. Your approach enforces `noprefix = false`.

rata: Yes, enforces that only if you want to run the linter locally. And it makes sense, you can't run the linter locally with that setting. IMHO, it's better to not try to hide that and be clear.

Makefile
```diff
@@ -60,10 +60,12 @@ build-slim:
...
 	golangci-lint run --enable-all --disable=godox,gochecknoglobals,goerr113 --max-same-issues=0 --max-issues-per-linter=0 --build-tags $(ALL_BUILD_TAGS) --new-from-rev=$$(git merge-base $$(cat .git/resource/base_sha 2>/dev/null || echo "origin/master") HEAD) --modules-download-mode=$(MOD) --timeout=5m --exclude-use-default=false ./...
+	if [ -z "$(diffnoprefix)" ]; then git config --unset diff.noprefix; else git config --replace-all diff.noprefix $(diffnoprefix); fi
```
rata, line 68: I don't think this will work correctly, if we used the `--replace-all` it will change it across the many references in the `gitconfig`. And we don't know to which value to use restore on all of the different places it might appear.

Makefile
invidian: What do you think @johananl ?

invidian:
@iaguis @johananl @ipochi what do you think?

Makefile
johananl: Sorry for the late response. Personally I'd avoid changing Git config on the user's behalf (even if it's local to the repo). IMO this falls under the "taking too much responsibility" category. Like @rata suggested above, I'd print a warning or error out if I have to but I wouldn't run `git config`.

johananl:
Commenting here, too: I'd avoid running `git config` on behalf of the user. If we can't fix the problem at the source (patch the liner?), I'd print a warning/error but not modify things on the user's laptop. As a user I wouldn't expect running `make` to change my tooling config or dev environment.

invidian:
Thanks. I'll make it fail if the config is wrong then.

invidian:
@rata @johananl I updated the PR as we discussed. Please have a look when you have time.

invidian:
>I'd consider exiting from make lint with a non-zero exit code to make it more visible that something went wrong.

Ugh, that was my attempt and failed :smile: I changed the logic to use `if`, it should be less error prone and it works now. Thanks for checking :)

>There are missing articles in the text printed to the user (AFAIK there should always be an article before singular nouns unless you are using "short phrasing" like in news titles). Could you please fix those?

I did my best, please have a look :see_no_evil:

Makefile
```diff
@@ -61,8 +61,7 @@ test: run-unit-tests
 
 .PHONY: lint
 lint: build-slim build-test
-	# Note: Make sure that you run `git config diff.noprefix false` in this repo
-	# See this issue for more details: https://github.com/golangci/golangci-lint/issues/948
+	if [ $$(git config --get diff.noprefix) == "true" ]; then echo -e "\nLinting won't work because you have a git setting 'diff.noprefix' set to 'true',\nwhich does not work reliably with the golangci-lint --new-from-rev flag. \n\nTo run the linter, set 'diff.noprefix' setting to 'true' for this repository using the following command:\n\n'git config diff.noprefix false'\n\nSee https://github.com/golangci/golangci-lint/issues/948 for more details."; exit 1; fi
```
rata, line 64: Why starting `\n` for the `echo`?

rata, line 64: `echo -e` is not portable (doesn't work on dash, for example). Please use `printf`

rata, line 64: This line is very difficult to add comments too :)

I'd do (but @johananl is the messages expert): s/Linting won't work/Linter can't run/

But I'm sure @johananl will think of a better way :)

invidian:
>echo -e is not portable and we can't choose the shell the user might be running. Just replacing with printf will do the trick :-)

Make commands are always run with `/bin/sh`, unless you specify `SHELL` variable IIRC.

Makefile
invidian: To separate the output from make printing command to execute, to make it more readable.

invidian: As said, I think it will never use dash :)

rata: Oh, makes sense. Thanks!

invidian: I think I actually wanted to write `Linting won't work reliably`. Would that make more sense? Because the linter can run, it just won't produce legitimate output.

invidian: https://www.gnu.org/software/make/manual/html_node/Choosing-the-Shell.html

rata:
> > echo -e is not portable and we can't choose the shell the user might be running. Just replacing with printf will do the trick :-)
> 
> Make commands are always run with `/bin/sh`, unless you specify `SHELL` variable IIRC.

Then, a stronger argument to change it. This is dash in several systems, like Ubuntu and Debian:

```
# ls -l /bin/sh
lrwxrwxrwx 1 root root 4 Mar 30 06:45 /bin/sh -> dash
```

Makefile
rata: What you said is actually the opposite ;)

Also, we really can't know which shell the user is using, using printf is a trivial change, `echo -e` is not recommended since years... I really don't see a point arguing about this.

invidian:
On Arch it is:
```sh
$ ls -la /bin/sh
lrwxrwxrwx 1 root root 4 May 22 14:29 /bin/sh -> bash
```
:scream: :smile:

Will change!

Makefile
rata: I don't think that is better, it seems confusing as I think it will work sometimes (i.e. not reliably, but sometimes).

I think something like this is better, but I'm okay with almost any text. It is clear enough in all versions:

`The linter has a bug and can't run with the current git config: 'git.diffprefix' is set to 'true'. To disable it, you can use the following command: 'git config diff.noprefix false'. See https://github.com/golangci/golangci-lint/issues/948 for more details.`

invidian: Okay, I rephrased your suggestion a bit. This is how the output looks like right now:
```sh
$ make lint
CGO_ENABLED=0 GO111MODULE=on go build \
        -mod=vendor \
        -ldflags "-X github.com/kinvolk/lokomotive/pkg/version.Version=`git describe --tags --always`-dirty -extldflags '-static'" \
        -buildmode=exe \
        -o lokoctl \
        github.com/kinvolk/lokomotive/cmd/lokoctl
go test -run=nonexistent -mod=vendor -tags="aws,packet,aks,e2e,baremetal,disruptivee2e,poste2e" -covermode=atomic -buildmode=exe -v ./... > /dev/null
if [ $(git config --get diff.noprefix) == "true" ]; then printf "\n\nThe linter has a bug and can't run with the current git config: 'diff.noprefix' is set to 'true'. To override this setting for this repository,
run the following command:\n\n'git config diff.noprefix false'\n\nFor more details, see https://github.com/golangci/golangci-lint/issues/948.\n\n\n"; exit 1; fi


The linter has a bug and can't run with the current git config: 'diff.noprefix' is set to 'true'. To override this setting for this repository, run the following command:

'git config diff.noprefix false'

For more details, see https://github.com/golangci/golangci-lint/issues/948.


make: *** [Makefile:64: lint] Error 1
```

What do you think?

rata: The text looks fine for me :)

invidian:
And thanks for your testing @rata. I'm sorry about lack of thorough testing on my side. I also tested with not having setting at all now.

>Add double quotes, like this: "$$(git config --get diff.noprefix)". In this case, it will compare the empty string to "true"

I implemented that.

>Bonus: to have nicer output, please change the line that starts with if ... to @if ... as then, the command being executed will not be printed, just the output (if any).

Good to know! But comments from other targets are printed anyway, so we need separation from them anyway, so I'll just leave it as it is.

Makefile
```diff
@@ -61,8 +61,7 @@ test: run-unit-tests
 
 .PHONY: lint
 lint: build-slim build-test
-	# Note: Make sure that you run `git config diff.noprefix false` in this repo
-	# See this issue for more details: https://github.com/golangci/golangci-lint/issues/948
+	if [ "$$(git config --get diff.noprefix)" = "true" ]; then printf "\n\nThe golangci-lint has a bug and can't run with the current git configuration: 'diff.noprefix' is set to 'true'. To override this setting for this repository, run the following command:\n\n'git config diff.noprefix false'\n\nFor more details, see https://github.com/golangci/golangci-lint/issues/948.\n\n\n"; exit 1; fi
```
iaguis, line 64: Small nit.
```suggestion
	if [ "$$(git config --get diff.noprefix)" = "true" ]; then printf "\n\ngolangci-lint has a bug and can't run with the current git configuration: 'diff.noprefix' is set to 'true'. To override this setting for this repository, run the following command:\n\n'git config diff.noprefix false'\n\nFor more details, see https://github.com/golangci/golangci-lint/issues/948.\n\n\n"; exit 1; fi
```

invidian: Done.

invidian:
>I'd use @if in this case (keeping the \n if you want) because it's a long code that includes the error text:

It's readability vs ambiguity here. I'd not expect make to run commands, which are not printed in the console, but I think it's fine in this case. Added and it indeed looks better!

PR #530, opened 2020-06-02
Deleting machine GOOS from Makefile for  build-slim and  install-slim 

### LINKED ISSUE #526: Makefile: Automatically detect machine GOOS
Right now we have hard coded `linux` in it.

knrt10:
Makefile: Previously it was hardcoded to linux. deleting it as go can automatically detect it`. Fixes #526

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,7 +49,7 @@ update-assets:
 # Once we change CI code to build outside GOPATH, GO111MODULE can be removed, so
 # we rely on defaults.
 build-slim:
-	CGO_ENABLED=0 GOOS=linux GO111MODULE=on go build \
+	CGO_ENABLED=0 GO111MODULE=on go build \
 		-mod=$(MOD) \
 		-ldflags $(LDFLAGS) \
 		-buildmode=exe \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -49,7 +58,7 @@ update-assets:
 # Once we change CI code to build outside GOPATH, GO111MODULE can be removed, so
 # we rely on defaults.
 build-slim:
-	CGO_ENABLED=0 GOOS=linux GO111MODULE=on go build \
+	CGO_ENABLED=0 GOOS=$(GOOS) GO111MODULE=on go build \
```
invidian, line 61: Hm, why do we need to set GOOS explicitly? Can we just remove it?

invidian: The produced binaries are exactly the same. If building on MacOS works just fine, I guess we can just leave it as it is?

knrt10: Yes, let me update the PR.

invidian: I'll actually test that on macOS before we remove it to understand how things work.

knrt10: I have macOS, I tested it, you can double check

invidian: If we build without it, it will work locally. But I guess for release builds, if we want to provide binaries for both macOS and Linux, we need to specify it then, to do cross-compile, right?

invidian: Just checked, we already do via goreleaser. All good, cool!

PR #532, opened 2020-06-02
platform/packet: Fix missing AWS provider on upgrade

johananl:
In #422 the AWS provider was removed from the root Terraform module. This breaks upgrades from versions which don't include #422.


pkg/platform/packet/template.go
```diff
@@ -207,6 +207,12 @@ module "worker-{{ $pool.Name }}" {
 module "dns" {
   source = "../lokomotive-kubernetes/dns/{{.Config.DNS.Provider}}"
 
+  {{ if eq .Config.DNS.Provider "route53" -}}
+  providers = {
```
invidian, line 211: This is not needed, provider install will be automatically passed to the module.

pkg/platform/packet/template.go
```diff
@@ -220,6 +226,17 @@ output "dns_entries" {
 }
 {{- end }}
 
+{{- if eq .Config.DNS.Provider "route53" }}
+provider "aws" {
+  version = "2.48.0"
+  alias   = "default"
```
invidian, line 232: `alias` is not needed.

pkg/platform/packet/template.go
johananl: I'm getting "region must be specified" without this.

pkg/platform/packet/template.go
johananl: I've re-added the provider as it was before #422. Also, other providers in this file do have `alias` so if we remove it we should do it for all of them.

pkg/platform/packet/template.go
invidian: Let me check.

invidian: With the following patch, things work just fine:
```diff
diff --git pkg/platform/packet/template.go pkg/platform/packet/template.go
index e3e6d572..521934f1 100644
--- pkg/platform/packet/template.go
+++ pkg/platform/packet/template.go
@@ -207,12 +207,6 @@ module "worker-{{ $pool.Name }}" {
 module "dns" {
   source = "../lokomotive-kubernetes/dns/{{.Config.DNS.Provider}}"

-  {{ if eq .Config.DNS.Provider "route53" -}}
-  providers = {
-    aws = aws.default
-  }
-
-  {{ end -}}
   cluster_name             = "{{ .Config.ClusterName }}"
   controllers_public_ipv4  = module.packet-{{.Config.ClusterName}}.controllers_public_ipv4
   controllers_private_ipv4 = module.packet-{{.Config.ClusterName}}.controllers_private_ipv4
@@ -229,7 +223,6 @@ output "dns_entries" {
 {{- if eq .Config.DNS.Provider "route53" }}
 provider "aws" {
   version = "2.48.0"
-  alias   = "default"
   # The Route 53 service doesn't need a specific region to operate, however
   # the AWS Terraform provider needs it and the documentation suggests to use
   # "us-east-1": https://docs.aws.amazon.com/general/latest/gr/r53.html.

```

pkg/platform/packet/template.go
invidian: I think we should have some development guidelines for quality and technical debt vs inconsistency, as this is recurring topic. I'll let other decide on this PR then :)

johananl: Good point. I've already started forming a list of proposed guidelines for this kind of cases. I'll add "consistency vs. quality".

pkg/platform/packet/template.go
johananl: @invidian I've just tried upgrading from 8ac5c734 to current master and it breaks with the following:

```
INFO[0002] Generating Terraform execution plan           phase=infrastructure

Error: Provider configuration not present

To work with module.dns.aws_route53_record.dns-records[2] its original
provider configuration at provider.aws.default is required, but it has been
removed. This occurs when a provider configuration is removed while objects
created by that provider still exist in the state. Re-add the provider
configuration to destroy module.dns.aws_route53_record.dns-records[2], after
which you can remove the provider configuration again.
```

Note that it complains about `provider.aws.default`.

pkg/platform/packet/template.go
johananl: Looks like `alias` **is** needed since the upgrade path fails when not explicitly injecting the provider using its alias into the `dns` module. This likely has to do with the fact that the providers were configured this way *to begin with*, however if we care about not breaking upgrades from `v0.1.0` looks like we don't have a choice but keep the `alias` there.

pkg/platform/packet/template.go
johananl: I'm going to leave it at that for the sake of getting to a working state ASAP. @invidian if you still have concerns I suggest you open a new PR to address them.

PR #540, opened 2020-06-03
Add error message for missing ipxe_script_url

### LINKED ISSUE #61: packet: print error/warning when os_version is specified and not ipxe_script_url
`os_version` only works for IPXE, print an error when this parameter is specified and `ipxe_script_url` is not. Currently there is not any error/warning, the installation continues installing a different version.

knrt10:
os_version only works for IPXE. Fixes #61

### DIFF
--- a/pkg/platform/packet/packet.go
+++ b/pkg/platform/packet/packet.go
@@ -388,6 +388,7 @@ func (c *config) checkValidConfig() hcl.Diagnostics {
 	diagnostics = append(diagnostics, c.checkNotEmptyWorkers()...)
 	diagnostics = append(diagnostics, c.checkWorkerPoolNamesUnique()...)
 	diagnostics = append(diagnostics, c.checkReservationIDs()...)
+	diagnostics = append(diagnostics, c.validateOSVersion()...)
 
 	if c.OIDC != nil {
 		_, diags := c.OIDC.ToKubeAPIServerFlags(c.clusterDomain())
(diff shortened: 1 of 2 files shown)


knrt10:
Yep, let me add unit tests and linter issue is fixes

pkg/platform/packet/packet_test.go
```diff
@@ -55,6 +55,25 @@ func TestCheckWorkerPoolNamesUniqueNotDup(t *testing.T) {
 	}
 }
 
+func TestCheckIPXEScriptURL(t *testing.T) {
```
invidian, line 58: Can we make table driven tests here to test all scenarios?
- OSVersion is NOT empty and IPXEScriptURL IS empty for worker: should give error
- OSVersion is NOT empty and IPXEScriptURL IS empty for controller: should give error
- OSVersion IS empty and IPXEScriptURL IS empty for worker: should be OK
- OSVersion IS empty and IPXEScriptURL IS empty for controller: should be OK
- OSVersion is NOT empty and IPXEScriptURL IS NOT empty for worker: should be OK
- OSVersion is NOT empty and IPXEScriptURL IS NOT empty for controller: should be OK

pkg/platform/packet/packet.go
```diff
@@ -451,6 +452,32 @@ func (c *config) checkReservationIDs() hcl.Diagnostics {
...
 
+// checkIPXEScriptURL checks that if os_version is specified then
+// ipxe_script_url should be present as os_version only works with IPXE.
+func (c *config) checkIPXEScriptURL() hcl.Diagnostics {
+	var diagnostics hcl.Diagnostics
+
+	if c.OSVersion != "" && c.IPXEScriptURL == "" {
+		diagnostics = append(diagnostics, &hcl.Diagnostic{
+			Severity: hcl.DiagError,
+			Summary:  "cannot find the ipxe_script_url",
```
johananl, line 463:
```suggestion
			Summary:  "ipxe_script_url not specified",
```

"cannot find" sounds strange to me. "not specified" makes it clearer that the user needed to specify something but didn't.

pkg/platform/packet/packet.go
```diff
@@ -451,6 +452,32 @@ func (c *config) checkReservationIDs() hcl.Diagnostics {
...
+			Detail:   "Cluster block should have ipxe_script_url if os_version is not empty",
```
johananl, line 464:
```suggestion
			Detail:   "ipxe_script_url must be specified if os_version is specified",
```

pkg/platform/packet/packet_test.go
knrt10: Sure

pkg/platform/packet/packet.go
```diff
@@ -451,6 +452,32 @@ func (c *config) checkReservationIDs() hcl.Diagnostics {
 	return diagnostics
 }
 
+// checkIPXEScriptURL checks that if os_version is specified then
```
johananl, line 455: IIUC the focus here is `os_version`, not `ipxe_script_url`. Therefore I would name the function `validateOSVersion`.

knrt10:
Fixed

PR #562, opened 2020-06-04
test/calico/metadata_access_test.go: add support for aws_edge platform

invidian:
Currently master pipeline fails because of this.


invidian:
:see_no_evil:

PR #565, opened 2020-06-04
examples: make Flatcar channels explicit

alban:
If Flatcar channels are not explicit, it is difficult for the user to find out how to change them. It also makes it less likely to forget to change it in the worker pool section.

Also fix a typo in the configuration.

### DIFF
--- a/docs/configuration-reference/platforms/aws.md
+++ b/docs/configuration-reference/platforms/aws.md
@@ -101,7 +101,7 @@ cluster "aws" {
 
   region = var.region
 
-  enable_aggreation = true
+  enable_aggregation = true
 
   disk_size = var.disk_size
 
--- a/examples/aws-production/cluster.lokocfg
+++ b/examples/aws-production/cluster.lokocfg
@@ -51,6 +51,9 @@ cluster "aws" {
   dns_zone_id      = var.route53_zone_id
   ssh_pubkeys      = var.ssh_public_keys
 
+  //os_channel       = "stable"
+  //os_version       = "current"
+
   oidc {
     issuer_url     = var.oidc_issuer_url
     client_id      = var.oidc_client_id
@@ -62,6 +65,9 @@ cluster "aws" {
     count         = var.workers_count
     instance_type = var.workers_type
     ssh_pubkeys   = var.ssh_public_keys
+
+    //os_channel    = "stable"
+    //os_version    = "current"
   }
 }
 
--- a/examples/aws-testing/cluster.lokocfg
+++ b/examples/aws-testing/cluster.lokocfg
@@ -30,10 +30,16 @@ cluster "aws" {
   dns_zone_id      = var.route53_zone_id
   ssh_pubkeys      = var.ssh_public_keys
 
+  //os_channel       = "stable"
+  //os_version       = "current"
+
   worker_pool "my-wp-name" {
     count         = var.workers_count
     instance_type = var.workers_type
     ssh_pubkeys   = var.ssh_public_keys
+
+    //os_channel    = "stable"
+    //os_version    = "current"
   }
 }
 
--- a/examples/packet-production/cluster.lokocfg
+++ b/examples/packet-production/cluster.lokocfg
@@ -57,6 +57,9 @@ cluster "packet" {
   cluster_name     = var.cluster_name
   controller_count = var.controllers_count
 
+  //os_channel       = "stable"
+
+
   dns {
     provider = "route53"
     zone     = var.dns_zone
@@ -80,6 +83,8 @@ cluster "packet" {
   worker_pool "pool-1" {
     count     = var.workers_count
     node_type = var.workers_type
+
+    //os_channel = "stable"
   }
 }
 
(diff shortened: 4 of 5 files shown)


examples/aws-production/cluster.lokocfg
```diff
@@ -50,6 +50,8 @@ cluster "aws" {
   dns_zone         = var.dns_zone
   dns_zone_id      = var.route53_zone_id
   ssh_pubkeys      = var.ssh_public_keys
+  os_channel       = "stable"
```
invidian, line 53: Hm, maybe we could leave them commented out, to show to the user, that those fields are optional? Example:
```suggestion
  //os_channel       = "stable"
```

And maybe add one empty line between required and optional parameters to easy distinguish between them.

invidian: I would also suggest doing the same change to all `examples` configuration files, so we are consistent.

docs/configuration-reference/platforms/aws.md
```diff
@@ -101,7 +101,7 @@ cluster "aws" {
 
   region = var.region
 
-  enable_aggreation = true
+  enable_aggregation = true
```
johananl, line 104: Looks like this code snippet was never tested.

examples/aws-production/cluster.lokocfg
invidian: @johananl what do you think?

alban: I can do that.

`os_channel` and `os_version` seems to be only supported on `aws`, `baremetal` and `packet` so I will only do the change there.

invidian: Thanks!

alban: Updated on aws and packet.

There are no examples on `baremetal`, so no need for changes here.

examples/aws-testing/cluster.lokocfg
```diff
@@ -29,11 +29,17 @@ cluster "aws" {
   dns_zone         = var.dns_zone
   dns_zone_id      = var.route53_zone_id
   ssh_pubkeys      = var.ssh_public_keys
+  os_channel       = "alpha"
```
invidian, line 32: This changes the testing configuration to use `alpha` version instead of previously used `stable`? Is this intentional? What is the reasoning for that?

Even for testing setup, I don't think we should be promoting use of `alpha` channel.

alban: Yes, I used `alpha` for testing setups. Why not? I think Flatcar alpha/beta should be tested by Lokomotive users before it reaches stable.

invidian: I think putting it here may result in many bug request reported to Lokomotive, which in fact will be caused by potential breakage in Flatcar. I'm not necessarily against it, I just think it should be a conscious decision. @iaguis what do you think?

examples/packet-production/cluster.lokocfg
```diff
@@ -57,6 +57,10 @@ cluster "packet" {
   cluster_name     = var.cluster_name
   controller_count = var.controllers_count
 
+  //os_channel       = "stable"
+  //os_version       = "current"
```
rata, line 61: I think `os_version` doesn't work on Packet. We used preinstalled Flatcar images on amd64, so we can't choose the version, we just run "whatever the preinstalled version is".

On ARM we use IPXE, IIRC, and we can choose that, though

invidian: @rata is right. On Packet, specifying `os_version` only will cause configuration error, as recently added in https://github.com/kinvolk/lokomotive/pull/540.

invidian: I think for Packet, we could just set commented out `os_channel` value then.

examples/aws-testing/cluster.lokocfg
alban: I can change it to `stable` for now until we have consensus. For me the main point was to make it obvious for users how to change the channels if they want to.

examples/packet-production/cluster.lokocfg
alban: I can changed os_version to "" since that's the only value accepted without `ipxe_script_url`. It is commented.

alban:
Updated.

examples/packet-production/cluster.lokocfg
```diff
@@ -57,6 +57,10 @@ cluster "packet" {
   cluster_name     = var.cluster_name
   controller_count = var.controllers_count
 
+  //os_channel       = "stable"
+  //os_version       = ""
```
invidian, line 61: I think having `os_version` here, commented out and with empty value does not bring much value to those example configs. We should either remove it or add commented out `ipxe_script_url` in my opinion, but considering that using `ipxe_script_url` is not very well documented, I'd just remove `os_version` for now.

alban: done

PR #566, opened 2020-06-04
Drop dependency on libcalico-go and improve TestHostEndpoints test

invidian:
While trying to upgrade Helm dependency to the latest version, we hit
the issue, that libcalico-go is still using old version of
'k8s.io/apimachinery', which prevents us from updating. On Calico slack,
we found, that libcalico-go shouldn't really be imported and used, as
the stability is not guaranteed.

We could upgrade apimachinery in libcalico-go and use the forked version
for the time being, but I think we should avoid that, as implementing
that, getting it merged upstream and released may take a long time.

As TestHostEndpoints test is the only place we use libcalico-go, the
easier way forward is to drop it as a dependency, which will unblock
updating Helm.

This commit changes this test to use Kubernetes REST client, which
mimics 'kubectl get --raw' instead of Calico client, to get HostEndpoint
objects from the cluster.

Also while testing this patch, I found that in my opinion this test is too easy to satisfy,
as any `HostEndpoint` object with right node name will make it pass, which means,
that if only the HostEndpoint for VPN interface exist, the test will pass, which should not be
the case, as our main concern is `bond0` interface, which is exposed to the internet, so this PR
also improves that, so we also check for `bond0` interface name.

CC @iaguis


invidian:
This is part of #484 and #545.

test/calico/calico_test.go
```diff
@@ -18,37 +18,60 @@
...
 
 	"github.com/kinvolk/lokomotive/test/components/util"
 )
 
 func TestHostEndpoints(t *testing.T) {
-	// Build Calico client.
-	cac := apiconfig.NewCalicoAPIConfig()
-	cac.Spec.DatastoreType = apiconfig.Kubernetes
-	cac.Spec.Kubeconfig = util.KubeconfigPath(t)
+	// Build rest client so we can do equivalent of 'kubectl get --raw'.
```
iaguis, line 35:
```suggestion
	// Build rest client so we can do the equivalent of 'kubectl get --raw'.
```

test/calico/calico_test.go
```diff
@@ -18,37 +18,60 @@
...
+
+	client, err := rest.RESTClientFor(config)
 	if err != nil {
-		t.Fatalf("failed creating Calico client: %v", err)
+		t.Fatalf("failed building rest client: %v", err)
 	}
 
-	// Build list of nodes which has associated HostEndpoint object.
-	hostEndpointList, err := c.HostEndpoints().List(context.TODO(), options.ListOptions{})
+	// This is minimal version of Calico HostEndpoint CRD object which we need to deserialize
```
iaguis, line 49:
```suggestion
	// This is minimal version of the Calico HostEndpoint CRD object which we need to deserialize
```

iaguis: Can we add some link to the upstream structure here?

test/calico/calico_test.go
```diff
@@ -18,37 +18,60 @@
...
 	if err != nil {
-		t.Fatalf("failed getting hostendpoint objects: %v", err)
+		t.Fatalf("failed getting HostEndpoint objects: %v", err)
+	}
+
+	if err := json.Unmarshal(response, &hostEndpoints); err != nil {
+		t.Fatalf("failed unmarshalling response: %v\n\n%s", err, string(response))
 	}
 
+	// Collect all received host endpoints into a map, so we can quickly look up if
```
iaguis, line 71:
```suggestion
	// Collect all received host endpoints into a map, so we can quickly look up if a
```

test/calico/calico_test.go
```diff
@@ -18,37 +18,60 @@
...
+	// specific object exists. We combine Node name and interface name to ensure, that
```
iaguis, line 72:
```suggestion
	// specific object exists. We combine Node name and interface name to ensure that
```

test/calico/calico_test.go
invidian: I'm not sure what I should link to. To https://github.com/projectcalico/libcalico-go/blob/master/lib/apis/crd.projectcalico.org/v1/hostendpoint_types.go#L27? + to K8s resource list? Maybe I just include shell command + returned json sample?

iaguis: I'd link to https://github.com/projectcalico/libcalico-go/blob/release-v3.14/lib/apis/v3/hostendpoint.go#L31

iaguis: or maybe without the line number since it's a branch...

invidian: Done.

iaguis:
Ooops, linter is not happy.

PR #577, opened 2020-06-05
New lokoctl command to generate a default config

alban:
It gives examples of configurations, allowing users to:
```
lokoctl template aws-production/cluster.lokocfg > mycluster.lokocfg
lokoctl cluster apply
```
instead of copy/paste from the Lokomotive documentation.

This patch also adds a basic syntax check of the configuration examples.
We don't have a command 'lokoctl validate' yet, but at least we check
syntax on the HCL level.


.github/workflows/terraform.yml
```diff
@@ -0,0 +1,31 @@
+name: 'Lokomotive Unit Tests'
```
invidian, line 1: Why is this part of this PR? We use Concourse for our CI. I'm not sure if we should be mixing multiple CI systems for our checks.

pkg/assets/assets_generate.go
```diff
@@ -25,6 +25,7 @@ func main() {
 	dirs := map[string]string{
 		"/lokomotive-kubernetes": "../../assets/lokomotive-kubernetes",
 		"/components":            "../../assets/components",
+		"/examples":              "../../examples",
```
invidian, line 28: I don't think `examples` should be part of the assets, as in the future, we want to use those assets to craft release archive, which will contain files required for installation. `examples` does not really belong there.

.github/workflows/terraform.yml
johananl, line 1: @alban just making sure you're aware we use Concourse as the CI platform for the project.

cli/cmd/template.go
```diff
@@ -0,0 +1,70 @@
...
+	"os"
+
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+
+	"github.com/kinvolk/lokomotive/pkg/assets"
+)
+
+var templateCmd = &cobra.Command{
+	Use:   "template",
```
johananl, line 30: I would consider naming the command `generate` or `init`. `template` makes me think of interpolating text files using placeholders, which isn't what this command does IIUC.

cli/cmd/template.go
```diff
@@ -0,0 +1,70 @@
...
+	if len(args) == 0 {
+		if err := assets.Assets.WalkFiles("/examples", walkList); err != nil {
+			ctxLogger.Fatalf("failed to walk assets: %s", err)
+		}
+	} else {
+		if err := assets.Assets.WalkFiles("/examples/"+args[0], walkPrint); err != nil {
+			ctxLogger.Fatalf("failed to walk assets: %s", err)
+		}
+	}
+}
```
invidian, line 70:
```suggestion
	if len(args) == 0 {
		if err := assets.Assets.WalkFiles("/examples", walkList); err != nil {
			ctxLogger.Fatalf("failed to walk assets: %s", err)
		}
		return
	}

	if err := assets.Assets.WalkFiles("/examples/"+args[0], walkPrint); err != nil {
		ctxLogger.Fatalf("failed to walk assets: %s", err)
	}
}
```

cli/cmd/template.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+	walkPrint := func(fileName string, fileInfo os.FileInfo, r io.ReadSeeker, err error) error {
+		if err == nil {
+			b, err := ioutil.ReadAll(r)
+			if err != nil {
+				return err
+			}
+			fmt.Printf("%s", b)
+		}
+		return err
```
invidian, line 59:
```suggestion
		if err != nil {
		        return err
		 }
		 
		b, err := ioutil.ReadAll(r)
		if err != nil {
			return err
		}
		
	        fmt.Printf("%s", b)

		return nil
```

.github/workflows/terraform.yml
alban: I thought we could use GitHub actions for compilation check and unit tests that does not require to create any external nodes (just the ubuntu VM used by GitHub action). And we could keep Concourse for all other tests that actually install Lokomotive with external nodes with additional cloud credentials.

johananl:
Thanks @alban!

I have two UX-related comments:

1. `template` feels like the wrong naming for generating sample config (see https://github.com/kinvolk/lokomotive/pull/577/files#r435877654).
1. I would rather have `lokoctl generate aws` or `lokoctl generate packet` etc. I don't have an opinion on how to "translate" these commands to file paths, however I feel this UX is nicer than specifying a "magic" file path which refers to some in-memory file such as `aws-production/cluster.lokocfg`. In case we want to segregate "test" from "prod" config, I would consider generating just *one* sample config and utilize comments to segregate prod from non-prod.

invidian:
Maybe something like `lokoctl config generate --platform aws` would make more sense?

.github/workflows/terraform.yml
johananl: I'm willing to discuss this, however I'd rather discuss the implications with the team first and not introduce a new element "implicitly" :-)

pkg/assets/assets_generate.go
alban: I am not sure I understand: the default configs generated by `lokoctl config generate --platform aws` (or other syntax) need to come from the assets and need to be in the release. Or do you suggest the repo should have both examples and default configs in separate directories?

johananl:
> Maybe something like `lokoctl config generate --platform aws` would make more sense?

Do we want to add a `config` "layer"? I'd think twice if `generate` is going to be the only knob there. `lokoctl health` for example isn't under `cluster` but is rather a top-level command. I'd be interested in @joaquimrocha 's thoughts here.

I don't have an opinion as to whether to put the platform in a subcommand or a flag.

invidian:
>Do we want to add a config "layer"? I'd think twice if generate is going to be the only knob there. lokoctl health for example isn't under cluster but is rather a top-level command. I'd be interested in @joaquimrocha 's thoughts here.

`config` layer could be then extended with `validate` subcommand, which would definitely be a valuable addition. But I'm also open for other suggestions.

iaguis:
I'm gonna close this since it's not clear we want it. We can revisit in the future.

PR #591, opened 2020-06-09
AppendTags improvements

invidian:
Moves from `util` package, which we shouldn't be using to just `pkg/platform` and adds tests for it.

Part of #290.


pkg/platform/platform.go
```diff
@@ -61,3 +62,18 @@ func GetPlatform(name string) (Platform, error) {
 	}
 	return platform, nil
 }
+
+// AppendTags append lokoctl-version tag to given tags map.
+func AppendTags(tags *map[string]string) {
```
iaguis, line 67: Now that we're at it, can we change it to a better name?

For example: `AppendVersionTag` or something similar?

invidian: Done.

pkg/platform/platform_test.go
```diff
@@ -0,0 +1,31 @@
...
+	"testing"
+)
+
+func TestAppendVersionTagUninitializedMap(t *testing.T) {
+	var f map[string]string
+
+	AppendVersionTag(&f)
+
+	if len(f) == 0 {
+		t.Fatalf("version tag should be appended to uninitialized map")
```
iaguis, line 13: Nit, for consistency

```suggestion
		t.Fatalf("should append version tag to uninitialized map")
```

invidian: Thanks.

pkg/platform/platform.go
```diff
@@ -61,3 +62,18 @@ func GetPlatform(name string) (Platform, error) {
 	}
 	return platform, nil
 }
+
+// AppendVersionTag append lokoctl-version tag to given tags map.
```
iaguis, line 66:
```suggestion
// AppendVersionTag appends the lokoctl-version tag to a given tags map.
```

pkg/platform/platform_test.go
```diff
@@ -0,0 +1,33 @@
+package platform_test
```
iaguis, line 1: Can you add a copyright header?

PR #593, opened 2020-06-09
formatting: remove extra space from markdown table

ipochi:
Removes the extra space
```
diff --git a/docs/configuration-reference/platforms/packet.md b/docs/configuration-reference/platforms/packet.md
index 3e946796..7e84ce26 100644
--- a/docs/configuration-reference/platforms/packet.md
+++ b/docs/configuration-reference/platforms/packet.md
@@ -188,7 +188,7 @@ node_type = var.custom_default_worker_type
 | `cluster_name`                        | Name of the cluster.                                                                                                                                                          | -               | true     |
 | `tags`                                | List of tags that will be propagated to master nodes.                                                                                                                         | -               | false    |
 | `controller_count`                    | Number of controller nodes.                                                                                                                                                   | 1               | false    |
-| `controller_type`                     | Packet instance type for controllers.                                                                                                                                         | "t1.small.x86"   | false    |
+| `controller_type`                     | Packet instance type for controllers.                                                                                                                                         | "t1.small.x86"  | false    |
 | `dns`                                 | DNS configuration block.                                                                                                                                                      | -               | true     |
 | `dns.zone`                            | A DNS zone to use for the cluster. The following format is used for cluster-related DNS records: `<record>.<cluster_name>.<dns_zone>`                                         | -               | true     |
 | `dns.provider`                        | DNS provider to use for the cluster. Valid values: `cloudflare`, `route53`, `manual`.                                                                                         | -               | true     |
```


johananl:
Maybe we should use tabs instead of spaces :trollface:

Issue #563, opened 2020-06-04
Rename packet-arm to packet_arm to be consistent with aws_edge


invidian:
As `-` cannot be used as in Go build tags, we changed `aws-edge` to `aws_edge`. `packet-arm` should follow for consistency. Also CI configuration needs changing when this is done.

PR #600, opened 2020-06-09
Add packet_arm as alias to packet-arm

invidian:
So CI can be configured to use packet_arm instead of packet-arm
gracefully.

This is to make packet-arm consistent with aws_edge. As - cannot be used
as in Go build tags, we changed aws-edge to aws_edge.

Part of #563.

### DIFF
--- a/test/calico/metadata_access_test.go
+++ b/test/calico/metadata_access_test.go
@@ -89,7 +89,7 @@ func TestNoMetadataAccessRandomPod(t *testing.T) { //nolint:funlen
 	platform := os.Getenv("PLATFORM")
 
 	switch platform {
-	case testutil.PlatformPacket, testutil.PlatformPacketARM:
+	case testutil.PlatformPacket, testutil.PlatformPacketARM, "packet-arm":
 		metadataAddress = "https://metadata.packet.net/metadata"
 	case testutil.PlatformAWS, testutil.PlatformAWSEdge:
 		metadataAddress = "http://169.254.169.254/latest/meta-data"
--- a/ci/packet-arm/packet_arm-cluster.lokocfg.envsubst
+++ b/ci/packet-arm/packet_arm-cluster.lokocfg.envsubst
@@ -1,0 +1,1 @@
+packet-arm-cluster.lokocfg.envsubst
--- a/ci/packet_arm
+++ b/ci/packet_arm
@@ -1,0 +1,1 @@
+packet-arm/
--- a/test/components/util/util.go
+++ b/test/components/util/util.go
@@ -337,7 +337,7 @@ const (
 	PlatformPacket = "packet"
 
 	// PlatformPacketARM is for Packet on ARM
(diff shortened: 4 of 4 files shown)


ipochi:
Question: I want to understand why we need to have  `packet_arm` as alias to `packet-arm` ? 
Why not remove `packet-arm` and only have `packet_arm` ?

invidian:
>Question: I want to understand why we need to have packet_arm as alias to packet-arm ?
Why not remove packet-arm and only have packet_arm ?

So CI can be configured to use `packet_arm` instead of `packet-arm`
**gracefully** :)

ipochi:
> > Question: I want to understand why we need to have packet_arm as alias to packet-arm ?
> > Why not remove packet-arm and only have packet_arm ?
> 
> So CI can be configured to use `packet_arm` instead of `packet-arm`
> **gracefully** :)

If we had `packet_arm`only, then how it would affect CI ?

invidian:
Then existing CI configuration would try to use `packet-arm` and it wouldn't find the cluster configuration and tests would fail with platform being not supported.

ipochi:
> Then existing CI configuration would try to use `packet-arm` and it wouldn't find the cluster configuration and tests would fail with platform being not supported.

Shouldn't we change the CI configuration to use `packet_arm` rather than adding an alias ?
Another way to put is , Is changing the CI configuration not so simple that we have to add an alias ?

invidian:
I could do it in one step, but then there is a time window, when things may break. This way, the change won't break anything.

ipochi:
> I could do it in one step, but then there is a time window, when things may break. This way, the change won't break anything.

I didn't understand the first line part of the sentence.

Time window, where things may break ?  How big of a time window. 
Now is as good a time to break things :laughing:

invidian:
If following event occurs:
1. Change from `packet-arm` to `packet_arm` gets merged.
2. One creates new PR, which will always be failing for ARM, until step 3. is complete
3. CI configuration is deployed.

So, in time window between events 1. and 3., things will be broken and I don't like that.

iaguis:
If we then get rid of the symlink after everything is settled I'm fine with this.

test/calico/metadata_access_test.go
```diff
@@ -89,7 +89,7 @@ func TestNoMetadataAccessRandomPod(t *testing.T) { //nolint:funlen
 	platform := os.Getenv("PLATFORM")
 
 	switch platform {
-	case testutil.PlatformPacket, testutil.PlatformPacketARM:
+	case testutil.PlatformPacket, testutil.PlatformPacketARM, "packet-arm":
```
iaguis, line 92: We can add a TODO to remove this once the CI config is updated. But if you're doing it right after this is merged don't bother with the TODO.

invidian: Yeah, will do it right after :)

PR #602, opened 2020-06-09
cli/cmd: rename --kubeconfig flag to --kubeconfig-file

### LINKED ISSUE #595: KUBECONFIG env variable takes precedence over cluster configuration while selecting kubeconfig
Due to unfortunate fact, that we have `--kubeconfig` flag and we use `viper.AutomaticEnv()`, it causes `KUBECONFIG` environment variable to be interpreted as `--kubeconfig` flag, which causes, that even if user is in the working directory with cluster configuration, the component will be installed/uninstalled to the cluster from `KUBECONFIG` environment variable. This is incorrect behavior and dangerous.

I suggest, as a workaround, we rename `--kubeconfig` flag to `--kubeconfig-file`. For long term solution, we should re-evaluate how we use viper and try to avoid using globals for it, as it causes side-effects, like described here.

Possibly related: #105

invidian:
This PR renames `--kubeconfig` flag to `--kubeconfig-file`  to avoid automatic reading of `KUBECONFIG` environment variable by viper and adds explicit read of `KUBECONFIG` variable **after** cluster configuration in the hierarchy, when selecting `kubeconfig` file to use.

For the time being, there is no tests for this patch, as testing that is not trivial to implement (involves creating lokocfg configuration file in temporary directory, testing private method, and testing viper flags parsing.

Closes #595.

### DIFF
--- a/cli/cmd/utils_internal_test.go
+++ b/cli/cmd/utils_internal_test.go
@@ -1,0 +1,204 @@
+// Copyright 2020 The Lokomotive Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cmd
+
+import (
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/spf13/viper"
+)
+
+type kubeconfigSources struct {
+	flag       string
+	env        string
+	configFile string
+}
+
+func prepareKubeconfigSource(t *testing.T, k *kubeconfigSources) {
+	// Ensure viper flag is NOT empty.
+	viper.Set(kubeconfigFlag, k.flag)
+
+	if k.env == "" {
+		// Ensure KUBECONFIG is not set.
+		if err := os.Unsetenv(kubeconfigEnvVariable); err != nil {
+			t.Fatalf("unsetting %q environment variable: %v", kubeconfigEnvVariable, err)
+		}
+	}
+
+	if k.env != "" {
+		// Ensure KUBECONFIG IS set.
+		if err := os.Setenv(kubeconfigEnvVariable, k.env); err != nil {
+			t.Fatalf("setting %q environment variable: %v", kubeconfigEnvVariable, err)
+		}
+	}
+
+	// Ensure there is no lokocfg configuration in working directory.
+	tmpDir, err := ioutil.TempDir("", "lokoctl-tests-")
+	if err != nil {
+		t.Fatalf("creating tmp dir: %v", err)
+	}
+
+	t.Cleanup(func() {
+		if err := os.RemoveAll(tmpDir); err != nil {
+			t.Logf("removing temp dir %q: %v", tmpDir, err)
+		}
+	})
+
+	if err := os.Chdir(tmpDir); err != nil {
+		t.Fatalf("changing working directory to %q: %v", tmpDir, err)
+	}
+
+	if k.configFile != "" {
+		path := filepath.Join(tmpDir, "cluster.lokocfg")
+		if err := ioutil.WriteFile(path, []byte(k.configFile), 0600); err != nil {
+			t.Fatalf("writing file %q: %v", path, err)
+		}
+	}
+}
+
+func TestGetKubeconfigFlag(t *testing.T) {
+	expectedPath := "/foo"
+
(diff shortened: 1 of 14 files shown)


invidian:
>For the time being, there is no tests for this patch, as testing that is not trivial to implement (involves creating lokocfg configuration file in temporary directory, testing private method, and testing viper flags parsing.

It turns out, I introduced some other bugs to the implementation, so I decided to write tests for it, even though they are not great.

cli/cmd/utils_internal_test.go
```diff
@@ -0,0 +1,184 @@
+// Copyright 2020 The Lokomotive Authors
```
iaguis, line 1: I'd like a test where both the KUBECONFIG flag is set and the lokocfg file is present to make sure it takes the file defined in the lokocfg. This is the most dangerous case IMO.

ipochi:
@invidian @iaguis 

1. I think that we should not use `~/.kube/config` to fall back.

2. Having `KUBECONFIG_FILE` environment variable gives an irritating behavior to the user as `lokoctl` is not a replacement for `kubectl`. 

Even after installing the component by setting the `KUBECONFIG_FILE` environment variable, a user still interfaces the cluster with `kubectl` or in some cases `helm`. In such a scenario `kubectl` or `helm` is pointing to `KUBECONFIG` env variable or `~/.kube/config` and lokoctl to `KUBECONFIG_FILE`.

Consider the the simplest scenario :

1. Installed a component X  with `KUBECONFIG_FILE` env var set.
2. Need to get some logs of the component via `kubectl`, now I need to set the `KUBECONFIG` to the Lokomotive managed cluster in order to perform operations.

IMO, we should just have a flag `--kubeconfig-file` in lokoctl without `KUBECONFIG_FILE` env var.

Can we perhaps use [Extensions](https://github.com/kubernetes/client-go/blob/3ab27431ef50d8ae77afe66593d04d5616cbd448/tools/clientcmd/api/types.go#L98 ) (not sure how extensions is used, I've never seen it in a kubeconfig file) in the kubeconfig file to specify the additional cluster information as managed by Lokomotive. If the `KUBECONFIG` or current context is not referring to the same then we print out a message to the user for confirmation before proceeding ?

invidian:
I agree with Imran, that having this env variable is not really desired. Unfortunately, this is a side effect of the cobra. I'm not sure if we can disable env variable reading for just one flag.

iaguis:
Just as I reminder, my main focus for this PR is fixing the following case (which happened to me twice with bad consequences):

1. A user runs `lokoctl component apply` from their cluster directory.
2. The user has `$KUBECONFIG` set to another cluster, probably not even managed by Lokomotive.
3. The components are applied to the cluster defined in `$KUBECONFIG` instead of the Lokomotive cluster (!)

I agree the `$KUBECONFIG_FILE` env var doesn't make sense. If it's not possible to disable it from cobra/viper I think just keeping it undocumented should be fine.

Regarding falling back to `~/.kube/config` if there's no `$KUBECONFIG` and the user is not in a Lokomotive cluster directory I'm on the fence.

On the one hand `kubectl` falls back to `~/.kube/config` so it makes sense if we do the same and commands to `kubectl` and `lokoctl` would apply to the same cluster. On the other hand this behavior might be dangerous and as a caution we might want to forbid it and ask the user to set `$KUBECONFIG`.

Since falling back to `~/.kube/config` is already happening now, we can keep it and decide later if we want to change this behavior.

iaguis:
Addressed my review since @invidian is out.

cli/cmd/utils_internal_test.go
```diff
@@ -0,0 +1,221 @@
...
+			t.Fatalf("writing file %q: %v", path, err)
+		}
+	}
+}
+
+func TestGetKubeconfigFlag(t *testing.T) {
+	expectedPath := "/foo"
+
+	k := &kubeconfigSources{
+		flag: expectedPath,
```
invidian, line 78: I'd add configFile, and env here, to ensure that file takes precedence over everything else.

cli/cmd/utils_internal_test.go
```diff
@@ -0,0 +1,221 @@
...
+	if kubeconfig != expectedPath {
+		t.Fatalf("expected %q, got %q", expectedPath, kubeconfig)
+	}
+}
+
+func TestGetKubeconfigConfigFile(t *testing.T) {
+	expectedPath := assetsKubeconfig("/foo")
+
+	k := &kubeconfigSources{
+		configFile: `cluster "packet" {
```
invidian, line 97: Ditto env

cli/cmd/utils_internal_test.go
```diff
@@ -0,0 +1,221 @@
...
+	if err != nil {
+		t.Fatalf("getting kubeconfig: %v", err)
+	}
+
+	if kubeconfig != expectedPath {
+		t.Fatalf("expected %q, got %q", expectedPath, kubeconfig)
+	}
+}
+
+func TestGetKubeconfigConfigAndEnvVariable(t *testing.T) {
```
invidian, line 169: With this test, TestGetKubeconfigConfigFile becomes redundant.

PR #628, opened 2020-06-15
pkg/k8sutil: change NewClientset to load kubeconfig from file content, not from path

invidian:
As a first step towards resolving #608, this PR changes `NewClientset` to create Kubernetes Clientset from `kubeconfig` file **content**, instead of from file **path**. The clients are changes to manually load the file content, which should simplify changing it to also accept the file content as an argument in the upcoming patches.

### DIFF
--- a/cli/cmd/cluster-apply.go
+++ b/cli/cmd/cluster-apply.go
@@ -125,12 +125,12 @@ func runClusterApply(cmd *cobra.Command, args []string) {
 }
 
 func verifyCluster(kubeconfigPath string, expectedNodes int) error {
-	client, err := k8sutil.NewClientsetFromFile(kubeconfigPath)
+	cs, err := k8sutil.NewClientsetFromFile(kubeconfigPath)
 	if err != nil {
 		return errors.Wrapf(err, "failed to set up clientset")
 	}
 
-	cluster, err := lokomotive.NewCluster(client, expectedNodes)
+	cluster, err := lokomotive.NewCluster(cs, expectedNodes)
 	if err != nil {
 		return errors.Wrapf(err, "failed to set up cluster client")
 	}
--- a/cli/cmd/cluster-apply.go
+++ b/cli/cmd/cluster-apply.go
@@ -16,6 +16,7 @@ package cmd
 
 import (
 	"fmt"
+	"io/ioutil"
 
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
@@ -125,7 +126,12 @@ func runClusterApply(cmd *cobra.Command, args []string) {
 }
 
 func verifyCluster(kubeconfigPath string, expectedNodes int) error {
-	cs, err := k8sutil.NewClientsetFromFile(kubeconfigPath)
+	kubeconfig, err := ioutil.ReadFile(kubeconfigPath) // #nosec G304
+	if err != nil {
+		return errors.Wrapf(err, "failed to read kubeconfig file")
+	}
+
+	cs, err := k8sutil.NewClientset(kubeconfig)
 	if err != nil {
 		return errors.Wrapf(err, "failed to set up clientset")
 	}
--- a/cli/cmd/cluster-apply.go
+++ b/cli/cmd/cluster-apply.go
@@ -125,7 +125,7 @@ func runClusterApply(cmd *cobra.Command, args []string) {
 }
 
 func verifyCluster(kubeconfigPath string, expectedNodes int) error {
-	client, err := k8sutil.NewClientset(kubeconfigPath)
+	client, err := k8sutil.NewClientsetFromFile(kubeconfigPath)
 	if err != nil {
 		return errors.Wrapf(err, "failed to set up clientset")
 	}
--- a/cli/cmd/component-delete.go
+++ b/cli/cmd/component-delete.go
@@ -167,7 +167,7 @@ func deleteHelmRelease(c components.Component, kubeconfig string, deleteNSBool b
 }
 
 func deleteNS(ns string, kubeconfig string) error {
-	cs, err := k8sutil.NewClientset(kubeconfig)
+	cs, err := k8sutil.NewClientsetFromFile(kubeconfig)
 	if err != nil {
 		return err
 	}
--- a/cli/cmd/component-delete.go
+++ b/cli/cmd/component-delete.go
@@ -17,6 +17,7 @@ package cmd
 import (
 	"context"
 	"fmt"
+	"io/ioutil"
 	"strings"
 
 	log "github.com/sirupsen/logrus"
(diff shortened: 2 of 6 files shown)


invidian:
Rebased to resolve conflicts.

pkg/components/util/install.go
```diff
@@ -108,6 +81,7 @@ func install(helmAction *helmAction, namespace string) error {
 	install := action.NewInstall(helmAction.actionConfig)
 	install.ReleaseName = helmAction.releaseName
 	install.Namespace = namespace
+	install.CreateNamespace = true
```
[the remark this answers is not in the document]
invidian: This has been added "back" (as it got removed in v3.0.0) in v3.2.0: https://github.com/helm/helm/commit/694cf61aacba6bc4ed59a65d23acd20f6c8b95bd, that's why we needed this code in the past.

cli/cmd/cluster-apply.go
```diff
@@ -125,7 +125,7 @@ func runClusterApply(cmd *cobra.Command, args []string) {
 }
 
 func verifyCluster(kubeconfigPath string, expectedNodes int) error {
-	client, err := k8sutil.NewClientset(kubeconfigPath)
+	client, err := k8sutil.NewClientsetFromFile(kubeconfigPath)
```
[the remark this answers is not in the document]
invidian: Right. Sorry about that, I wrote complete nonsense. I rephrased now, it should be better :)

invidian:
I applied your suggestions, thanks for reviewing @iaguis :pray:

pkg/k8sutil/client.go
```diff
@@ -20,6 +20,8 @@ import (
 	"k8s.io/client-go/tools/clientcmd"
 )
 
+// NewClientsetFromFile creates a new Kubernetes Client set object from the given
```
ipochi: nit:  the first and this commit should be merged into a single commit.

pkg/components/util/install.go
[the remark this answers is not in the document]
ipochi: @invidian If Helm is taking care of creating namespace, then I'd wait for this merge for the PR #646 as then we avoid create step of namespace.

nit: seems this change doesn't belong in this PR.

comments like 
```
This has been added "back" (as it got removed in v3.0.0) in v3.2.0: helm/helm@694cf61, 
that's why we needed this code in the past.
```

will be lost , if you can put the same comment in a separate PR that would be great.

cli/cmd/cluster-apply.go
```diff
@@ -125,7 +126,12 @@ func runClusterApply(cmd *cobra.Command, args []string) {
 }
 
 func verifyCluster(kubeconfigPath string, expectedNodes int) error {
-	cs, err := k8sutil.NewClientsetFromFile(kubeconfigPath)
+	kubeconfig, err := ioutil.ReadFile(kubeconfigPath) // #nosec G304
```
ipochi, line 129: perhaps reading of the kubeconfig file as a function in  `internal` package rather than having the same code across multiple places ?

cli/cmd/component-delete.go
```diff
@@ -167,7 +168,12 @@ func deleteHelmRelease(c components.Component, kubeconfig string, deleteNSBool b
 }
 
 func deleteNS(ns string, kubeconfig string) error {
-	cs, err := k8sutil.NewClientsetFromFile(kubeconfig)
+	kubeconfigContent, err := ioutil.ReadFile(kubeconfig) // #nosec G304
```
ipochi, line 171: ditto: as function in `internal` package.

cli/cmd/health.go
```diff
@@ -47,7 +48,13 @@ func runHealth(cmd *cobra.Command, args []string) {
 	if err != nil {
 		contextLogger.Fatalf("Error in finding kubeconfig file: %s", err)
 	}
-	cs, err := k8sutil.NewClientsetFromFile(kubeconfig)
+
+	kubeconfigContent, err := ioutil.ReadFile(kubeconfig) // #nosec G304
```
ipochi, line 52: ditto: as function in `internal` package.

pkg/k8sutil/client.go
```diff
@@ -22,22 +22,6 @@ import (
 	"k8s.io/client-go/tools/clientcmd"
 )
 
-// NewClientsetFromFile creates a new Kubernetes Client set object from the given
```
ipochi: is this commit [f74c7f7](https://github.com/kinvolk/lokomotive/pull/628/commits/f74c7f7c1d008d54c527cbe0f09b6c7514436ed2) before or after the current commit ? I guess after rebase github messes up the commit order.

If its after, then I'd change it to before to truly reflect the commit message
```
NewClientsetFromFile is now deprecated, as we ...
```

pkg/k8sutil/client.go
invidian: Nah, I'm not sure about that. Any specific arguments why to merge them instead of keeping them separate? It's 2 separate tasks, one is to rename the function, so refactoring (and ignoring . This one is fixing the linter. Perhaps one reason to merge them would be to make linter succeed per commit, though it's not a big deal as it's only the linter.

pkg/components/util/install.go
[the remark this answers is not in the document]
invidian: True, this change does not really belong to this PR. I'll remove it. There is also an interest to get that in in #686. Also, if we let Helm to create the namespace, applying labels must happen **after** Helm release is installed. This may actually introduce some nasty bugs I imaging. Consider following scenario:

Perform Helm install action, with `Wait: true`. Installed manifests require bi-namespace NetworkPolicy objects to be created. One in component's namespace and other in component interacting with this namespace. The installation will never converge, as the namespace labels will be missing, so the network policies will not be correctly applied.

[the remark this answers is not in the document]
surajssd: Is this hypothesis or backed by empirical data?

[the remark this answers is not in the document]
invidian: >Is this hypothesis or backed by empirical data?

I'd say neither. It's a logical conclusion. But yeah, more like hypothesis, as I didn't do tests to confirm that.

cli/cmd/cluster-apply.go
invidian: The duplication is intentional, so "target" function (`k8sutil.NewClientset`) is already being called. Next PR will be changing the parameter types we pass around, so `ioutil.ReadFile` call will be removed. Adding function to de-duplicate has little sense here.

See #631 for more details.

pkg/k8sutil/client.go
invidian: I merged it anyway, so keep linter happy even between the commits

pkg/k8sutil/client.go
invidian: First goes `cli/cmd: use NewClientset to construct Kubernetes clientset`, then `pkg/k8sutil: remove unused NewClientsetFromFile`. I'm not sure what change are you suggesting.

ipochi: Never mind, I meant Github was showing the wrong order to me

PR #631, opened 2020-06-16
Pass kubeconfig content around rather than file path

### LINKED ISSUE #623: Use kubeconfig content rather than kubeconfig path between functions
To resolve #608, we should change the way we use `kubeconfig` in our code from using it as a path to use the content of this file, to decouple it from requiring a file on the disk, as then we would be able to pull this content from any place, e.g. Terraform output/state.

Ideally, `kubeconfig` content should become a **context** for all operations we use it for, perhaps in a **cluster** context, so all operations don't need to accept it explicitly. However, this might be out of scope for this task. See also #603.

invidian:
**This PR includes commits from #628, please review this one first and don't leave commit comments for this PR here.**

This PR changes a lot of the code to accept kubeconfig content rather than kubeconfig path, so we will be able to read it from Terraform state/output in the future, which is needed for #608. It also shows, how poorly the code is structured, as such change requite so much code changes. We should make most of this function to execute in the scope/context (so make them have a receiver), which already has access to the kubeconfig content, as the kubeconfig content is almost constant in calls of this functions.

Please review the PR commit by commit, as each of them contains small, logical and independent changes, described a bit in commit message.

Now that reading kubeconfig file is in single place, we can implement reading it from Terraform state.

Closes #623

Part of #608

### DIFF
--- a/cli/cmd/cluster-apply.go
+++ b/cli/cmd/cluster-apply.go
@@ -81,13 +81,14 @@ func runClusterApply(cmd *cobra.Command, args []string) {
 	fmt.Printf("\nYour configurations are stored in %s\n", assetDir)
 
 	kubeconfigPath := assetsKubeconfig(assetDir)
-	if err := verifyCluster(kubeconfigPath, p.Meta().ExpectedNodes); err != nil {
-		ctxLogger.Fatalf("Verify cluster: %v", err)
-	}
 
 	kubeconfigContent, err := ioutil.ReadFile(kubeconfigPath) // #nosec G304
 	if err != nil {
 		ctxLogger.Fatalf("Failed to read kubeconfig file: %q: %v", kubeconfigPath, err)
+	}
+
+	if err := verifyCluster(kubeconfigContent, p.Meta().ExpectedNodes); err != nil {
+		ctxLogger.Fatalf("Verify cluster: %v", err)
 	}
 
 	// Do controlplane upgrades only if cluster already exists and it is not a managed platform.
@@ -130,12 +131,7 @@ func runClusterApply(cmd *cobra.Command, args []string) {
 	}
 }
 
-func verifyCluster(kubeconfigPath string, expectedNodes int) error {
-	kubeconfig, err := ioutil.ReadFile(kubeconfigPath) // #nosec G304
-	if err != nil {
-		return errors.Wrapf(err, "failed to read kubeconfig file")
-	}
-
+func verifyCluster(kubeconfig []byte, expectedNodes int) error {
 	cs, err := k8sutil.NewClientset(kubeconfig)
 	if err != nil {
 		return errors.Wrapf(err, "failed to set up clientset")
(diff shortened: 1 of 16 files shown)


cli/cmd/cluster-apply.go
```diff
@@ -89,10 +93,10 @@ func runClusterApply(cmd *cobra.Command, args []string) {
 		fmt.Printf("\nEnsuring that cluster controlplane is up to date.\n")
 
 		cu := controlplaneUpdater{
-			kubeconfigPath: kubeconfigPath,
-			assetDir:       assetDir,
-			ctxLogger:      *ctxLogger,
-			ex:             *ex,
+			kubeconfig: kubeconfig,
```
knrt10, line 96: You can just do

```go
cu := controlplaneUpdater{
		kubeconfig,
		assetDir,
		ctxLogger:      *ctxLogger,
		ex:             *ex,
}
```

invidian: Right, but I think we should prefer used convention here, to avoid implicit coupling on struct fields order. Also, such change would be unrelated to this already large PR.

knrt10: Ok, sounds reasonable.

invidian:
After this PR, last step is to read kubeconfig file content from Terraform state. This requires:
- changing `getKubeconfig` to also read from Terraform output
- if one runs `lokoctl health` with cluster configuration, Terraform files needs to be dumped and Terraform binary must be present on the system.

cli/cmd/component-delete.go
```diff
@@ -16,6 +16,7 @@ package cmd
 
 import (
```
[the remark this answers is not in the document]
invidian: Fixed, thanks :+1:

invidian:
>Commits that move code from from cli to util package such as deleteComponent.
Its great that you have taken this out of cli package but I have reservations about this in this PR.

Do you mean that I move `deleteHelmRelease` to `pkg/components/util` package, rather than to `/pkg/components` for example? I'd say this is on purpose for now, to keep all Helm-related functionality in one place. Later on, we should move all of that to `pkg/components` or to `pkg/components/helm`, but I'd do that separately.

invidian:
>Looks great @invidian. And by previous comment for @ipochi I think he meant moving code out of cli package might not be suitable for this PR.

Ah, that make sense. If you insist, that this shouldn't be part of this PR, I'll remove it, but I don't see it doing much harm in the existing state.

knrt10:
Yea, I would say no need to remove it now, this looks completely fine for now

ipochi:
> > Looks great @invidian. And by previous comment for @ipochi I think he meant moving code out of cli package might not be suitable for this PR.
> 
> Ah, that make sense. If you insist, that this shouldn't be part of this PR, I'll remove it, but I don't see it doing much harm in the existing state.

If its a small thing to do, then I'd like that to keep them separate. However if its going to be bit more than tiny overhead, its fine as is.

invidian:
>If its a small thing to do, then I'd like that to keep them separate. However if its going to be bit more than tiny overhead, its fine as is.

> I'll let you make the call.

Unfortunately, it won't be a tiny thing, as moved function is changed afterwards, so I'll just keep it as it is and merge. Thanks!

PR #638, opened 2020-06-18
terraform/aws/workers: fix upgrades from v0.1.0

iaguis:
This changes LoadBalancer names and reverts ASG names to the old scheme so upgrades from v0.1.0 are smooth.

In a following release we can change the ASG name again and we can look into using something like `create_before_destroy` to avoid downtime.


iaguis:
I did test this several times, so I'm fine with merging it.

PR #642, opened 2020-06-18
packet/nodes: switch back to baremetal_0 as default node type

iaguis:
Otherwise upgrading from v0.1.0 will cause all nodes to be recreated.


invidian:
I think this PR was not necessary. Instead, users wishing to **not** migrate their worker pools to new type should define the value explicitly and we should recommend users to re-create the pools.

iaguis:
> I think this PR was not necessary. Instead, users wishing to **not** migrate their worker pools to new type should define the value explicitly and we should recommend users to re-create the pools.

That's a fair point. We could update this and mention in the release notes the change and recommend to define explicitly the instance type if using the default one.

PR #651, opened 2020-06-22
Add Type column in Attribute reference

### LINKED ISSUE #621: Add Type column in Attribute reference for all components
Sometimes there is no default value so, it's difficult to understand the type of argument i.e whether it is string, bool, map etc. For reference see [contour.md#attribute-reference](https://github.com/kinvolk/lokomotive/blob/master/docs/configuration-reference/components/contour.md#attribute-reference).

This issue will require to add a new column `type` for all components in `Attribute reference` which will specify the type of arguments.

knrt10:
Fixes #621 


docs/configuration-reference/components/contour.md
```diff
@@ -67,13 +67,14 @@ Table of all the arguments accepted by the component.
...
+|---------------------|---------------------------------------------------------------------------------------------------------|:--------------:|:---------------------------------------------------------------------------------------------------------------|:--------:|
+| `enable_monitoring` | Create Prometheus Operator configs to scrape Contour and Envoy metrics. Also deploys Grafana Dashboard. |     false      | bool                                                                                                           |  false   |
+| `ingress_hosts`     | [ExternalDNS component](external-dns.md) creates DNS entries from the values provided.                  |       ""       | list(string)                                                                                                   |  false   |
+| `node_affinity`     | Node affinity for deploying the operator pod and envoy daemonset.                                       |       -        | list(object({key = string, operator = string, values = list(string)}))                                         |  false   |
```
invidian, line 74: In my opinion, if we have a list or map of blocks, those should get separate table with parameters. It will be much more readable then.

johananl: I don't want to sidetrack the discussion too much, but I've been wanting to re-evaluate using tables for a while now, so if we figure tables are causing trouble with this PR, maybe it's time to question this decision.

Rationale:

I find tables too limiting for a config reference as there isn't a lot of space in a table cell for *really* explaining what some knob is/does and also there is no convenient way to add note blocks or code snippets to help explain some point.

I'd rather have a more flexible format which allows us to write as much text as we want to explain a knob. I remind us that a config reference should be *exhaustive* rather than concise, i.e. "if it's not in the reference, it doesn't exist". I don't have a specific format in mind but am happy to propose one if folks are open to that.

johananl: A concrete example:

The description for `cluster_name` is `Name of the cluster.`. IMO this description is nearly useless as it doesn't tell the user what would happen if they put one value or the other in this field. I'd rather include a longer text which explains how this value is used by `lokoctl`, however I'd need some space to do that if I don't want to have a really tall table cell with lots of line breaks.

invidian: @johananl what are you suggesting we should do then? Should we keep it as it is? Should we not add type for lists and maps for now?

johananl: I have nothing against adding a type column, I wrote the above since it sounded to me like we were discussing the overall structure of the document and wanted to see if it's an opportunity to rethink our usage of tables.

I'm leaning towards a paragraph-based format with a ToC at the top, a bit like [this](https://rancher.com/docs/rke/latest/en/config-options/) maybe. This format allows a lot of flexibility since you can provide full-text paragraphs for each knob, there is room for notes and sample snippets etc. But this is likely out of scope for this PR. Again, I wanted to mainly give this as something to think about *if* we are re-evaluating the overall structure here.

iaguis: To move forward this how about leaving it like it is now (even if it's not very readable) and creating an issue to switch to a paragraph-based format?

knrt10: I am comfortable with this. So should I close this PR?

iaguis: I'd merge this PR if it looks OK (it adds valuable information) and then we can work on moving to a paragraph-based format. I'll review this.

knrt10: Cool

docs/configuration-reference/backend/local.md
```diff
@@ -36,8 +36,9 @@ backend "local" {
 
 Default backend is local.
 
-| Argument                    | Description                                                  | Default | Required |
-|-----------------------------|--------------------------------------------------------------|:-------:|:--------:|
-| `backend.local`             | Local backend configuration block.                           | -       | false    |
-| `backend.local.path`        | Location where Lokomotive stores the cluster state.          | -       | false    |
+| Argument             | Description                                         | Default |  Type  | Required |
+|----------------------|-----------------------------------------------------|:-------:|:------:|:--------:|
+| `backend.local`      | Local backend configuration block.                  |    -    | block  |  false   |
```
iaguis, line 41: We're using `object` for this later, we should be consistent. I'd keep `object`.

Same for other occurrences of `block`.

docs/configuration-reference/components/metallb.md
```diff
@@ -84,14 +83,15 @@ Table of all the arguments accepted by the component.
...
-| `service_monitor`           | Create ServiceMonitor for Prometheus to scrape MetalLB metrics.                            | false   | false    |
+| Argument                    | Description                                                                                | Default | Type                                                                                                           | Required |
+|-----------------------------|--------------------------------------------------------------------------------------------|:-------:|:---------------------------------------------------------------------------------------------------------------|:--------:|
+| `address_pools`             | A map which allows specifying one or more CIDRs which MetalLB can use to expose services.  |    -    | object({default = list(string), special_addresses = list(string)})                                             |   true   |
+| `controller_node_selectors` | A map with specific labels to run MetalLB controller pods selectively on a group of nodes. |    -    | map(string)                                                                                                    |  false   |
```
iaguis, line 89: I find weird that this is defined as `map(string)` instead of `map(string->string)` or similar but I guess that's how terraform rolls...

Nothing to do, just a comment.

docs/configuration-reference/platforms/aks.md
```diff
@@ -93,24 +93,25 @@ block in the cluster configuration.
...
+| `manage_resource_group` | If `true`, a resource group for the AKS object will be created on behalf of the user.                                                                                                                                                                                                                                                                                                                                                                                                                   |     true      |     bool     |  false   |
+| `worker_pool`           | Configuration block for worker pools. At least one worker pool must be defined.                                                                                                                                                                                                                                                                                                                                                                                                                         |       -       | list(block)  |   true   |
```
iaguis, line 109: This should probably be `list(object)`.

docs/configuration-reference/platforms/packet.md
```diff
@@ -192,59 +192,60 @@ node_type = var.custom_default_worker_type
...
+| `controller_count`                    | Number of controller nodes.                                                                                                                                                                                                                                                       |        1        |    number    |  false   |
+| `controller_type`                     | Packet instance type for controllers.                                                                                                                                                                                                                                             |  "baremetal_0"  |    string    |  false   |
+| `controller_clc_snippets`             | Controller Flatcar Container Linux Config snippets.                                                                                                                                                                                                                               |       []        | map(string)  |  false   |
```
iaguis, line 203: This should be `list(string)`.

knrt10:
@iaguis updated

PR #675, opened 2020-06-25
.codespell.skip: ignore typos in Calico CDRs

invidian:
So when we enable spell checking on CI, builds will still be passing.


invidian:
Upstream PR while we're at it: https://github.com/projectcalico/libcalico-go/pull/1263 :smile:

PR #707, opened 2020-07-07
WIP: Run Cloud Controller Manager on AWS

### LINKED ISSUE #145: Run cloud-controller-manager (CCM) on AWS
Right now we don't run the CCM on AWS so, for example, creating Services of type LoadBalancer doesn't work so we should run it on the AWS platform.

### Cloud providers in Kubernetes

It seems in the beginning the code that communicated with the cloud provider was living in each core Kubernetes component (except the scheduler and kube-proxy), so they all had a `--cloud-provider` flag and they communicated with the cloud.

Today, there are out-of-tree providers too and a new component called **cloud-controller-manager**. In this way, only that component communicates with the cloud and it can be released independently from Kubernetes. This is the recommended way forward. Check https://kubernetes.io/blog/2019/04/17/the-future-of-cloud-providers-in-kubernetes/ for more details.

### Options

There's configuring the cloud-controller-manager component with `--cloud-provider=aws`. You can find an example DaemonSet here: https://kubernetes.io/docs/tasks/administer-cluster/running-cloud-controller/#examples. As mentioned there, the kubelet needs to run with the flag `--cloud-provider=external` and there should be no `--cloud-provider` flags in the API Server nor kube-controller-manager. Also, we might need to play with the CLI flags to get it working correctly.

There's also the [cloud-provider-aws](https://github.com/kubernetes/cloud-provider-aws) repo, which adds to the confusion. It seems to be the home of the out-of-tree AWS cloud provider but development doesn't seem very active (see https://github.com/kubernetes/cloud-provider-aws/issues/42). In the readme it shows the IAM policy needed and the proper node names requirement but it tells to pass `--cloud-provider=external` to the kubelet, API Server and kube-controller manager, contradicting the previous paragraph. Also, there's no example or yaml file on how to deploy this.

### What to do

I think what we should do is use the `cloud-controller-manager` and try to deploy it as mentioned in [Kubernetes Cloud Controller Manager](https://kubernetes.io/docs/tasks/administer-cluster/running-cloud-controller/) for now, and if in the future [cloud-provider-aws](https://github.com/kubernetes/cloud-provider-aws) is more active we can consider switching to it.

BrainBlasted:
Closes #145


invidian (commented):
We also need CCM as static controlplane pod, otherwise node will never become untained and cluster will not converge.

PR #727, opened 2020-07-21
pkg/components/util: use Helm post-renderer instead of Manifests field

invidian:
As Manifests filed in Chart object is a custom field, added to our fork
of Helm. Now Helm offers post-renderer feature, which allows to further
customize the manfiests after Helm's templating engine and before the
installation of the manifests.

This allows us to use post-renderer feature and drop one of maintained
custom patches, which means less maintenance for us.

See https://github.com/helm/helm/issues/8497#issuecomment-662114830 for motivation.


invidian:
Unfortunately, I'm not why this fails :disappointed:

Issue #615, opened 2020-06-10
Refactor AKS quickstart


johananl:
In #273 we've enhanced the quickstart guide for Packet. We should do the same for AKS.

We should use https://github.com/kinvolk/lokomotive/blob/master/docs/quickstarts/packet.md as a template for consistency.

Issue #613, opened 2020-06-10
Refactor AWS quickstart


johananl:
In https://github.com/kinvolk/lokomotive/issues/273 we've enhanced the quickstart guide for Packet. We should do the same for AWS.

We should use https://github.com/kinvolk/lokomotive/blob/master/docs/quickstarts/packet.md as a template for consistency.

johananl:
Our AWS quickstart guide still isn't refactored. I think it's crucial that we have a good UX on AWS. Let's prioritize this.

Issue #368, opened 2020-04-27
Clean up config reference docs


johananl:
The configuration reference docs currently contain lots of extra information and is missing some other information.

IMO, the critical information a config reference should describe is the **configuration syntax** and **a list of all the knobs** with an explanation about each one of them.

Looking at [this](https://github.com/kinvolk/lokomotive/blob/2f92fbe8789dcb49d707a233fe703f017be7721d/docs/configuration-reference/platforms/aws.md) for example, we should probably change the following:

- Remove the "Introduction" and "Prerequisites" sections. These are useful in a tutorial but have no real meaning in a reference guide AFAICT (e.g. prerequisites for what? For reading the document?).
- Add a section explaining the directory structure expected by .lokoctl. For example, if we allow reading multiple .lokocfg files from the CWD, we should explicitly note it.
- Change the "Configuration" section to "Configuration Syntax" and explain the **syntax**. This means explaining the high-levle structure of a .lokocfg file and its various sections.
- Clean up the config sample. There is no value in including a `variable` block for each and every config knob. Rather, we should demonstrate the use of HCL variables in the "Configuration Syntax" section once, and then keep things simple in the rest of the document. Also, we should remove all the extra whitespace in the sample.
- Remove the "Applying" and "Destroying" sections. These don't belong in a config reference.

invidian:
>Break down the attribute reference table. I find the section.attribute syntax confusing. I'd rather have a section for each config area. For example, worker_pool knobs should be described in their own table.

I disagree with that. Helm charts README.md use this "dot" syntax and it works really well, especially for deeply nested structs. IMO way better than other kind of nesting. Example: https://github.com/helm/charts/tree/master/stable/prometheus-operator#prometheus-operator-1

>Cover all of the config knobs. For example, right now there is no information regarding the backend block in the reference.

We do have this: https://github.com/kinvolk/lokomotive/blob/master/docs/configuration-reference/backend/local.md

>- Add a section explaining the directory structure expected by .lokoctl. For example, if we allow reading multiple .lokocfg files from the CWD, we should explicitly note it.
>  - Change the "Configuration" section to "Configuration Syntax" and explain the syntax. This means explaining the high-levle structure of a .lokocfg file and its various sections.
[...]

PR #736, opened 2020-07-27
Packet Quickstart: add machine types to the example config

marga-kinvolk:
Currently it's not possible to deploy the latest lokomotive release on Packet without specifying the machine type. Figuring out the right parameters in the config takes a while. Make this easier by adding them to the example config.

### DIFF
--- a/docs/quickstarts/packet.md
+++ b/docs/quickstarts/packet.md
@@ -116,6 +116,8 @@ cluster "packet" {
   facility = "ams1"
   project_id = "89273817-4f44-4b41-9f0c-cb00bf538542"
 
+  controller_type = "c3.small.x86"
+
   ssh_pubkeys       = ["ssh-rsa AAAA..."]
   management_cidrs  = ["0.0.0.0/0"]
   node_private_cidr = "10.0.0.0/8"
@@ -124,6 +126,7 @@ cluster "packet" {
 
   worker_pool "pool-1" {
     count       = 2
+    node_type = "c3.small.x86"
   }
 }
 


johananl:
Thanks for the PR @marga-kinvolk.

`c3.small.x86` is the default machine type in current `master`:

https://github.com/kinvolk/lokomotive/blob/1afc56c3da7a21c4f0dabbe89bd5e7e04852f8b7/assets/lokomotive-kubernetes/packet/flatcar-linux/kubernetes/variables.tf#L43-L47

On `v0.2.1` the default is `baremetal_0`.

My guess is your Packet project doesn't allow `baremetal_0` machines. The next release should use `c3.small.x86` as the default, and since we're trying to keep the number of knobs the user needs to worry about small in the quickstarts, I'd rather not explicitly specify `c3.small.x86`.

marga-kinvolk:
My project (the Flatcar sponsored project) doesn't allow neither `baremetal_0` nor `c3.small.x86`, so even with the new release it wouldn't have worked. And then there's also the fact that Packet locations don't always have availability of all hardware, leading to failure to provision even when the project allows the machine type.

I understand wanting to keep the documentation simple, but I was following the "quick start" guide and had to instead read a rather long and confusing reference page to understand how to change the machine types. Given my experience with Packet so far (for deploying Flatcar infra, outside of Lokomotive) I think that choosing the machine type is something that belongs in the quick start and not in the "For more things read this long and unfriendly page".

Why I say the "Packet configuration reference" is not particularly friendly:
 * I was looking for the machine type config and I had to process a huge config with tons of variables and parameters and no explanation (the explanation comes later, I'm not sure why the config is at the top, but it's not what I expected from a "reference").
* There's a variable called `workers_type` at the top, but the parameter is actually called `node_type` which needs to be part of the `worker_pool`. The `controller_type` is listed with it's type, but the `node_type` is listed with a variable so it's less evident that this is the machine type.
* Why have all that stuff with the variables at the top? As a newbie to Lokomotive I was pretty puzzled by the variables and kept wondering whether I needed to set a variable or not to change the value I wanted.
* There's an example of how to change the `node_type` but it doesn't actually work because it's not inside the `worker_pool`.  And said example doesn't talk about `controller_type` at all. 
* Yes, by the time I reached the reference, it was pretty clear what parameters I needed to change, but I had to go through all the other things first... I think just the reference instead of the huge config would actually be more helpful.

Apart from this, the current example config in the quick start includes several parameters that aren't explained at all, so I'm not sure adding a couple of parameters that are self-explanatory means the user would have to "worry" about them.

I asked on Slack about this and my understanding was that adding it to the example config would be helpful. If you think it's definitely not ok, would you agree to add a troubleshooting section about not being able to provision the default machine type and how to change it?

rata:
I think it's quite important to add defaults for this in the quickstart.

The current handling in lokomotive, as discussed [in the 0.3.0 release PR][link], *if the controller machine type is changed*, your cluster will be completely broken and *etcd data totally gone*. For workers, you will experience downtime.

Therefore, I would not ever advice anyone to create a cluster such as the quickstart is doing.

We can (and should!) create an issue for this, but as long as the currrent behavior is present, no example should rely on default for controller nodes.


[link]: https://github.com/kinvolk/lokomotive/pull/740#discussion_r462398797

docs/quickstarts/packet.md
```diff
@@ -116,6 +116,8 @@ cluster "packet" {
   facility = "ams1"
   project_id = "89273817-4f44-4b41-9f0c-cb00bf538542"
 
+  controller_type = "c3.small.x86"
```
rata, line 119: You mentioned [here][link] that:
> My project (the Flatcar sponsored project) doesn't allow neither baremetal_0 nor c3.small.x86

Is that a particular setting of the project? I mean, I'd like to use as default something that will work for most projects. I guess this was something on that project and not some generic packet issue?

Just to be sure that this instance type example makes sense as a quickstart default.

[link]: https://github.com/kinvolk/lokomotive/pull/736#issuecomment-665633971

marga-kinvolk: I re-checked this and the issue is that c3.small is only available in Dallas and Sunnyvale (at least for the Flatcar Infra project) and I was deploying the cluster in Amsterdam. So yeah, I think we really need to have this here.

rata: Hm, ok, doesn't seem like the best default but I understand the situation will change and is already the default when no param is specified. So, LGTM :)

rata:
@johananl @surajssd @invidian friendly ping?

johananl:
@marga-kinvolk true, the config reference could definitely use an overhaul. It's tracked here: https://github.com/kinvolk/lokomotive/issues/368

AFAICT the main use for a default value for the machine types is when deploying test clusters, likely using the quickstart. That's arguably the only case where the user might say "just give me cluster, I don't care about the hardware". So, following the above maybe we should just make the machine type knobs required (and add values to the config sample in the quickstart like you've done here).

rata:
@johananl you are completely ignoring my comment in your reasoning. I think what you said is just not true if you take into account my comment.

I also disagree with what you are saying. Someone creating a cluster will pay for the resources, so I don't think it is "I don't care, I trust your defaults, give me some hardware". Furthermore, I would not assume that people deploting with a quickstart will throw away their clusters.

johananl:
> @johananl you are completely ignoring my comment in your reasoning. I think what you said is just not true if you take into account my comment.
> 
> I also disagree with what you are saying. Someone creating a cluster will pay for the resources, so I don't think it is "I don't care, I trust your defaults, give me some hardware". Furthermore, I would not assume that people deploting with a quickstart will throw away their clusters.

@rata looks like a miscommunication here 🙂 I've read your comment and took it into account.

I'm OK with adding the machine types to the docs, and suggested that in addition we consider making the knobs required since I can't think of cases where there user **wouldn't** want to determine the machine type (i.e. the user will likely always want to decide what hardware to use and pay for).

Seems to me we are aligned about adding the values to the quickstart. I'll leave it up to you to decide if you want to also change these knobs to required (that was a just suggestion).

rata:
Sorry, as we talked on slack, I completely misread your last sentence.

Cool, IMHO, we should:
 * Probably merge this
 * Create an issue to handle this on lokomotive (i.e. not let users break their clusters) in _some way_ (documentation or actual code). If we decide to go with docs, we might want to remove the default value as @johananl mentioned.
 * Create an issue to do for quickstarts for other platforms

Am I missing something?

marga-kinvolk:
Hey, can this be merged then? Should I merge it or will one of the Lokomotive folks do it?

johananl:
I'll merge this and open follow up issues.

johananl:
The existing quickstart guides don't contain config snippets (rather, they contain a link to a sample config file).

We already have issues for refactoring the remaining quickstarts following the Packet one:

- https://github.com/kinvolk/lokomotive/issues/613
- https://github.com/kinvolk/lokomotive/issues/614
- https://github.com/kinvolk/lokomotive/issues/615

The description suggests to use the Packet quickstart as a reference so I think we should be good.

johananl:
>Create an issue to handle this on lokomotive (i.e. not let users break their clusters) in some way (documentation or actual code). If we decide to go with docs, we might want to remove the default value as @johananl mentioned.

@rata I'm not sure I understand what you mean here. If you think we need a new issue, could you please open one?

rata:
Sorry it wasn't clear. I created this issue for that: https://github.com/kinvolk/lokomotive/issues/803

Thanks!

PR #742, opened 2020-07-29
Make it obvious that cert-manager is required

marga-kinvolk:
When installing the prometheus-operator component, the instructions don't mention that the cert-manager component is needed if HTTPS is going to work. The example config assumes there is such a cert as it includes a `certmanager_cluster_issuer = "letsencrypt-production"` line, but doesn't point to the separate component.

This commit adds a link to the component as a requirement. I wasn't sure on the exact wording. Better wording welcome.

### DIFF
--- a/docs/configuration-reference/components/prometheus-operator.md
+++ b/docs/configuration-reference/components/prometheus-operator.md
@@ -23,6 +23,10 @@ instances.
   [built-in](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#types-of-persistent-volumes)
   plugins.
 
+* An ingress controller such as [Contour](contour.md) for HTTP ingress.
+
+* The [cert-manager component](cert-manager.md) is required for enabling HTTPS.
+
 ## Configuration
 
 Prometheus Operator component configuration example:


docs/configuration-reference/components/prometheus-operator.md
```diff
@@ -22,6 +22,7 @@ instances.
   [OpenEBS](openebs-operator.md) or one of the
   [built-in](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#types-of-persistent-volumes)
   plugins.
+* The [cert-manager component](cert-manager.md) is required for enabling HTTPS.
```
invidian, line 25: Actually, Ingress controller is also required for both and it is not mentioned :|

marga-kinvolk: Yeah, I actually had to setup contour, metallb, external-dns and cert-manager.  For some reason the other three were clear to me that were needed, the cert-manager one wasn't clear (I thought it would just work).  Should I add links to all 4 of them here?

invidian: Ideally that complexity should be hide from the user, but we are not there yet.

I think having `contour` and `cert-manager` here is OK. `contour` should mention the requirement of `metallb` then and `external-dns` is always optional, as user may create DNS entries on their own, manually too. Perhaps we should document that, if it's not there yet.

marga-kinvolk: Alright, I've added a bullet point for contour, taken from the external-dns doc.

invidian: Thanks! LGTM.

PR #754, opened 2020-08-03
test/components/prometheus-operator: reduce test timeout

invidian:
As global 'go test' timeout is 10 minutes, having 10 minutes here as
well results in triggering the panic caused by global timeout.


invidian:
Forgot to include the link to the failing job, but I don't remember now where I've seen it :smile:

PR #787, opened 2020-08-10
Move walkers to pkg/assets

johananl:
The walkers deal with extracting arbitrary assets from memory to disk, therefore it makes sense to manage them under `pkg/assets`.


pkg/assets/walkers.go
```diff
@@ -12,35 +12,66 @@
...
-	"github.com/kinvolk/lokomotive/pkg/assets"
 )
 
-func CopyingWalker(path string, newDirPerms os.FileMode) assets.WalkFunc {
+// CopyingWalker returns a WalkFunc which writes the given file to disk.
+func CopyingWalker(path string, newDirPerms os.FileMode) WalkFunc {
 	return func(fileName string, fileInfo os.FileInfo, r io.ReadSeeker, err error) error {
 		if err != nil {
-			return errors.Wrapf(err, "error during walking at %q", fileName)
+			return fmt.Errorf("error while walking at %q: %v", fileName, err)
```
johananl, line 29: TODO: Change `%v` to `%w`.

pkg/assets/walkers.go
```diff
@@ -12,35 +12,66 @@
...
 		}
 
 		fileName = filepath.Join(path, fileName)
 
 		if err := os.MkdirAll(filepath.Dir(fileName), newDirPerms); err != nil {
-			return errors.Wrap(err, "failed to create dir")
+			return fmt.Errorf("failed to create dir: %v", err)
```
johananl, line 35: TODO: Change `%v` to `%w`.

pkg/assets/walkers.go
```diff
@@ -12,35 +12,66 @@
...
+	if len(allowedExts) > 0 {
+		extsMap = make(map[string]struct{}, len(allowedExts))
+		for _, ext := range allowedExts {
+			extsMap[ext] = struct{}{}
+		}
+	}
+
+	return func(fileName string, fileInfo os.FileInfo, r io.ReadSeeker, err error) error {
+		if err != nil {
+			return fmt.Errorf("error while walking at %q: %v", fileName, err)
```
johananl, line 55: TODO: Change `%v` to `%w` and rephrase the error.

pkg/assets/walkers.go
```diff
@@ -12,35 +12,66 @@
...
+
+		if extsMap != nil {
+			if _, ok := extsMap[filepath.Ext(fileName)]; !ok {
+				return nil
+			}
+		}
+
+		contents, err := ioutil.ReadAll(r)
+		if err != nil {
+			return fmt.Errorf("failed to read %q: %v", fileName, err)
```
johananl, line 66: TODO: Change `%v` to `%w` and rephrase the error.

pkg/assets/walkers.go
```diff
@@ -12,35 +12,66 @@
...
 	"path/filepath"
-
-	"github.com/pkg/errors"
-
-	"github.com/kinvolk/lokomotive/pkg/assets"
 )
 
-func CopyingWalker(path string, newDirPerms os.FileMode) assets.WalkFunc {
+// CopyingWalker returns a WalkFunc which writes the given file to disk.
+func CopyingWalker(path string, newDirPerms os.FileMode) WalkFunc {
```
invidian, line 26: I'm not sure if that's the right time to touch it, but it seems to me, that term "Walker" is an implementation detail, which shouldn't be part of the assets public API.

I'd like to have something like `assets.Read` or `assets.Unpack`, this would make other code more readable.

johananl: I'm fine with looking into this (but in a new issue please).

johananl:
Going to merge this once CI is green if there is no further feedback.

pkg/assets/walkers.go
invidian: Created https://github.com/kinvolk/lokomotive/issues/796 :+1:

Issue #716, opened 2020-07-13
Refactor platform handling


johananl:
## General

At the moment the `platform` package has multiple responsibilities:

- It defines which Terraform modules are used to deploy a cluster on a specific platform.
- It handles the Terraform execution logic for each platform. This logic is largely generic and therefore duplicated among the platform.
- If https://github.com/kinvolk/lokomotive/pull/655 is merged, the package will also define which Helm charts are used by a platform.
- If https://github.com/kinvolk/lokomotive/pull/655 is merged, the package will also handle the extraction of Helm-related assets to disk.

We should consider refactoring the `platform` package to have a single purpose: **defining the contents of a Lokomotive platform**. The rationale for doing the refactor:

- Less code duplication.
- Better delineation between the packages. This makes it easier to figure out where a piece of code should live and also makes testing easier.

Following the refactoring, the following should be true:

- The `platform` package exposes a minimal API which allows specifying the "recipe" for deploying a Lokomotive cluster on a given cloud (or bare metal) platform. The package doesn't perform any API calls, file operations or other forms of I/O.
- Specifically, the `platform` package doesn't depend on the `terraform` package. Any code related to executing Terraform operations should be generic (i.e. not platform specific) and reside in the `terraform` package.
- Specifically, the `platform` package doesn't depend on the `assets` package. Any code related to generating assets into the binary and extracting assets to disk should be generic and reside in the `assets` package.

## Implementation

When implementing this change, we should look for cases where we have "specialized" logic for a platform and ideally eliminate such cases. This is important if we want to make Terraform-related code generic. An example of such specialized logic:

https://github.com/kinvolk/lokomotive/blob/fee1edde76bf3d1a667f1fc0443802863569958e/pkg/platform/packet/packet.go#L142-L160

https://github.com/kinvolk/lokomotive/blob/fee1edde76bf3d1a667f1fc0443802863569958e/pkg/platform/aws/aws.go#L143-L152

As can be seen above, the Packet implementation of `Initialize()` contains things like API token validation and DNS config validation whereas the AWS implementation of the same function does not.

If we realize we can't make Terraform-related code completely generic, we can look into allowing arbitrary logic to be specified in a platform implementation, for example using callback functions. We should do our best to keep things well defined if we go in this direction. For example, we may want to consider defining a small set of "hooks" to allow a platform implementation to specific, for example, that some arbitrary logic needs to be run before/after executing `terraform apply`, before/after applying Lokomotive components etc.

[...]

PR #794, opened 2020-08-11
Refactor Terraform executor

johananl:
This PR paves the way towards https://github.com/kinvolk/lokomotive/issues/716. By making `terraform.Executor` accept an arbitrary series of "steps" we can now declaratively define the infrastructure pieces a platform is composed of.

Example:

```go
steps := []terraform.ExecutionStep{
    {
        Description: "create controllers",
        Args: []string{
            "apply",
            "-auto-approve",
            fmt.Sprintf("-target=module.packet-%s.packet_device.controllers", c.ClusterName),
        },
    },
    {
        Description: "create DNS records",
        Args:        []string{"apply", "-auto-approve", "-target=module.dns"},
    },
    {
        Description: "refresh Terraform state",
        Args:        []string{"refresh"},
    },
    {
        Description:      "complete infrastructure creation",
        Args:             []string{"apply", "-auto-approve"},
        // Verify DNS entries were manually created by the user before proceeding.
        PreExecutionHook: dns.ManualConfigPrompt(&c.DNS),
    },
}
```

I've deliberately avoided refactoring all the various platforms to keep the scope of this PR narrow. I've only refactored `platform/packet` (which is our most complex platform implementation at the moment) to demonstrate how the new structure can be used. The rest of the platforms will be refactored in subsequent PRs.

### DIFF
--- a/pkg/dns/dns.go
+++ b/pkg/dns/dns.go
@@ -61,37 +61,39 @@ func (c *Config) Validate() error {
 	return fmt.Errorf("invalid DNS provider %q", c.Provider)
 }
 
-// AskToConfigure reads the required DNS entries from a Terraform output,
-// asks the user to configure them and checks if the configuration is correct.
-func (c *Config) AskToConfigure(ex *terraform.Executor) error {
-	dnsEntries, err := readDNSEntries(ex)
-	if err != nil {
-		return err
-	}
-
-	fmt.Printf("Please configure the following DNS entries at the DNS provider which hosts %q:\n", c.Zone)
-	prettyPrintDNSEntries(dnsEntries)
-
-	for {
-		fmt.Printf("Press Enter to check the entries or type \"skip\" to continue the installation: ")
-
-		var input string
-		fmt.Scanln(&input)
-
-		if input == "skip" {
-			break
-		} else if input != "" {
-			continue
+// ManualConfigPrompt returns a terraform.ExecutionHook which prompts the user to configure DNS
+// entries manually and verifies the entries were created successfully.
+func (c *Config) ManualConfigPrompt() terraform.ExecutionHook {
+	return func(ex *terraform.Executor) error {
+		dnsEntries, err := readDNSEntries(ex)
+		if err != nil {
+			return err
 		}
 
-		if checkDNSEntries(dnsEntries) {
-			break
+		fmt.Printf("Please configure the following DNS entries at the DNS provider which hosts %q:\n", c.Zone)
+		prettyPrintDNSEntries(dnsEntries)
+
+		for {
+			fmt.Printf("Press Enter to check the entries or type \"skip\" to continue the installation: ")
+
+			var input string
+			fmt.Scanln(&input)
+
+			if input == "skip" {
+				break
+			} else if input != "" {
+				continue
+			}
+
+			if checkDNSEntries(dnsEntries) {
+				break
+			}
+
+			fmt.Println("Entries are not correctly configured, please verify.")
 		}
 
-		fmt.Println("Entries are not correctly configured, please verify.")
+		return nil
 	}
-
-	return nil
 }
 
 func readDNSEntries(ex *terraform.Executor) ([]dnsEntry, error) {
--- a/pkg/dns/dns.go
+++ b/pkg/dns/dns.go
@@ -107,7 +107,7 @@ func (c *Config) ManualConfigPrompt() terraform.ExecutionHook {
 }
 
 func readDNSEntries(ex *terraform.Executor) ([]dnsEntry, error) {
-	output, err := ex.ExecuteSync("output", "-json", "dns_entries")
+	output, err := ex.OutputBytes("dns_entries")
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to get DNS entries")
 	}
--- a/pkg/dns/dns.go
+++ b/pkg/dns/dns.go
@@ -15,11 +15,14 @@
 package dns
 
 import (
+	"bufio"
 	"encoding/json"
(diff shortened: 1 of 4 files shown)


pkg/dns/dns.go
```diff
@@ -61,41 +61,47 @@ func (c *Config) Validate() error {
...
-		var input string
-		fmt.Scanln(&input)
+			var input string
 
-		if input == "skip" {
-			break
-		} else if input != "" {
-			continue
-		}
+			_, err := fmt.Scanln(&input)
```
johananl, line 81: This breaks manual DNS. Working on a fix.

johananl: Done.

pkg/terraform/executor.go
```diff
@@ -176,14 +207,27 @@ func tailFile(path string, done chan struct{}, wg *sync.WaitGroup) {
...
+// Execute accepts one or more ExecutionSteps and executes them sequentially in the order they were
+// provided. If a step has a PreExecutionHook defined, the hook is run prior to executing the step.
+// If any error is encountered, the error is returned and the execution is halted.
+func (ex *Executor) Execute(steps ...ExecutionStep) error {
+	for _, s := range steps {
+		if s.PreExecutionHook != nil {
+			ex.logger.Printf("Running pre-execution hook for step %q", s.Description)
+
+			if err := s.PreExecutionHook(ex); err != nil {
+				return fmt.Errorf("pre-execution hook failed: %w", err)
```
invidian, line 219:
```suggestion
				return fmt.Errorf("running pre-execution hook: %w", err)
```

pkg/dns/dns.go
```diff
@@ -61,37 +64,46 @@ func (c *Config) Validate() error {
...
-		var input string
-		fmt.Scanln(&input)
+			var input string
 
-		if input == "skip" {
-			break
-		} else if input != "" {
-			continue
-		}
+			reader := bufio.NewReader(os.Stdin)
```
invidian, line 84: It seems switch to NewReader should be done in separate commit?

johananl: I did this to satisfy the linter. We didn't actually touch this code in this PR, we just moved it around a bit but it was enough for the linter to complain. The complaint was valid though - we had an unhandled error from `fmt.Scanln()`.

Does it still make sense to you to split this?

johananl: Moved this to a separate commit to save time. I can re-squash if necessary.

invidian: Ah, I didn't know the change was driven by the linter. I think it make sense to do that in separate commit then, thanks!

pkg/dns/dns.go
```diff
@@ -15,11 +15,14 @@
 package dns
```
invidian, line 15: By reading the commit message, I'm under the impression, it would be good to test such use cases (user blindly hitting Enter key etc.). But it's out of scope I guess.

pkg/dns/dns.go
```diff
@@ -61,41 +64,50 @@ func (c *Config) Validate() error {
...
-// AskToConfigure reads the required DNS entries from a Terraform output,
-// asks the user to configure them and checks if the configuration is correct.
-func (c *Config) AskToConfigure(ex *terraform.Executor) error {
-	dnsEntries, err := readDNSEntries(ex)
-	if err != nil {
-		return err
-	}
+// ManualConfigPrompt returns a terraform.ExecutionHook which prompts the user to configure DNS
+// entries manually and verifies the entries were created successfully.
+func ManualConfigPrompt(c *Config) terraform.ExecutionHook {
```
invidian, line 69: Why change the receiver to the argument?

pkg/dns/dns.go
```diff
@@ -61,41 +64,50 @@ func (c *Config) Validate() error {
...
+	return func(ex *terraform.Executor) error {
+		dnsEntries, err := readDNSEntries(ex)
+		if err != nil {
+			return err
+		}
 
-	fmt.Printf("Please configure the following DNS entries at the DNS provider which hosts %q:\n", c.Zone)
-	prettyPrintDNSEntries(dnsEntries)
+		fmt.Printf("Please configure the following DNS entries at the DNS provider which hosts %q:\n", c.Zone)
```
invidian, line 76: Perhaps we could have another unexported function to ask questions, which would take dnsEntries as an argument? That would simplify the testing.

pkg/dns/dns.go
johananl: IMO - yes, and yes.

pkg/dns/dns.go
johananl: What do you mean by a function for asking questions? Wrap `fmt.Printf()` in a function?
Anyway, IMO this belongs in a "refactor `pkg/dns`" task, i.e. out of scope.

pkg/dns/dns.go
johananl: No good reason. Fixed.

pkg/dns/dns.go
invidian: I meant, instead of taking executor as an argument and then execute the prints etc. it would be better to have a function like this, which takes an executor, gets dnsEntries using helper and passes that to function which does prints and reads user input. This way, we could fully test prints and reading user input.

johananl: OK, feel free to open an issue if you think we should do that. I'm resolving this since I think we agree this is out of scope here.

invidian: Not sure if such small things are worth creating an issue and allocating dedicated time. Seems more like a thing to do together with something IMO.

PR #805, opened 2020-08-14
Update bootkube and HAProxy

invidian:


invidian:
It seems this requires some controversial changes to be introduced, so perhaps it would be better to wait for K8s 1.19 before adding wait to bootkube.

invidian:
Hm, this will actually fix some transient errors we hit during bootstrapping, as this is the moment when kubelet didn't try restarting kube-apiserver pod yet and bootstrap kube-apiserver is disabled already.

invidian:
As K8s 1.19.0 is out, we should wait with this PR until we upgrade and then we do it nicely.

invidian:
Will close.

PR #359, opened 2020-04-27
(WIP) Refactoring loading of configuration, cluster operations management.

ipochi:
After receiving feedback from the design document on the PR #296 

I have remove all user facing changes , hence this PR tries to address the following things:

1. Remove all cluster operations related logic from `cli/cmd` package.
2. A central place for loading,parsing and validating HCL configuration (w.i.p for loading components).
3. Units tests.

I am currently exploring design options for removing the duplicity of fields in different platforms. Your opinions and feedback are welcome.

`pkg/lokomotive` is the the place where most of the action is happening around operations on Lokomotive cluster.

P.S - @rata I tried to rebase your reservations-id PR on Packet, please let me know if I did a correct job on it or not. I will make the requisite changes if not done correctly.

Issue #312, opened 2020-04-22
Decouple Lokomotive versioning from lokoctl


iaguis:
Currently the Lokomotive version is embedded in the `lokoctl` binary. That is, to update Lokomotive you need to get the new `lokoctl` binary and run `lokoctl cluster apply`.

We should consider decoupling the Lokomotive version from `lokoctl` and have some kind of release manifest.

invidian:
I have some ideas how to address that.

invidian:
Something interesting about Go plugins I found:
```
FATA[0000] Getting release: opening plugin ./release.so: plugin.Open("./release"): plugin was built with a different version of package github.com/kinvolk/lokomotive/pkg/release  command="lokoctl component list"
```

invidian:
So, the main thing I struggle right now with Go plugins implementation is fact, that `github.com/kinvolk/lokomotive/pkg/assets` package import must be only imported in the release plugin code and not in `cli/cmd` package. This is required, as otherwise embedded assets files are still taken from the `lokoctl` binary rather than from `release.so` file.

This turns out to be surprisingly difficult because of the way our code is structured.

The fact that CLI code currently implements most of the cluster operations logic rather than just being a gate to trigger it is a problem. We also have many tiny imports, which additionally clutters the dependencies tree and we still use #630 in too many places, which makes it more difficult to move code around.

PR #806, opened 2020-08-14
Refactor pkg/config

johananl:
This PR is rather cosmetic and focuses on better naming and readability.

The whitespace changes are for satisfying the linter.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -61,91 +61,59 @@ type Config struct {
 	EvalContext *hcl.EvalContext
 }
 
-func loadLokocfgPaths(configPath string) ([]string, error) {
-	isDir, err := util.PathIsDir(configPath)
-	if err != nil {
-		return nil, fmt.Errorf("failed to stat config path %q: %w", configPath, err)
-	}
-	var lokocfgPaths []string
-	if isDir {
-		globPattern := filepath.Join(configPath, "*.lokocfg")
-		configFiles, err := filepath.Glob(globPattern)
-		if err != nil {
-			return nil, fmt.Errorf("bad filepath glob pattern %q: %w", globPattern, err)
+// Component finds a component named name and returns its configuration as a pointer to an
+// hcl.Body. If the component isn't found, nil is returned.
+func (c *Config) Component(name string) *hcl.Body {
+	for _, component := range c.RootConfig.Components {
+		if name == component.Name {
+			return &component.Config
 		}
-		lokocfgPaths = append(lokocfgPaths, configFiles...)
-	} else {
-		lokocfgPaths = append(lokocfgPaths, configPath)
 	}
-	return lokocfgPaths, nil
+
+	return nil
 }
 
-func LoadConfig(lokocfgPath, lokocfgVarsPath string) (*Config, hcl.Diagnostics) {
-	lokocfgPaths, err := loadLokocfgPaths(lokocfgPath)
+// ReadHCL reads all HCL files at path and one HCL values file at valuesPath, constructs a Config
+// and returns a pointer to it.
+func ReadHCL(path, valuesPath string) (*Config, hcl.Diagnostics) {
+	names, err := lokocfgFilenames(path)
 	if err != nil {
-		return nil, hcl.Diagnostics{
-			&hcl.Diagnostic{
-				Severity: hcl.DiagError,
-				Summary:  err.Error(),
-			},
-		}
+		return nil, diag(fmt.Errorf("getting lokocfg paths: %w", err))
 	}
 
-	hclParser := hclparse.NewParser()
-
-	var hclFiles []*hcl.File
-	for _, f := range lokocfgPaths {
-		hclFile, diags := hclParser.ParseHCLFile(f)
-		if len(diags) > 0 {
-			return nil, diags
-		}
-		hclFiles = append(hclFiles, hclFile)
-	}
-
-	configBody := hcl.MergeFiles(hclFiles)
-
-	exists, err := util.PathExists(lokocfgVarsPath)
+	files, err := readFiles(names)
 	if err != nil {
-		return nil, hcl.Diagnostics{
-			&hcl.Diagnostic{
-				Severity: hcl.DiagError,
-				Summary:  fmt.Sprintf("could not stat %q: %v", lokocfgVarsPath, err),
-			},
-		}
+		return nil, diag(err)
 	}
+
+	body, diags := parseHCLFiles(files)
+	if diags.HasErrors() {
+		return nil, diags
+	}
+
+	varsExist, err := util.PathExists(valuesPath)
+	if err != nil {
+		return nil, diag(fmt.Errorf("could not stat %q: %w", valuesPath, err))
+	}
+
 	var userVals map[string]cty.Value
-	var diags hcl.Diagnostics
-	if exists {
-		userVals, diags = LoadValuesFile(lokocfgVarsPath)
-		if len(diags) > 0 {
+	if varsExist {
+		userVals, diags = readValuesFile(valuesPath)
+		if diags.HasErrors() {
 			return nil, diags
 		}
 	}
 
 	var rootConfig RootConfig
-	diags = gohcl.DecodeBody(configBody, nil, &rootConfig)
-	if len(diags) > 0 {
+
+	diags = gohcl.DecodeBody(body, nil, &rootConfig)
+	if diags.HasErrors() {
 		return nil, diags
 	}
 
-	variables := map[string]cty.Value{}
-	for _, v := range rootConfig.Variables {
-		if userVal, ok := userVals[v.Name]; ok {
-			variables[v.Name] = userVal
-			continue
-		}
-		if len(v.Default) == 0 {
-			continue
-		}
-		defaultValue, hasDefaultValue := v.Default["default"]
-		if !hasDefaultValue {
-			continue
-		}
-		defaultVal, diags := defaultValue.Expr.Value(nil)
-		if len(diags) > 0 {
-			return nil, diags
-		}
-		variables[v.Name] = defaultVal
+	variables, diags := populateVars(rootConfig.Variables, userVals)
+	if diags.HasErrors() {
+		return nil, diags
 	}
 
 	evalContext := hcl.EvalContext{
@@ -162,6 +130,44 @@ func LoadConfig(lokocfgPath, lokocfgVarsPath string) (*Config, hcl.Diagnostics) 
 		RootConfig:  &rootConfig,
 		EvalContext: &evalContext,
 	}, nil
+}
+
+// readFiles reads all files based on the provided names and returns a map where the key is the
+// name of the file and the value is a slice of bytes holding the file's contents.
+func readFiles(names []string) (map[string][]byte, error) {
+	files := map[string][]byte{}
+
+	for _, fn := range names {
+		data, err := ioutil.ReadFile(fn) //nolint:gosec
+		if err != nil {
+			return nil, fmt.Errorf("reading file %q: %w", fn, err)
+		}
+
+		files[fn] = data
+	}
+
+	return files, nil
+}
+
+// parseHCLFiles reads the provided HCL files, merges them into a single hcl.Body and returns it.
+func parseHCLFiles(files map[string][]byte) (hcl.Body, hcl.Diagnostics) {
+	parser := hclparse.NewParser()
+
+	hclFiles := []*hcl.File{}
+
+	var diags hcl.Diagnostics
+
+	for name, data := range files {
+		hclFile, d := parser.ParseHCL(data, name)
+		diags = append(diags, d...)
+		hclFiles = append(hclFiles, hclFile)
+	}
+
+	if diags.HasErrors() {
+		return nil, diags
+	}
+
+	return hcl.MergeFiles(hclFiles), nil
 }
 
 func evalFuncPathExpand() function.Function {
@@ -195,13 +201,68 @@ func evalFuncFile() function.Function {
 	})
 }
 
-// LoadComponentConfigBody returns nil if no component with the given
-// name is found in the configuration
-func (c *Config) LoadComponentConfigBody(componentName string) *hcl.Body {
-	for _, component := range c.RootConfig.Components {
-		if componentName == component.Name {
-			return &component.Config
-		}
+// lokocfgFilenames returns a slice containing the names of all .lokocfg files found under path p.
+// If p is not a directory, p is returned by itself in the slice.
+func lokocfgFilenames(p string) ([]string, error) {
+	isDir, err := util.PathIsDir(p)
+	if err != nil {
+		return nil, fmt.Errorf("checking if %q is a directory: %w", p, err)
 	}
-	return nil
+
+	if !isDir {
+		return []string{p}, nil
+	}
+
+	globPattern := filepath.Join(p, "*.lokocfg")
+
+	names, err := filepath.Glob(globPattern)
+	if err != nil {
+		return nil, fmt.Errorf("bad filepath glob pattern %q: %w", globPattern, err)
+	}
+
+	return names, nil
+}
+
+// populateVars accepts a slice of variables and a map of user-provided values. The function
+// populates each variable with the corresponding user-provided value or with a default. If a
+// variable doesn't have a matching value and there is no default, an error is returned (as
+// hcl.Diagnostics).
+func populateVars(vars []variable, values map[string]cty.Value) (map[string]cty.Value, hcl.Diagnostics) {
+	res := map[string]cty.Value{}
+
+	for _, v := range vars {
+		if value, ok := values[v.Name]; ok {
+			res[v.Name] = value
+
+			continue
+		}
+
+		if len(v.Default) == 0 {
+			continue
+		}
+
+		defaultValue, hasDefaultValue := v.Default["default"]
+		if !hasDefaultValue {
+			continue
+		}
+
+		dv, diags := defaultValue.Expr.Value(nil)
+		if diags.HasErrors() {
+			return nil, diags
+		}
+
+		res[v.Name] = dv
+	}
+
+	return res, nil
+}
+
+// diag returns an hcl.Diagnostics with a single hcl.Diagnostic based on the provided error.
+func diag(err error) hcl.Diagnostics {
+	return hcl.Diagnostics{
+		&hcl.Diagnostic{
+			Severity: hcl.DiagError,
+			Summary:  err.Error(),
+		},
+	}
 }
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -35,8 +35,8 @@ type variable struct {
 }
 
 type cluster struct {
-	Name   string   `hcl:"name,label"`
-	Config hcl.Body `hcl:",remain"`
+	Platform string   `hcl:"platform,label"`
+	Config   hcl.Body `hcl:",remain"`
 }
 
 type component struct {
--- a/pkg/config/config_internal_test.go
+++ b/pkg/config/config_internal_test.go
@@ -1,0 +1,242 @@
+// Copyright 2020 The Lokomotive Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package config
+
+import (
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"reflect"
+	"testing"
+)
+
+//nolint:funlen
+func TestReadFiles(t *testing.T) {
+	tests := []struct {
+		desc  string
+		files map[string][]byte
+	}{
+		{
+			desc: "Single file",
+			files: map[string][]byte{
+				"foo.txt": []byte(`foo`),
+			},
+		},
+		{
+			desc: "Multiple files",
+			files: map[string][]byte{
+				"foo.txt": []byte(`foo`),
+				"bar.txt": []byte(`bar`),
+				"baz.txt": []byte(`baz`),
+			},
+		},
+		{
+			desc: "Empty file",
+			files: map[string][]byte{
+				"foo.txt": []byte(``),
+			},
+		},
+		{
+			desc:  "No files",
+			files: map[string][]byte{},
+		},
+	}
+
+	d, err := ioutil.TempDir("", "lokoctl-tests-")
+	if err != nil {
+		t.Fatalf("Could not create temporary directory: %v", err)
+	}
+
+	t.Cleanup(func() {
+		if err := os.RemoveAll(d); err != nil {
+			t.Logf("Could not remove temporary directory %q: %v", d, err)
+		}
+	})
+
+	for _, test := range tests {
+		// test.files maps a short filename to its contents. We need full paths to verify the
+		// result so we create a copy of the original map with full paths as the keys.
+		wantFiles := map[string][]byte{}
+
+		for fn, data := range test.files {
+			p := filepath.Join(d, fn)
+			wantFiles[p] = data
+
+			if err := ioutil.WriteFile(p, data, 0600); err != nil {
+				t.Fatalf("Could not write test file %q: %v", p, err)
+			}
+		}
+
+		in := []string{}
+		for fn := range wantFiles {
+			in = append(in, fn)
+		}
+
+		out, err := readFiles(in)
+		if err != nil {
+			t.Fatalf("Unexpected error: %v", err)
+		}
+
+		if !reflect.DeepEqual(out, wantFiles) {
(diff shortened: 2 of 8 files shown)


pkg/config/config.go
```diff
@@ -195,13 +192,28 @@ func evalFuncFile() function.Function {
...
 		}
+
+		paths = append(paths, configFiles...)
+	} else {
+		paths = append(paths, p)
 	}
-	return nil
+
+	return paths, nil
 }
```
invidian, line 219: This can be simplified with
```go
if !isDir {
  return []string{p}
}

glob etc...
```

pkg/config/config.go
```diff
@@ -61,27 +61,23 @@ type Config struct {
...
+
+	return nil
 }
 
-func LoadConfig(lokocfgPath, lokocfgVarsPath string) (*Config, hcl.Diagnostics) {
-	lokocfgPaths, err := loadLokocfgPaths(lokocfgPath)
+// New reads HCL files at path and HCL variable files at varPath, constructs a Config and returns a
+// pointer to it.
+//nolint:funlen
+func New(path, varPath string) (*Config, hcl.Diagnostics) {
```
invidian, line 79: Again, `LoadConfig` seems to me like the right name for this function. `New` is usually used as a constructor to get new instance of the object, however this function executes specific action, which is loading the configuration using given paths. We could do things llike `config.Load()` or `config.LoadHCL()`, but just `New` seems more ambiguous to me.

invidian (changes requested):
BTW @ipochi did awesome job refactoring `pkg/config`, which never got trough: https://github.com/kinvolk/lokomotive/pull/359/files#diff-091ace02b49a46f55f2586cfec89a160

pkg/config/config.go
```diff
@@ -195,13 +192,28 @@ func evalFuncFile() function.Function {
 	})
 }
 
-// LoadComponentConfigBody returns nil if no component with the given
```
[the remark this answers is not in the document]
johananl: The problems I have with `LoadComponentConfigBody`:

- "Load" doesn't mean anything to me. I can interpret it as "read from disk", "unmarshal" and probably a bunch of other options.
- It's long.
- "Config" is redundant since we are in the `config` package ("don't stutter"...).

>By the function Component I would expect to get an instance of Component interface, not HCL body.

I don't think there is anything which hints that you would get a `components.Component` here. We are in a package named `config`.
By the way, I've been contemplating whether `hcl` would be a better name for the package. On one hand it makes it easier to distinguish between "cluster config" (i.e. HCL files) and platform config (i.e. a struct containing platform-specific knobs).

How would this look to you if it were `hcl.Component()` and maybe also `hcl.Cluster()`?

pkg/config/config.go
```diff
@@ -61,27 +61,23 @@ type Config struct {
...
-	return lokocfgPaths, nil
+
+	return nil
 }
 
-func LoadConfig(lokocfgPath, lokocfgVarsPath string) (*Config, hcl.Diagnostics) {
-	lokocfgPaths, err := loadLokocfgPaths(lokocfgPath)
+// New reads HCL files at path and HCL variable files at varPath, constructs a Config and returns a
+// pointer to it.
+//nolint:funlen
```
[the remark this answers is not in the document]
johananl: Was trying to keep the changeset small.
I've broken out some of the logic in `New()` into smaller functions.

pkg/config/config.go
johananl: Yup, and we also don't need `append()`. Fixed.

pkg/config/config.go
johananl: I don't think `LoadConfig` is a good name, because:

- "Load" is a very ambiguous term IMO.
- "Config" is stuttering (`config.LoadConfig()` could be e.g. `config.Load()`).

How about `config.ReadHCL()`? IMO "read" makes it clearer we are reading files from disk.

johananl: I've changed to `config.ReadHCL()` to save time. I can change again if you still have concerns.

pkg/config/config.go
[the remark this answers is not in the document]
invidian: Thanks

pkg/config/config.go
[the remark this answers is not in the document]
invidian: Regarding the config/HCL naming for the package, I think ideally we should have a `config` package, which allows expressing Lokomotive configuration using Go code and this one could have a `hcl` package, which would handle reading HCL files and converting them to Go structs. This way, if we want, we can later on also add different formats of the configuration.

>"Load" doesn't mean anything to me. I can interpret it as "read from disk", "unmarshal" and probably a bunch of other options.

Hm, in this particular case, we don't even load or read anything, so yeah, the name could be improved :)

I have a feeling, that getting rid of `LoadConfig` method from `Component` interface would save us time, as we would remove this method completely and we just return the `Component` to the user (or `map[string]Component`).

invidian (changes requested):
Also, when we do refactoring, we should also add tests to the code to ensure, that newly introduce code is easily testable.

johananl:
Added some unit tests. They are non-exhaustive but they do demonstrate the testability of the new functions and provides a base for easily adding new tests.

johananl:
- [ ] TODO: Test `populateVars()`

pkg/config/config_test.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+
+	t.Cleanup(func() {
+		if err := os.RemoveAll(d); err != nil {
+			t.Logf("Could not remove temporary directory %q: %v", d, err)
+		}
+	})
+
+	for _, test := range tests {
+		// Pin test to avoid scope bugs: https://github.com/kyoh86/scopelint
```
invidian, line 105: This is a standard thing to do, I think we can remove this comment.

pkg/config/config_test.go
```diff
@@ -0,0 +1,215 @@
...
+    count         = 3
+    instance_type = "fake"
+    ssh_pubkeys   = ["stuff"]
+  }
+}`),
+		},
+		{
+			desc: "Empty config",
+			conf: []byte(``),
+			// TODO: Is it OK that an empty config doesn't generate errors?
```
invidian, line 73: Syntactically it is OK, while parsing the HCL. On higher level we should fail anyway, if no components or no cluster is configured.

invidian (changes requested):
Hm, I still struggle to see the reasoning behind changing the private method names in `pkg/config`. Perhaps putting a move of the functions inside the file to separate commit would be better, as it would highlight which changes has been performed exactly, maybe I comment code, which didn't change. I'd also split the changes into smaller commits, so then the extra reasoning can be added in the commit messages.

Changes to the exported API seem to indeed make sense.

pkg/config/config.go
```diff
@@ -195,13 +185,58 @@ func evalFuncFile() function.Function {
...
 
-// LoadComponentConfigBody returns nil if no component with the given
-// name is found in the configuration
-func (c *Config) LoadComponentConfigBody(componentName string) *hcl.Body {
-	for _, component := range c.RootConfig.Components {
-		if componentName == component.Name {
-			return &component.Config
+// lokocfgFiles returns a slice containing the paths of all .lokocfg files found under path p. If p
+// is not a directory, p is returned by itself in the slice.
+func lokocfgFiles(p string) ([]string, error) {
```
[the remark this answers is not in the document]
johananl: As long as we only use this logic for `.lokocfg` files what you're suggesting is IMO unnecessary indirection. Let's keep things simple until we actually need to complicate them.

pkg/config/config.go
```diff
@@ -164,6 +126,34 @@ func LoadConfig(lokocfgPath, lokocfgVarsPath string) (*Config, hcl.Diagnostics)
...
+		}
+	}
+
+	hclParser := hclparse.NewParser()
+
+	hclFiles := []*hcl.File{}
+
+	for _, f := range lokocfgFiles {
+		hclFile, diags := hclParser.ParseHCLFile(f)
+		if len(diags) > 0 {
```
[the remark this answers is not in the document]
johananl: Why?

pkg/config/config.go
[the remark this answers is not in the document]
invidian: It won't really be the redirection, just passing the extension via the argument. IMO that would allow giving the function a name, which would better describe what we do. `lokocfgFiles` is IMO ambiguous. We could name it `glob` or `filesWithExtension()`, which reads nicely: "give me list of files with extension X at given path".

pkg/config/config.go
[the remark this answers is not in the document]
invidian: The diagnostics may also include messages with different severity, which we currently don't care about. It's also reads more nicely IMO.

pkg/config/config.go
```diff
@@ -164,6 +126,34 @@ func LoadConfig(lokocfgPath, lokocfgVarsPath string) (*Config, hcl.Diagnostics)
...
+			return nil, diags
```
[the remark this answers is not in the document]
johananl: Done.

pkg/config/config.go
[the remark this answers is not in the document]
johananl: >It won't really be the redirection, just passing the extension via the argument.

That's indirection as far as I can tell :slightly_smiling_face: Instead of using a hardcoded `.lokocfg` you create the string somewhere else and pass it to the function. You now have a symbol which refers to the string rather than simply a string. Hence indirection.

I have no problem with your `filesWithExtension()` idea, I just don't see why we need it here. Making things more generic than they need to be makes it harder to understand the code. The more you need to hop around to figure out what some value is, the harder is the code to understand.

BTW, I've renamed the function to `lokocfgPaths()` again as I realized it's more accurate.

pkg/config/config.go
```diff
@@ -195,13 +185,58 @@ func evalFuncFile() function.Function {
...
+	}
+
+	return files, nil
+}
+
+// populateVars accepts a slice of variables and a map of user-provided values. The function
+// populates each variable with the corresponding user-provided value or with a default. If a
+// variable doesn't have a matching value and there is no default, an error is returned (as
+// hcl.Diagnostics).
+func populateVars(vars []variable, vals map[string]cty.Value) (map[string]cty.Value, hcl.Diagnostics) {
```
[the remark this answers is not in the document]
johananl: Done.

pkg/config/config.go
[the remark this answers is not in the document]
invidian: >That's indirection as far as I can tell slightly_smiling_face Instead of using a hardcoded .lokocfg you create the string somewhere else and pass it to the function. You now have a symbol which refers to the string rather than simply a string. Hence indirection.

Well, if you put that in the const (and we should), you would get `filesWithExtension(configPath, lokocfgExtension)`. This way, you don't need to know what the extension actually is to understand the code.

Anyway, `lokocfgPaths()` seems OK to me too, thanks :+1:

pkg/config/config.go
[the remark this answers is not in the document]
johananl: I've never seen this happen, but OK.

pkg/config/config_internal_test.go
```diff
@@ -0,0 +1,148 @@
...
+  ssh_pubkeys      = ["stuff"]  
+  worker_pool "my-wp-name" {
+    count         = 3
+    instance_type = "fake"
+    ssh_pubkeys   = ["stuff"]
+  }
+}`),
+				},
+				{
+					filename: "foo.txt",
```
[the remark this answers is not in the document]
johananl: We already have `desc: "Unrelated files in config dir"`. I can make it "Unrelated files in config dir should be ignored" but I find it kind of obvious since the files are unrelated...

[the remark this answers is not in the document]
johananl: In this specific test function, there are only two possible outcomes for a test case: "errors" and "no errors". I thought it's clear what when we don't have e.g. `wantDiags` there shouldn't be any errors.

pkg/config/config.go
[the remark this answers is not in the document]
johananl: I understand your point but I don't think this is the right time to make the function generic. The right time would IMO be the moment we have at least 2 extensions. The current state is simpler than the state after your proposed refactoring and I think we should aim at keeping things simple as long as we can.

pkg/config/config.go
```diff
@@ -61,91 +61,53 @@ type Config struct {
...
 			},
 		}
 	}
+
 	var userVals map[string]cty.Value
-	var diags hcl.Diagnostics
-	if exists {
-		userVals, diags = LoadValuesFile(lokocfgVarsPath)
+	if varsExist {
+		userVals, diags = readValues(varPath)
```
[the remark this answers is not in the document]
johananl: I've changed `readValues` to `readValuesFile`. AFAICT both of these names are better than `LoadValuesFile` since "load" could mean many things but "read" typically refers to copying data from some source (a file in this case) to memory.

`populateVars()` is a function I've broken down from `ReadHCL()` following your [feedback](https://github.com/kinvolk/lokomotive/pull/806#discussion_r470735192). AFAICT its name is accurate since "populate vars" is exactly what this function does: it accepts a set of variables and a set of user-provided values and attempts to populate each variable with the corresponding value or with a default.

Is this clearer? Do you have better names in mind?

pkg/config/config_internal_test.go
```diff
@@ -0,0 +1,148 @@
...
+			}
+
+			t.Cleanup(func() {
+				if err := os.RemoveAll(d); err != nil {
+					t.Logf("Could not remove temporary directory %q: %v", d, err)
+				}
+			})
+
+			for _, f := range test.files {
+				err = ioutil.WriteFile(filepath.Join(d, f.filename), f.content, 0600)
```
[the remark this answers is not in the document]
johananl: Not sure I understand the suggestion. Do you mean that we should split `bodyFromPath` into two functions - one which reads the files and another which parses the data?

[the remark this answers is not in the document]
johananl: I've looked into it and it made sense to me to separate reading from parsing. We now have the following:

- `func readFiles(names []string) (map[string][]byte, error)`
- `func parseHCLFiles(files map[string][]byte) (hcl.Body, hcl.Diagnostics)`

I've covered both with tests.

pkg/config/config_test.go
johananl: Is it? I've encountered it only recently.

johananl:
@invidian I've handled most of your feedback. I still need to edit the commits though.

pkg/config/config_test.go
invidian: Well, we do it only in couple of places, so I guess it's fine to leave it :smile: 
```
$ git grep -A 1 range pkg/**/*.go cmd/**/*.go test/**/*.go | grep -v range | grep := | grep -v '(' | grep -v '{'
test/monitoring/components_alerts_test.go-              tc := tc
test/monitoring/components_metrics_test.go-             tc := tc
test/monitoring/monitoring_test.go-             test := test
```

invidian:
>@invidian I've handled most of your feedback. I still need to edit the commits though.

Thanks @johananl! I'll have a look.

pkg/config/config.go
```diff
@@ -164,6 +132,47 @@ func LoadConfig(lokocfgPath, lokocfgVarsPath string) (*Config, hcl.Diagnostics)
...
+func parseHCLFiles(files map[string][]byte) (hcl.Body, hcl.Diagnostics) {
+	parser := hclparse.NewParser()
+
+	hclFiles := []*hcl.File{}
+
+	var diags hcl.Diagnostics
+
+	for name, data := range files {
+		hclFile, d := parser.ParseHCL(data, name)
+		if len(d) > 0 {
```
invidian, line 162: On a fence here, as we could do either :smile: 
```suggestion
		if d.HasErrors() {
```

Or actually, perhaps you can always append. If diags are empty, nothing gets appended.

pkg/config/config.go
```diff
@@ -195,13 +204,68 @@ func evalFuncFile() function.Function {
...
 		}
+
+		res[v.Name] = dv
+	}
+
+	return res, nil
+}
+
+// diag returns an hcl.Diagnostics with a single hcl.Diagnostic based on the provided error string.
+func diag(err string) hcl.Diagnostics {
```
invidian, line 264: Maybe it would be better to accept `error` here, rather than `string`, as we always return diagnostic with `Severity: hcl.DiagError` ?
```suggestion
func diag(err error) hcl.Diagnostics {
```

pkg/config/config.go
```diff
@@ -195,13 +204,68 @@ func evalFuncFile() function.Function {
...
-func (c *Config) LoadComponentConfigBody(componentName string) *hcl.Body {
-	for _, component := range c.RootConfig.Components {
-		if componentName == component.Name {
-			return &component.Config
+// lokocfgFilenames returns a slice containing the names of all .lokocfg files found under path p.
+// If p is not a directory, p is returned by itself in the slice.
+func lokocfgFilenames(p string) ([]string, error) {
+	isDir, err := util.PathIsDir(p)
+	if err != nil {
+		return nil, fmt.Errorf("failed to stat config path %q: %w", p, err)
```
invidian, line 212:
```suggestion
		return nil, fmt.Errorf("checking if %q is a directory: %w", p, err)
```

pkg/config/config_internal_test.go
```diff
@@ -0,0 +1,245 @@
...
+	for _, test := range tests {
+		// Pin test to avoid scope bugs: https://github.com/kyoh86/scopelint
+		test := test
+
+		t.Run(test.desc, func(t *testing.T) {
+			for _, f := range test.files {
+				p := filepath.Join(d, f)
+
+				_, err = os.Create(p)
+				if err != nil {
```
invidian, line 224:
```suggestion		
				if _, err = os.Create(p); err != nil {
```

invidian (changes requested):
Thanks again for addressing my feedback @johananl, PR looks much better now. I commented on some small nits, but overall it looks nice.

This `populateVars` still bugs me and now I've noticed, that having `RootConfig` and `Config` structs in `pkg/config` also make little sense and it's related to `populateVars`. I'll have a look if I can come up with something to improve it, but I think it could be done in separate PR.

pkg/config/config_internal_test.go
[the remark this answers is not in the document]
invidian: Thanks for addressing that, I think it's much more readable right now.

pkg/config/config_internal_test.go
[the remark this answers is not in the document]
invidian: Yeah, I read the tests again and they seem all right, resolving.

pkg/config/config.go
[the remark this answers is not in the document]
invidian: `readValuesFile` sounds indeed better :+1:

pkg/config/config.go
```diff
@@ -61,91 +61,59 @@ type Config struct {
...
+	files, err := readFiles(names)
+	if err != nil {
+		return nil, diag(err.Error())
 	}
 
-	configBody := hcl.MergeFiles(hclFiles)
+	body, diags := parseHCLFiles(files)
+	if diags.HasErrors() {
+		return nil, diags
+	}
```
invidian, line 92: To avoid such lengthy functions and to name the blocks of code, I think we could put this three actions into separate function, but I'm not sure how to phrase the name of this function. However, after looking at this code (and code after it), it's clear to me, that this part reads the configuration files (and not values files) and returns them in a form of HCL body.

invidian (changes requested):
@johananl I looked at the code around `populateVars()` and I think I understand now why it was confusing.

I would like to propose following changes on top of yours: https://github.com/kinvolk/lokomotive/commit/8c76770cae2af3ed53b9733af3967736ac11fe72

They split `ReadHCL` further, so we can have better named code blocks (as mentioned in the comment) and changes `populateVars()` into "convert to Map" + "merge maps", which I think is more idiomatic approach than doing both things at the same time, as `populateVars()` do right now.

pkg/config/config_test.go
johananl: I would actually rather do without the comment but I thought people would be confused by this strange assignment.

johananl: I removed these comments. The linter complains when you don't pin these vars so I think it's obvious enough.

pkg/config/config.go
johananl: Removed the `if`.

pkg/config/config.go
johananl: Yes, somehow I thought we'd have a problem in cases where we want to format the error string, but we can just wrap with `fmt.Errorf()` before calling `diag()`. Fixed.

pkg/config/config.go
```diff
@@ -61,91 +61,59 @@ type Config struct {
...
-	return lokocfgPaths, nil
+
+	return nil
 }
 
-func LoadConfig(lokocfgPath, lokocfgVarsPath string) (*Config, hcl.Diagnostics) {
-	lokocfgPaths, err := loadLokocfgPaths(lokocfgPath)
+// ReadHCL reads all HCL files at path and one HCL variables file at varPath, constructs a Config
+// and returns a pointer to it.
+func ReadHCL(path, varPath string) (*Config, hcl.Diagnostics) {
```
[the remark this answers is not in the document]
johananl: Right. Fixed.

pkg/config/config.go
johananl: I don't think `ReadHCL()` is so long at the moment, and breaking it down further would start to be on the "too much indirection" side. *However*, looking at the function again I'm not happy with the name `ReadHCL` since this function does more than just read HCL files. Frankly I think `config.New()` was the best option from the discussions so far. Do you have an alternative which doesn't include the word "load"? If not, I'd just go with `config.New()`.

invidian: >I don't think ReadHCL() is so long at the moment, and breaking it down further would start to be on the "too much indirection" side.

Given that this is a exported function, so we should be able to cover it fully with tests, currently has 9 different paths, which IMO is to high cognitive complexity. In https://github.com/kinvolk/lokomotive/commit/8c76770cae2af3ed53b9733af3967736ac11fe72 I broke it down into nice smaller functions, so we have only 4 paths, which is very reasonable and easy to understand. This is also what tools like https://codeclimate.com/ encourages to do.

I think with the right naming you don't create the indirection, quite the opposite. You hide unrelated details, which makes things simpler and allows understanding only the paths of code which are interesting for the reader. With long function, reader must go trough all of it and understand all the statements to analyze the function. 

>However, looking at the function again I'm not happy with the name ReadHCL since this function does more than just read HCL files. Frankly I think config.New() was the best option from the discussions so far. Do you have an alternative which doesn't include the word "load"? If not, I'd just go with config.New().

`FromHCLFiles()` maybe? Or `ReadHCLFiles()` ? It gives you configuration from HCL files, so I think the name is appropriate. `config.New()` would be good IMO if we could define the loader type as an argument. Then the name shouldn't imply from where or how do we load the configuration. However, if we take HCL files paths as an argument, name should reflect that too in my opinion.

pkg/config/config.go
```diff
@@ -61,91 +61,59 @@ type Config struct {
...
-	return lokocfgPaths, nil
+
+	return nil
 }
 
-func LoadConfig(lokocfgPath, lokocfgVarsPath string) (*Config, hcl.Diagnostics) {
-	lokocfgPaths, err := loadLokocfgPaths(lokocfgPath)
+// ReadHCL reads all HCL files at path and one HCL values file at valuesPath, constructs a Config
+// and returns a pointer to it.
+func ReadHCL(path, valuesPath string) (*Config, hcl.Diagnostics) {
```
invidian, line 78: I still think this function is too long and should be splitted into smaller ones like I suggested in 8c76770

pkg/config/config.go
```diff
@@ -195,13 +201,68 @@ func evalFuncFile() function.Function {
...
+	}
+
+	return names, nil
+}
+
+// populateVars accepts a slice of variables and a map of user-provided values. The function
+// populates each variable with the corresponding user-provided value or with a default. If a
+// variable doesn't have a matching value and there is no default, an error is returned (as
+// hcl.Diagnostics).
+func populateVars(vars []variable, values map[string]cty.Value) (map[string]cty.Value, hcl.Diagnostics) {
```
invidian, line 230: >@johananl I looked at the code around populateVars() and I think I understand now why it was confusing.

>I would like to propose following changes on top of yours: 8c76770

>They split ReadHCL further, so we can have better named code blocks (as mentioned in the comment) and changes >populateVars() into "convert to Map" + "merge maps", which I think is more idiomatic approach than doing both things >at the same time, as populateVars() do right now.

invidian (changes requested):

pkg/config/config.go
invidian: @johananl I'm currently looking into untangling and isolating code from `cli/cmd` package, as part of #312 and at the moment I find it confusing, that we call `config.LoadConfig()`, but it doesn't really load entire configuration. It's even hard to define of what exactly this code is doing. Maybe that will help a bit with naming and documentation here.

invidian: This is the definition of existing `LoadConfig` function I came up with:
```go
// LoadConfig reads and parses top-level HCL configuration from given lokocfgPath and
// evaluates the variables from lokocfgVarsPath into HCL context, so it can be used
// for futher configuration parsing.
```

Which leads to the conclusion, that what is being returned is not the configuration (as "user configuration"), but rather a facility, which allows you to further read the configuration.

The `lokomotive/pkg/config.Config` struct does not ship proper information about itself. It's not "a Lokomotive configuration". It does not represent cluster + components configuration (which we didn't name yet).

PR #812, opened 2020-08-18
Add golangci-lint config file

### LINKED ISSUE #786: Move golangci-lint config to a file
We should consider managing our `golangci-lint` configuration in a [file](https://golangci-lint.run/usage/configuration/#config-file).

Doing so has the following advantages:

- Should allow integrating with VS Code and other editors.
- Visually nicer than CLI arguments inside `Makefile`.

knrt10:
Fixes #786

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,82 @@
+run:
+  build-tags: 
+    - aws
+    - packet
+    - aks
+    - e2e
+    - baremetal
+    - disruptivee2e
+    - poste2e
+
+  modules-download-mode: vendor
+
+  timeout: 5m
+
+issues:
+  exclude-use-default: false
+  max-same-issues: 0
+  max-issues-per-linter: 0
+
+# List for enabled linters can be generated for updates using the following command.
+# golangci-lint linters | grep -v -E '(by your configuration linters|^$)' | cut -d: -f1 | cut -d ' ' -f1 | sort | sed 's/^/    - /g'
+linters:
+  disable-all: false
+  enable:
+    - asciicheck
+    - bodyclose
+    - deadcode
+    - depguard
+    - dogsled
+    - dupl
+    - errcheck
+    - exhaustive
+    - funlen
+    - gochecknoinits
+    - gocognit
+    - goconst
+    - gocritic
+    - gocyclo
+    - godot
+    - gofmt
+    - gofumpt
+    - goimports
+    - golint
+    - gomnd
+    - goprintffuncname
+    - gosec
+    - gosimple
+    - govet
+    - ineffassign
+    - interfacer
+    - lll
+    - maligned
+    - misspell
+    - nakedret
+    - nestif
+    - noctx
+    - nolintlint
+    - prealloc
+    - rowserrcheck
+    - scopelint
+    - staticcheck
+    - structcheck
+    - stylecheck
+    - testpackage
+    - typecheck
+    - unconvert
+    - unparam
+    - unused
+    - varcheck
+    - whitespace
+    - wsl
+
+  disable:
+    - exportloopref
+    - gci
+    - gochecknoglobals
+    - godox
+    - goerr113
+    - goheader
+    - gomodguard
+    - nlreturn
+    - sqlclosecheck
--- a/Makefile
+++ b/Makefile
@@ -62,7 +62,7 @@ test: run-unit-tests
 .PHONY: lint
 lint: build-slim build-test
 	@if [ "$$(git config --get diff.noprefix)" = "true" ]; then printf "\n\ngolangci-lint has a bug and can't run with the current git configuration: 'diff.noprefix' is set to 'true'. To override this setting for this repository, run the following command:\n\n'git config diff.noprefix false'\n\nFor more details, see https://github.com/golangci/golangci-lint/issues/948.\n\n\n"; exit 1; fi
-	golangci-lint run --enable-all --disable=gci,godox,gochecknoglobals,goerr113 --max-same-issues=0 --max-issues-per-linter=0 --build-tags $(ALL_BUILD_TAGS) --new-from-rev=$$(git merge-base $$(cat .git/resource/base_sha 2>/dev/null || echo "origin/master") HEAD) --modules-download-mode=$(MOD) --timeout=5m --exclude-use-default=false ./...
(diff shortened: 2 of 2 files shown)


.golangci.yml
```diff
@@ -0,0 +1,69 @@
...
+    - disruptivee2e
+    - poste2e
+
+  modules-download-mode: vendor
+
+  timeout: 5m
+
+issues:
+  exclude-use-default: false
+  new-from-rev: HEAD~
```
invidian, line 17: This won't work, as it will only be checking last commit and we want to check everything on top of master branch.

.golangci.yml
```diff
@@ -0,0 +1,69 @@
...
+  max-same-issues: 0
+  max-issues-per-linter: 0
+
+linters:
+  disable-all: true
+  enable:
```
invidian, line 23: How do we update this list to include new linters?

.golangci.yml
```diff
@@ -0,0 +1,69 @@
...
+  build-tags: 
+    - aws
+    - packet
+    - aks
+    - e2e
+    - baremetal
+    - disruptivee2e
+    - poste2e
+
+  modules-download-mode: vendor
```
invidian, line 11: With this patch, it is no longer possible to build using different modules (not something we would do anyway, but pointing out reduced flexibility).

.golangci.yml
```diff
@@ -0,0 +1,69 @@
+run:
+  build-tags: 
+    - aws
```
invidian, line 3: Hm, with this, now we have to update the tags in 2 places :/

.golangci.yml
```diff
@@ -0,0 +1,69 @@
...
+  timeout: 5m
+
+issues:
+  exclude-use-default: false
+  new-from-rev: HEAD~
+  max-same-issues: 0
+  max-issues-per-linter: 0
+
+linters:
+  disable-all: true
```
invidian, line 22: Rather than `disable-all`, could we have a disable list? Or commented out linters in `enable` list, which are disabled, so it is easy to find, which ones are disabled.

johananl:
> (It should've been a comment, instead of change request).
> 
> To be honest, I don't see much of a gain from switching from CLI options to the config file.
> 
> Perhaps we can leave the CLI options we use as they are and add config file "below" it, so it satisfies @johananl usecase?

The rationale is explained in #786 (easier to read compared to args, ability to integrate with IDEs).

invidian:
>The rationale is explained in #786 (easier to read compared to args, ability to integrate with IDEs).

I'm aware of that. What I meant is, when I compare the rationale and implementation issues/limitations, IN my opinion issues outweighs the rationale, this is why I proposed an alternative (config file + CLI flags), as in my opinion configuration file is currently not able to express our needs and I'd prefer to avoid changing the setup we currently have.

.golangci.yml
knrt10: yes

.golangci.yml
knrt10: Just add new linters to list I think. If you are asking for new linters on new updates, we check it via `golangci-lint linters`. Also previously we were using `--enable-all` which is deprecated and will soon be removed. So I guess we had to move to config file

.golangci.yml
knrt10: Cool, will add disabled linters list.

knrt10:
Updated PR

.golangci.yml
invidian: I'd put a comment, that the list can be generated for updates using the following command:
```sh
golangci-lint linters | grep -v -E '(by your configuration linters|^$)' | cut -d: -f1 | cut -d ' ' -f1 | sort | sed 's/^/    - /g'
```

.golangci.yml
```diff
@@ -0,0 +1,80 @@
...
+    - stylecheck
+    - testpackage
+    - typecheck
+    - unconvert
+    - unparam
+    - unused
+    - varcheck
+    - whitespace
+    - wsl
+    # - exportloopref
```
invidian, line 70: What's up with those? Why they are commented out?

knrt10: Oh man, I had mixed mind regarding adding then to disabled list or commenting them, after reading your comment 🤦. Fixing it

Makefile
```diff
@@ -59,10 +59,12 @@ build-slim:
 .PHONY: test
 test: run-unit-tests
 
+# List for enabled linters can be generated for updates using the following command.
+# golangci-lint linters | grep -v -E '(by your configuration linters|^$)' | cut -d: -f1 | cut -d ' ' -f1 | sort | sed 's/^/    - /g'
```
invidian, line 63: I'd rather put this comment in `.golangci.yml` file, right before the list.

Makefile
```diff
@@ -59,10 +59,12 @@ build-slim:
...
 .PHONY: lint
 lint: build-slim build-test
 	@if [ "$$(git config --get diff.noprefix)" = "true" ]; then printf "\n\ngolangci-lint has a bug and can't run with the current git configuration: 'diff.noprefix' is set to 'true'. To override this setting for this repository, run the following command:\n\n'git config diff.noprefix false'\n\nFor more details, see https://github.com/golangci/golangci-lint/issues/948.\n\n\n"; exit 1; fi
-	golangci-lint run --enable-all --disable=gci,godox,gochecknoglobals,goerr113 --max-same-issues=0 --max-issues-per-linter=0 --build-tags $(ALL_BUILD_TAGS) --new-from-rev=$$(git merge-base $$(cat .git/resource/base_sha 2>/dev/null || echo "origin/master") HEAD) --modules-download-mode=$(MOD) --timeout=5m --exclude-use-default=false ./...
+	golangci-lint run --new-from-rev=$$(git merge-base $$(cat .git/resource/base_sha 2>/dev/null || echo "origin/master") HEAD) --config ./.golangci.yml --verbose ./...
```
invidian, line 67: I think we don't need to use `--config` flag explicitly, it should be picked up automatically.

knrt10:
Updated

invidian (commented):
LGTM, but I leave the approval to @johananl to confirm that it works for him as expected :)

johananl (approved):
LGTM. VS Code integration works :clap: Other integrations are supported, too: https://golangci-lint.run/usage/integrations/

The only thing I've noticed is that now `make lint` is more verbose and logs many lines with `[INFO]`. I don't have a problem with that, just pointing out we didn't have that before.

invidian:
@knrt10 Did you add `--verbose` on purpose? If so, can you please elaborate why?

PR #831, opened 2020-08-23
Use fmt.Errorf instead of errors.Wrap where error handling make sense

invidian:
This commit partially addresses issue #59, which encourages to use new
error wrapping semantics available since Go 1.13, using %w formating
verb instead of 3rd party 'pkg/errors' package.

This commit changes errors handling code from using errors.Wrap with
fmt.Errorf where error handling should remain as is. For other uses of
errors.Wrap, it adds a comment with TODO how code should be changed to
either avoid or delegate error handling to someone else.

### DIFF
--- a/pkg/components/external-dns/component.go
+++ b/pkg/components/external-dns/component.go
@@ -20,7 +20,6 @@ import (
 
 	"github.com/hashicorp/hcl/v2"
 	"github.com/hashicorp/hcl/v2/gohcl"
-	"github.com/pkg/errors"
 
 	"github.com/kinvolk/lokomotive/internal/template"
 	"github.com/kinvolk/lokomotive/pkg/components"
@@ -118,7 +117,7 @@ func (c *component) RenderManifests() (map[string]string, error) {
 	if c.AwsConfig.AccessKeyID == "" {
 		accessKeyID, ok := os.LookupEnv("AWS_ACCESS_KEY_ID")
 		if !ok || accessKeyID == "" {
-			return nil, errors.New("AWS Credentials not found.")
+			return nil, fmt.Errorf("AWS access key ID not found")
 		}
 		c.AwsConfig.AccessKeyID = accessKeyID
 	}
@@ -126,19 +125,19 @@ func (c *component) RenderManifests() (map[string]string, error) {
 	if c.AwsConfig.SecretAccessKey == "" {
 		secretAccessKey, ok := os.LookupEnv("AWS_SECRET_ACCESS_KEY")
 		if !ok || secretAccessKey == "" {
-			return nil, errors.New("AWS Credentials not found.")
+			return nil, fmt.Errorf("AWS secret access key not found")
 		}
 		c.AwsConfig.SecretAccessKey = secretAccessKey
 	}
 
 	values, err := template.Render(chartValuesTmpl, c)
 	if err != nil {
-		return nil, errors.Wrap(err, "render chart values template")
+		return nil, fmt.Errorf("rendering chart values template: %w", err)
 	}
 
 	renderedFiles, err := util.RenderChart(helmChart, name, c.Namespace, values)
 	if err != nil {
-		return nil, errors.Wrap(err, "render chart")
+		return nil, fmt.Errorf("rendering chart: %w", err)
 	}
 
 	return renderedFiles, nil
--- a/pkg/components/gangway/component.go
+++ b/pkg/components/gangway/component.go
@@ -308,6 +308,7 @@ func (c *component) LoadConfig(configBody *hcl.Body, evalContext *hcl.EvalContex
 	return gohcl.DecodeBody(*configBody, evalContext, c)
 }
 
+// TODO: Convert to Helm chart.
 func (c *component) RenderManifests() (map[string]string, error) {
 	tmpl, err := template.New("config-map").Parse(configMapTmpl)
 	if err != nil {
--- a/pkg/components/metallb/component.go
+++ b/pkg/components/metallb/component.go
@@ -15,6 +15,8 @@
 package metallb
 
 import (
+	"fmt"
+
 	"github.com/hashicorp/hcl/v2"
 	"github.com/hashicorp/hcl/v2/gohcl"
 	"github.com/pkg/errors"
@@ -54,6 +56,7 @@ func (c *component) LoadConfig(configBody *hcl.Body, evalContext *hcl.EvalContex
 	return gohcl.DecodeBody(*configBody, evalContext, c)
 }
 
+// TODO: Convert to Helm chart.
 func (c *component) RenderManifests() (map[string]string, error) {
 	// Here are `nodeSelectors` and `tolerations` that are set by upstream. To make sure that we
 	// don't miss them out we set them manually here. We cannot make these changes in the template
@@ -78,7 +81,7 @@ func (c *component) RenderManifests() (map[string]string, error) {
 
 	t, err := util.RenderTolerations(c.SpeakerTolerations)
 	if err != nil {
-		return nil, errors.Wrap(err, "failed to marshal speaker tolerations")
+		return nil, fmt.Errorf("marshaling speaker tolerations: %w", err)
 	}
 	c.SpeakerTolerationsJSON = t
 
--- a/pkg/helm/charts.go
+++ b/pkg/helm/charts.go
@@ -16,11 +16,11 @@
 package helm
 
 import (
+	"fmt"
 	"io/ioutil"
 	"os"
 
 	"github.com/kinvolk/lokomotive/pkg/assets"
-	"github.com/pkg/errors"
 	"helm.sh/helm/v3/pkg/chart"
 	"helm.sh/helm/v3/pkg/chart/loader"
 )
@@ -39,7 +39,7 @@ type LokomotiveChart struct {
 func ChartFromAssets(location string) (*chart.Chart, error) {
 	tmpDir, err := ioutil.TempDir("", "lokoctl-chart-")
 	if err != nil {
-		return nil, errors.Wrap(err, "creating temporary dir")
+		return nil, fmt.Errorf("creating temporary directory: %w", err)
 	}
 
 	// TODO: os.RemoveAll() returns an error which we currently don't handle. Handling the error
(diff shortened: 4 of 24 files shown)


ipochi:
Isn't @johananl changing those in the files related to platform package as part of his refactoring PR ?

invidian:
>Isn't @johananl changing those in the files related to platform package as part of his refactoring PR ?

I'm not sure how is this related. The platform package refactoring PR is not even opened yet, but still it's focus would be different.

knrt10:
@invidian should we add //TODOs in the same commit?

invidian:
>@invidian should we add //TODOs in the same commit?

I can't come up with any pro or con for committing it separately vs committing it together, maybe except that doing it together is a bit less work.

pkg/helm/charts.go
```diff
@@ -39,7 +39,7 @@ type LokomotiveChart struct {
 func ChartFromAssets(location string) (*chart.Chart, error) {
 	tmpDir, err := ioutil.TempDir("", "lokoctl-chart-")
 	if err != nil {
-		return nil, errors.Wrap(err, "creating temporary dir")
+		return nil, fmt.Errorf("creating temporariy directory: %w", err)
```
knrt10, line 42: Cool, I understand. No need to add it in a seperate commit, it will be extra work.

`temporary`

pkg/components/external-dns/component.go
```diff
@@ -118,27 +117,27 @@ func (c *component) RenderManifests() (map[string]string, error) {
...
-			return nil, errors.New("AWS Credentials not found.")
+			return nil, fmt.Errorf("AWS Secret Access key not found")
 		}
 		c.AwsConfig.SecretAccessKey = secretAccessKey
 	}
 
 	values, err := template.Render(chartValuesTmpl, c)
 	if err != nil {
-		return nil, errors.Wrap(err, "render chart values template")
+		return nil, fmt.Errorf("rednering chart values template: %w", err)
```
knrt10, line 135: `rendering`

knrt10 (changes requested):
Just a small nit. Rest LGTM!

pkg/helm/charts.go
invidian: Nice catches, thanks! That was a lot of typing :smile: Fixed.

pkg/components/external-dns/component.go
invidian: Fixed, thanks :)

knrt10 (approved):
LGTM +1

pkg/components/external-dns/component.go
```diff
@@ -118,27 +117,27 @@ func (c *component) RenderManifests() (map[string]string, error) {
 	if c.AwsConfig.AccessKeyID == "" {
 		accessKeyID, ok := os.LookupEnv("AWS_ACCESS_KEY_ID")
 		if !ok || accessKeyID == "" {
-			return nil, errors.New("AWS Credentials not found.")
+			return nil, fmt.Errorf("AWS Access Key ID not found")
```
johananl, line 120: Is it valid to use `fmt.Errorf()` when no string formatting is used?

pkg/components/gangway/component.go
```diff
@@ -308,6 +308,7 @@ func (c *component) LoadConfig(configBody *hcl.Body, evalContext *hcl.EvalContex
 	return gohcl.DecodeBody(*configBody, evalContext, c)
 }
 
+// TODO: Convert to helm chart.
```
johananl, line 311:
```suggestion
// TODO: Convert to Helm chart.
```

If you change this, please fix the rest of the occurrences as well.

pkg/components/metallb/component.go
```diff
@@ -78,7 +81,7 @@ func (c *component) RenderManifests() (map[string]string, error) {
 
 	t, err := util.RenderTolerations(c.SpeakerTolerations)
 	if err != nil {
-		return nil, errors.Wrap(err, "failed to marshal speaker tolerations")
+		return nil, fmt.Errorf("marshalling speaker tolerations: %w", err)
```
johananl, line 84: Both "marshaling" and "marshalling" seem to be valid spellings. Go uses the former in the standard library. We currently have only one occurrence: https://github.com/kinvolk/lokomotive/blob/a175b3736d68954fbd4f9fc18c5f19accea62ac7/test/calico/calico_test.go#L72

I'm fine with either as long as we're consistent. "Marshalling" seems to be in wider use historically: https://books.google.com/ngrams/graph?smoothing=3&corpus=26&year_end=2019&content=marshalling%2Cmarshaling&year_start=1800&direct_url=t1%3B%2Cmarshalling%3B%2Cc0%3B.t1%3B%2Cmarshaling%3B%2Cc0

pkg/install/verify.go
```diff
@@ -58,7 +56,7 @@ func Verify(cl *lokomotive.Cluster) error {
 		return ns.Ready(), nil // Retry if not ready
 	})
 	if nsErr != nil {
-		return errors.Wrapf(nsErr, "error determining node status within the allowed time")
+		return fmt.Errorf("determining node status within allowed time: %w", nsErr)
```
johananl, line 59: I find "determining node status within allowed time" strange. IIUC this error is the *last* error received from `GetNodeStatus()` after giving up due to timer expiration. I'd consider using the following phrasing:

`return fmt.Errorf("waiting for nodes: %w", nsErr)`

pkg/platform/aws/aws.go
```diff
@@ -180,23 +180,26 @@ func createTerraformConfigFile(cfg *config, terraformRootDir string) error {
...
 	f, err := os.Create(path)
 	if err != nil {
-		return errors.Wrapf(err, "failed to create file %q", path)
+		return fmt.Errorf("creating file %q: %w", path, err)
 	}
 	defer f.Close()
 
 	keyListBytes, err := json.Marshal(cfg.SSHPubKeys)
 	if err != nil {
+		// TODO: Render them manually instead of marshalling.
```
johananl, line 196: Can we replace "them" with something concrete to avoid ambiguity?

pkg/platform/baremetal/baremetal.go
```diff
@@ -143,58 +144,67 @@ func createTerraformConfigFile(cfg *config, terraformPath string) error {
...
 	}
+
 	defer f.Close()
 
 	// Configure oidc flags and set it to KubeAPIServerExtraFlags.
 	if cfg.OIDC != nil {
 		// Skipping the error checking here because its done in checkValidConfig().
 		oidcFlags, _ := cfg.OIDC.ToKubeAPIServerFlags(cfg.K8sDomainName)
-		//TODO: Use append instead of setting the oidcFlags to KubeAPIServerExtraFlags
+		// TODO: Use append instead of setting the oidcFlags to KubeAPIServerExtraFlags
```
johananl, line 163:
```suggestion
		// TODO: Use append instead of setting oidcFlags to KubeAPIServerExtraFlags.
```

I find "the" before a var name confusing.

johananl (commented):
Thanks. Added some comments.

pkg/platform/baremetal/baremetal.go
invidian: OK, though I didn't really touch that line, only because linter was complaining `//TODO:` vs `// TODO:` :smile:

pkg/components/metallb/component.go
invidian: I'll go with just `marshaling`, as we have more occurrences and our vendor directory also have more:
```
0 ✓ (637ms) 14:07:55 invidian@dellxps15mateusz ~/repos/kinvolk/lokomotive (invidian/errors-wrap)*$ git grep marshaling vendor/ | wc -l
138
0 ✓ (76.3ms) 14:07:56 invidian@dellxps15mateusz ~/repos/kinvolk/lokomotive (invidian/errors-wrap)*$ git grep marshalling vendor/ | wc -l
48
```

pkg/components/external-dns/component.go
invidian: I didn't find any linter or rule being against it. Otherwise we would have to still import 3rd part `errors` for `errors.New`. This way, we have consistent convention using core library.

pkg/install/verify.go
invidian: Changed as you suggested.

pkg/platform/aws/aws.go
invidian: I removed it to be consistent with other TODOs. Now it's just `// TODO: Render manually instead of marshalling.`.

invidian:
Thanks for reviewing @johananl, I addressed all your comments, please have a look again.

pkg/platform/baremetal/baremetal.go
johananl: Not looking for suspects, just want to have cleaner errors while we're at it :-)

pkg/platform/baremetal/baremetal.go
```diff
@@ -143,58 +144,67 @@ func createTerraformConfigFile(cfg *config, terraformPath string) error {
...
 	}
+
 	defer f.Close()
 
 	// Configure oidc flags and set it to KubeAPIServerExtraFlags.
 	if cfg.OIDC != nil {
 		// Skipping the error checking here because its done in checkValidConfig().
 		oidcFlags, _ := cfg.OIDC.ToKubeAPIServerFlags(cfg.K8sDomainName)
-		//TODO: Use append instead of setting the oidcFlags to KubeAPIServerExtraFlags
+		// TODO: Use append instead of setting oidcFlags to KubeAPIServerExtraFlags
```
johananl, line 163:
```suggestion
		// TODO: Use append instead of setting oidcFlags to KubeAPIServerExtraFlags.
```

pkg/platform/packet/packet.go
```diff
@@ -220,23 +220,26 @@ func createTerraformConfigFile(cfg *config, terraformPath string) error {
 	t := template.New(tmplName)
 	t, err := t.Parse(terraformConfigTmpl)
 	if err != nil {
+		// TODO: Use template.Must()
```
johananl, line 223:
```suggestion
		// TODO: Use template.Must().
```

johananl (dismissed):
Found two missing periods. LGTM otherwise.

pkg/platform/baremetal/baremetal.go
invidian: If the dot, I also need to capitalize next line :smile: done.

johananl (approved):

PR #840, opened 2020-08-25
pkg/lokomotive: make NewCluster to not return error

invidian:
As we never do that anyway.

### DIFF
--- a/pkg/lokomotive/cluster.go
+++ b/pkg/lokomotive/cluster.go
@@ -31,6 +31,7 @@ type Cluster struct {
 	ExpectedNodes int
 }
 
+// NewCluster constructs and returns a new Cluster object.
 func NewCluster(client *kubernetes.Clientset, expectedNodes int) *Cluster {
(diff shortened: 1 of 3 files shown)


knrt10 (dismissed):
+1

pkg/lokomotive/cluster.go
```diff
@@ -31,6 +31,7 @@ type Cluster struct {
 	ExpectedNodes int
 }
 
+// NewCluster constructs and returns new Cluster object.
```
iaguis, line 34:
```suggestion
// NewCluster constructs and returns a new Cluster object.
```

invidian: Thanks, fixed.

iaguis (approved):

knrt10 (dismissed):
Hoping for AKS to pass too. LGTM

iaguis (dismissed):

knrt10 (approved):
+1

PR #849, opened 2020-08-25
pkg/platform/aks: update Kubernetes version to 1.17.9

invidian:


pkg/platform/aks/aks.go
```diff
@@ -76,7 +76,7 @@ const (
 	subscriptionIDEnv = "LOKOMOTIVE_AKS_SUBSCRIPTION_ID"
 	tenantIDEnv       = "LOKOMOTIVE_AKS_TENANT_ID"
 
-	kubernetesVersion = "1.16.10"
```
ipochi, line 79: is that first available jump ? No other minor version before `1.17.9` ?

invidian: There are following version available for updates: `1.16.13, 1.17.7, 1.17.9`, but I think we should just go to the latest available.

knrt10 (approved):
+1

surajssd (approved):
ok

pkg/platform/aks/aks.go
ipochi: ok.

ipochi (approved):
Thanks for the PR @invidian 
LGTM.

PR #880, opened 2020-08-28
Add autocomplete for bash and zsh in lokoctl

### LINKED ISSUE #18: Add Bash/zsh autocompletion
Cobra [supports](https://github.com/spf13/cobra#generating-shell-completions) generating autocomplete files for Bash and zsh. We could leverage this to make working with lokoctl nicer.

knrt10:
Fixes #18

### DIFF
--- a/cli/cmd/completion.go
+++ b/cli/cmd/completion.go
@@ -1,0 +1,288 @@
+// Copyright 2020 The Lokomotive Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cmd
+
+import (
+	"fmt"
+	"io"
+	"os"
+
+	"github.com/spf13/cobra"
+)
+
+const completionDesc = `  Generate the completion code for lokoctl for the specified shell (Bash or zsh).
+`
+
+const completionExample = `  # Load the lokoctl completion code for Bash into the current shell.
+  source <(lokoctl completion bash)
+
+  # Load the lokoctl completion code for zsh into the current shell.
+  source <(lokoctl completion zsh)
+
+  # Generate a Bash completion file and load it for every shell.
+  lokoctl completion bash > ~/.bash_lokoctl_completion
+  echo "source ~/.bash_lokoctl_completion" >> ~/.bashrc && source ~/.bashrc
+
+  # Set the lokoctl completion code for zsh to autoload on startup.
+  lokoctl completion zsh > "${fpath[1]}/_lokoctl" && exec $SHELL`
+
+const bashCompDesc = `  Generate the completion code for lokoctl for the Bash shell.
+`
+
+const bashExample = `  # If running Bash 3.2 that is included with macOS, install Bash completion using Homebrew.
+  brew install bash-completion
+	
+  # If running Bash 4.1+ on macOS, install Bash completion using homebrew.
+  brew install bash-completion@2
+
+  # Load the lokoctl completion code for Bash into the current shell.
+  source <(lokoctl completion bash)
+
+  # Generate a Bash completion file and load it for every shell.
+  lokoctl completion bash > ~/.bash_lokoctl_completion
+  echo "source ~/.bash_lokoctl_completion" >> ~/.bashrc && source ~/.bashrc
+`
+
+const zshCompDesc = `  Generate the completion code for lokoctl for the zsh shell.
+`
+
+const zshExample = `  # Load the lokoctl completion code for zsh into the current shell.
+  source <(lokoctl completion zsh)
+
+  # Set the lokoctl completion code for zsh to autoload on startup.
+  lokoctl completion zsh > "${fpath[1]}/_lokoctl" && exec $SHELL
+`
+
+const zshInitialization = `#compdef lokoctl
+__lokoctl_bash_source() {
+	alias shopt=':'
+	alias _expand=_bash_expand
+	alias _complete=_bash_comp
+	emulate -L sh
+	setopt kshglob noshglob braceexpand
+	source "$@"
+}
+__lokoctl_type() {
+	# -t is not supported by zsh.
+	if [ "$1" == "-t" ]; then
+		shift
+		# Fake Bash 4 to disable "complete -o nospace". Instead
+		# "compopt +-o nospace" is used in the code to toggle trailing
+		# spaces. We don't support that, but leave trailing spaces on
+		# all the time.
+		if [ "$1" = "__lokoctl_compopt" ]; then
+			echo builtin
+			return 0
+		fi
+	fi
+	type "$@"
+}
+__lokoctl_compgen() {
+	local completions w
+	completions=( $(compgen "$@") ) || return $?
+	# Filter by given word as prefix.
+	while [[ "$1" = -* && "$1" != -- ]]; do
+		shift
+		shift
+	done
+	if [[ "$1" == -- ]]; then
+		shift
+	fi
+	for w in "${completions[@]}"; do
+		if [[ "${w}" = "$1"* ]]; then
+			# Use printf instead of echo because it is possible that
+			# the value to print is -n, which would be interpreted
+			# as a flag to echo.
+			printf "%s\n" "${w}"
+		fi
+	done
+}
+__lokoctl_compopt() {
+	true # Don't do anything. Not supported by bashcompinit in zsh.
+}
+__lokoctl_ltrim_colon_completions()
+{
+	if [[ "$1" == *:* && "$COMP_WORDBREAKS" == *:* ]]; then
+		# Remove colon-word prefix from COMPREPLY items.
+		local colon_word=${1%${1##*:}}
+		local i=${#COMPREPLY[*]}
+		while [[ $((--i)) -ge 0 ]]; do
+			COMPREPLY[$i]=${COMPREPLY[$i]#"$colon_word"}
+		done
+	fi
+}
+__lokoctl_get_comp_words_by_ref() {
+	cur="${COMP_WORDS[COMP_CWORD]}"
+	prev="${COMP_WORDS[${COMP_CWORD}-1]}"
+	words=("${COMP_WORDS[@]}")
+	cword=("${COMP_CWORD[@]}")
+}
+__lokoctl_filedir() {
+	local RET OLD_IFS w qw
+	__debug "_filedir $@ cur=$cur"
+	if [[ "$1" = \~* ]]; then
+		# Somehow does not work. Maybe, zsh does not call this at all.
+		eval echo "$1"
+		return 0
+	fi
+	OLD_IFS="$IFS"
+	IFS=$'\n'
+	if [ "$1" = "-d" ]; then
+		shift
+		RET=( $(compgen -d) )
+	else
+		RET=( $(compgen -f) )
+	fi
+	IFS="$OLD_IFS"
+	IFS="," __debug "RET=${RET[@]} len=${#RET[@]}"
+	for w in ${RET[@]}; do
+		if [[ ! "${w}" = "${cur}"* ]]; then
+			continue
+		fi
+		if eval "[[ \"\${w}\" = *.$1 || -d \"\${w}\" ]]"; then
+			qw="$(__lokoctl_quote "${w}")"
+			if [ -d "${w}" ]; then
+				COMPREPLY+=("${qw}/")
+			else
+				COMPREPLY+=("${qw}")
+			fi
+		fi
+	done
+}
+__lokoctl_quote() {
+	if [[ $1 == \'* || $1 == \"* ]]; then
+		# Leave out first character.
+		printf %q "${1:1}"
+	else
+		printf %q "$1"
+	fi
+}
+autoload -U +X bashcompinit && bashcompinit
+# Use word boundary patterns for BSD or GNU sed.
+LWORD='[[:<:]]'
+RWORD='[[:>:]]'
+if sed --help 2>&1 | grep -q 'GNU\|BusyBox'; then
+	LWORD='\<'
+	RWORD='\>'
+fi
+__lokoctl_convert_bash_to_zsh() {
+	sed \
+	-e 's/declare -F/whence -w/' \
+	-e 's/_get_comp_words_by_ref "\$@"/_get_comp_words_by_ref "\$*"/' \
+	-e 's/local \([a-zA-Z0-9_]*\)=/local \1; \1=/' \
+	-e 's/flags+=("\(--.*\)=")/flags+=("\1"); two_word_flags+=("\1")/' \
+	-e 's/must_have_one_flag+=("\(--.*\)=")/must_have_one_flag+=("\1")/' \
+	-e "s/${LWORD}_filedir${RWORD}/__lokoctl_filedir/g" \
+	-e "s/${LWORD}_get_comp_words_by_ref${RWORD}/__lokoctl_get_comp_words_by_ref/g" \
+	-e "s/${LWORD}__ltrim_colon_completions${RWORD}/__lokoctl_ltrim_colon_completions/g" \
+	-e "s/${LWORD}compgen${RWORD}/__lokoctl_compgen/g" \
+	-e "s/${LWORD}compopt${RWORD}/__lokoctl_compopt/g" \
+	-e "s/${LWORD}declare${RWORD}/builtin declare/g" \
+	-e "s/\\\$(type${RWORD}/\$(__lokoctl_type/g" \
+	-e 's/aliashash\["\(.\{1,\}\)"\]/aliashash[\1]/g' \
+	-e 's/FUNCNAME/funcstack/g' \
+	<<'BASH_COMPLETION_EOF'
+`
+
+const zshTail = `
+BASH_COMPLETION_EOF
+}
+__lokoctl_bash_source <(__lokoctl_convert_bash_to_zsh)
+`
+
+func newCompletionCmd(out io.Writer) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:               "completion",
+		Short:             "Generate the completion code for the specified shell",
+		Long:              completionDesc,
+		Example:           completionExample,
+		Args:              noArgs,
+		ValidArgsFunction: noCompletions, // Disable file completion.
+	}
+
+	bash := &cobra.Command{
+		Use:                   "bash",
+		Short:                 "Generate the completion code for Bash",
+		Long:                  bashCompDesc,
+		Example:               bashExample,
+		Args:                  noArgs,
+		DisableFlagsInUseLine: true,
+		ValidArgsFunction:     noCompletions,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runCompletionBash(out, cmd)
+		},
+	}
+
+	zsh := &cobra.Command{
+		Use:                   "zsh",
+		Short:                 "Generate the completion code for zsh",
+		Long:                  zshCompDesc,
+		Example:               zshExample,
+		Args:                  noArgs,
+		DisableFlagsInUseLine: true,
+		ValidArgsFunction:     noCompletions,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runCompletionZsh(out, cmd)
+		},
+	}
+
+	cmd.AddCommand(bash, zsh)
+
+	return cmd
+}
+
+func runCompletionBash(out io.Writer, cmd *cobra.Command) error {
+	return cmd.Root().GenBashCompletion(out)
+}
+
+func runCompletionZsh(out io.Writer, cmd *cobra.Command) error {
+	// TODO: Change the whole process to GenZshCompletion once
+	// cobra releases their new updates.
+	if _, err := out.Write([]byte(zshInitialization)); err != nil {
+		return fmt.Errorf("writing zsh initialization: %w", err)
+	}
+
+	if err := runCompletionBash(out, cmd); err != nil {
+		return fmt.Errorf("running Bash completion: %w", err)
+	}
+
+	if _, err := out.Write([]byte(zshTail)); err != nil {
+		return fmt.Errorf("writing zsh tail: %w", err)
+	}
+
+	return nil
+}
+
+// noCompletions is used to disable file completion.
+func noCompletions(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+	return nil, cobra.ShellCompDirectiveNoFileComp
+}
+
+// noArgs returns an error if any args are included.
+func noArgs(cmd *cobra.Command, args []string) error {
+	if len(args) > 0 {
+		return fmt.Errorf(
+			"%q accepts no arguments\n\nUsage:  %s",
+			cmd.CommandPath(),
+			cmd.UseLine(),
+		)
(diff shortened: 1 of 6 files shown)


cli/cmd/completion.go
```diff
@@ -0,0 +1,251 @@
...
+	cmd.AddCommand(bash, zsh)
+
+	return cmd
+}
+
+func runCompletionBash(out io.Writer, cmd *cobra.Command) error {
+	return cmd.Root().GenBashCompletion(out)
+}
+
+func runCompletionZsh(out io.Writer, cmd *cobra.Command) error { //nolint:funlen
```
invidian, line 88: Why not use `GenZshCompletion()` completion for zsh?

knrt10: I was using that before, but it flags were preferred over nouns. https://github.com/spf13/cobra/issues/1189

marckhouzam: Cobra 1.0.0 does not support `ValidArgsFunction` for `zsh`.  You will need to wait for the next release, or you can point to the master branch which has the full support required.

invidian: @knrt10 what's the diff between the version generated by cobra and this one? Can we maybe patch it dynamically to avoid hardcoding it?

docs/cli/lokoctl_completion_zsh.md
```diff
@@ -0,0 +1,40 @@
+## lokoctl completion zsh
+
+Generate autocompletions script for zsh
```
BrainBlasted, line 3: Here it's autocompletions, and below it's autocompletion. In keeping them consistent, I think it should be autocompletion here?

cli/cmd/completion.go
```diff
@@ -0,0 +1,251 @@
...
+
+	"To load completions in your current shell session: \n" +
+	"`$ source <(lokoctl completion zsh)` \n\n" +
+
+	"To load completions for every new session, execute once: \n" +
+	"`$ lokoctl completion zsh > \"${fpath[1]}/_lokoctl\" && exec $SHELL`"
+
+func newCompletionCmd(out io.Writer) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:               "completion",
```
BrainBlasted, line 48: Here as well, in addition to other places in the documentation.

cli/cmd/completion.go
marckhouzam: > what's the diff between the version generated by cobra and this one? Can we maybe patch it dynamically to avoid hardcoding it?

If you allow me to answer this one.
On cobra's master branch, there was a complete redesign of the zsh completion solution.  
It was a major change as you can see from:
https://github.com/spf13/cobra/pull/1070

Until it is released, the zsh completion (of cobra 1.0) is incomplete.

The alternative used in this pr is what `helm` and `kubectl` currently use for zsh, until the next cobra release.  This temporary solution is based on bash completion which is fully supported in cobra 1.0.

invidian: @marckhouzam thanks for the reply! And when do you plan 1.0 release roughly?

knrt10:
@invidian, so I think, currently go for this untill cobra releases it's new updates. Do you agree?

knrt10:
For 0.4 we can release it with this, if by 0.5 cobra releases their new updates, we can change it to `GenZshCompletion`

cli/cmd/completion.go
marckhouzam: @invidian I'm not a maintainer of the Cobra project but the suggestion of a new release has been mentioned. I'm hopeful it will happen in the next few weeks, but no guarantee.

invidian:
>For 0.4 we can release it with this, if by 0.5 cobra releases their new updates, we can change it to GenZshCompletion

Or we could do bash completion for 0.4 and zsh completion for the next release? What do you think @johananl? Or we could postpone this completely.

knrt10:
As mentioned in https://github.com/kinvolk/lokomotive/pull/880#discussion_r479638982 we don't know when the next cobra release will be, but if we do want to use `GenZshCompletion` one alternative is to use, `ValidArgs` instead of `ValidArgsFunction` and provide arguments explicitly for both cluster apply and delete.

cli/cmd/completion.go
```diff
@@ -0,0 +1,251 @@
...
+
+	"github.com/spf13/cobra"
+	"helm.sh/helm/v3/cmd/helm/require"
+)
+
+const completionDesc = `
+Generate autocompletion script for lokoctl for the specified shell (bash or zsh).
+`
+
+const bashCompDesc = "\nGenerate the autocompletion script for lokoctl for the bash shell. \n\n" +
```
invidian, line 29: Maybe we can use string literal here to make the formatting simpler? https://golang.org/ref/spec#String_literals

Also, it seems we could make that a function, to de-duplicate parts of the text.

cli/cmd/completion.go
```diff
@@ -0,0 +1,251 @@
...
+	"To load completions for every new session, execute once: \n" +
+	"`$ lokoctl completion zsh > \"${fpath[1]}/_lokoctl\" && exec $SHELL`"
+
+func newCompletionCmd(out io.Writer) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:               "completion",
+		Short:             "Generate autocompletion script for the specified shell",
+		Long:              completionDesc,
+		Args:              require.NoArgs,
+		ValidArgsFunction: noCompletions, // Disable file completion
```
invidian, line 52:
```suggestion
		ValidArgsFunction: noCompletions, // Disable file completion.
```

cli/cmd/completion.go
```diff
@@ -0,0 +1,251 @@
...
+
+const zshCompDesc = "\nGenerate the autocompletion script for lokoctl for the zsh shell. \n\n" +
+
+	"To load completions in your current shell session: \n" +
+	"`$ source <(lokoctl completion zsh)` \n\n" +
+
+	"To load completions for every new session, execute once: \n" +
+	"`$ lokoctl completion zsh > \"${fpath[1]}/_lokoctl\" && exec $SHELL`"
+
+func newCompletionCmd(out io.Writer) *cobra.Command {
```
invidian, line 46: Is this possible to mark parameter as not used?
```suggestion
func newCompletionCmd(_ io.Writer) *cobra.Command {
```

cli/cmd/completion.go
```diff
@@ -0,0 +1,251 @@
...
+
+	return cmd
+}
+
+func runCompletionBash(out io.Writer, cmd *cobra.Command) error {
+	return cmd.Root().GenBashCompletion(out)
+}
+
+func runCompletionZsh(out io.Writer, cmd *cobra.Command) error { //nolint:funlen
+	zshInitialization := `#compdef lokoctl
```
invidian, line 89: I think we could move those fragments to constants.

cli/cmd/completion.go
```diff
@@ -0,0 +1,251 @@
...
+	-e "s/\\\$(type${RWORD}/\$(__lokoctl_type/g" \
+	-e 's/aliashash\["\(.\{1,\}\)"\]/aliashash[\1]/g' \
+	-e 's/FUNCNAME/funcstack/g' \
+	<<'BASH_COMPLETION_EOF'
+`
+
+	_, err := out.Write([]byte(zshInitialization))
+	if err != nil {
+		return err
+	}
```
invidian, line 223:
```suggestion
	if _, err := out.Write([]byte(zshInitialization)); err != nil {
	  fmt.Errorf("writing zsh initialization: %w", err)
	}
```

cli/cmd/completion.go
```diff
@@ -0,0 +1,251 @@
...
+
+	err = runCompletionBash(out, cmd)
+	if err != nil {
+		return err
+	}
```
invidian, line 228:
```suggestion
	if err := runCompletionBash(out, cmd); err != nil {
		return fmt.Errorf("running bash completion: %w, err)
	}
```

cli/cmd/completion.go
```diff
@@ -0,0 +1,251 @@
...
+	zshTail := `
+BASH_COMPLETION_EOF
+}
+__lokoctl_bash_source <(__lokoctl_convert_bash_to_zsh)
+`
+
+	_, err = out.Write([]byte(zshTail))
+	if err != nil {
+		return err
+	}
```
invidian, line 239:
```suggestion
	if _, err = out.Write([]byte(zshTail)); err != nil {
		return fmt.Errorf("writing zsh tail: %w", err)
	}
```

cli/cmd/completion.go
```diff
@@ -0,0 +1,251 @@
...
+
+	"To load completions for every new session, execute once: \n" +
+	"`$ lokoctl completion zsh > \"${fpath[1]}/_lokoctl\" && exec $SHELL`"
+
+func newCompletionCmd(out io.Writer) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:               "completion",
+		Short:             "Generate autocompletion script for the specified shell",
+		Long:              completionDesc,
+		Args:              require.NoArgs,
```
invidian, line 51: Do we really need to depend on Helm code for that?

invidian (changes requested):
Some nits, otherwise LGTM :+1:

cli/cmd/completion.go
knrt10: But we need this writer. It is passed onto `runCompletionZsh` and `runCompletionBash`

invidian: Ah, sorry, I missed that :+1:

docs/cli/lokoctl.md
```diff
@@ -21,6 +21,7 @@ Manage Lokomotive clusters
 ### SEE ALSO
 
 * [lokoctl cluster](lokoctl_cluster.md)	 - Manage a cluster
+* [lokoctl completion](lokoctl_completion.md)	 - Generate autocompletion script for the specified shell
```
invidian, line 24: I'd actually put this all in a single commit, so if one runs the build per commit, it stays atomic.

cli/cmd/completion.go
```diff
@@ -0,0 +1,294 @@
...
+
+	"github.com/spf13/cobra"
+)
+
+const completionDesc = `
+Generate autocompletion script for lokoctl for the specified shell (bash or zsh).
+`
+
+const completionExample = `
+		# Installing bash completion on macOS using homebrew
```
invidian, line 30: Why do we need indentation here and extra newline at the beginning?

cli/cmd/completion.go
```diff
@@ -0,0 +1,294 @@
...
+	return cmd
+}
+
+func runCompletionBash(out io.Writer, cmd *cobra.Command) error {
+	return cmd.Root().GenBashCompletion(out)
+}
+
+func runCompletionZsh(out io.Writer, cmd *cobra.Command) error {
+	_, err := out.Write([]byte(zshInitialization))
+	if err != nil {
```
invidian, line 257: Please consolidate lines which only assigns `err` with `if` statement lines.
```suggestion
	if _, err := out.Write([]byte(zshInitialization)); err != nil {
```

invidian (changes requested):

cli/cmd/completion.go
```diff
@@ -0,0 +1,290 @@
...
+	"os"
+
+	"github.com/spf13/cobra"
+)
+
+const completionDesc = `
+Generate autocompletion script for lokoctl for the specified shell (bash or zsh).
+`
+
+const completionExample = `# Installing bash completion on macOS using homebrew
```
invidian, line 29: Looking at current help message format:
```
Usage:
  lokoctl completion [command]

Examples:
# Installing bash completion on macOS using homebrew
## If running Bash 3.2 included with macOS
  brew install bash-completion
## or, if running Bash 4.1+
  brew install bash-completion@2
## You need to add the completion to your completion directory
  lokoctl completion bash > /usr/local/etc/bash_completion.d/lokoctl && exec $SHELL

# Installing bash completion on Linux
## If bash-completion is not installed on Linux, please install the 'bash-completion' package
## via your distribution's package manager.
## You need to add the completion to your completion directory
  lokoctl completion bash > /etc/bash_completion.d/lokoctl && exec $SHELL

# Load the lokoctl completion code for bash into the current shell
  source <(lokoctl completion bash)
# Load the lokoctl completion code for zsh into the current shell
  source <(lokoctl completion zsh)

# Set the lokoctl completion code for zsh to autoload on startup
  lokoctl completion zsh > "${fpath[1]}/_lokoctl" && exec $SHELL

Available Commands:
  bash        Generate autocompletion script for bash
  zsh         Generate autocompletion script for zsh

Flags:
  -h, --help   help for completion
```

It seems example text could be indented with 2 spaces and should not use `#` characters. Can you re-format it @knrt10 to some better format?

invidian (changes requested):

cli/cmd/completion.go
johananl: +1. Also, we should pay more attention to the following in our codebase in general and especially in text the users see:

- Punctuation.
  - Some lines in the examples text end in periods whereas others don't.
- Capitalization.
  - We use both `Bash` and `bash`. We should use `Bash`, and anyway we should be consistent.
  - Some lines begin with a capital letter whereas others don't.
- Consistent tone and style.
  - We should be consistent about using the imperative tone and avoiding "please" as in "please do this and that".
  - In one instance we use `To load completions in your current shell session:` and then immediately following we use `To load completions for every new session, execute once:`.
- Correct English.
  - I'm fairly certain the following sentence has some words missing in it (probably `that's` before `included`): `If running Bash 3.2 included with macOS`.

knrt10:
@johananl can you please review again

cli/cmd/completion.go
```diff
@@ -0,0 +1,281 @@
...
+import (
+	"fmt"
+	"io"
+	"os"
+
+	"github.com/spf13/cobra"
+)
+
+const completionDesc = `
+Generate autocompletion script for lokoctl for the specified shell (bash or zsh).
```
johananl, line 26:
```suggestion
Generate autocompletion code for lokoctl for the specified shell (bash or zsh).
```

I find "code" better than "script" in this context. In any case, if we use "script" then it should be "Generate **an** autocompletion script..." since "script" is countable (unlike "code" in this context).

cli/cmd/completion.go
```diff
@@ -0,0 +1,281 @@
...
+`
+
+const completionExample = `  # Load the lokoctl completion code for bash into the current shell.
+  source <(lokoctl completion bash)
+
+  # Load the lokoctl completion code for zsh into the current shell.
+  source <(lokoctl completion zsh)
+
+  # Generate Bash completion file and load it for every shell.
```
johananl, line 35:
```suggestion
  # Generate a Bash completion file and load it for every shell.
```

cli/cmd/completion.go
```diff
@@ -0,0 +1,281 @@
...
+  lokoctl completion bash > ~/.bash_lokoctl_completion
+  echo "source ~/.bash_lokoctl_completion" >> ~/.bashrc && source ~/.bashrc
+
+  # Set the lokoctl completion code for zsh to autoload on startup.
+  lokoctl completion zsh > "${fpath[1]}/_lokoctl" && exec $SHELL`
+
+const bashCompDesc = `  ### If running Bash 3.2 that is included with macOS, install bash completion using homebrew.
```
johananl, line 42: Why use `###`? Is this some sort of convention?

johananl, line 42:
```suggestion
const bashCompDesc = `  ### If running Bash 3.2 that is included with macOS, install bash completion using Homebrew.
```

cli/cmd/completion.go
johananl, line 26: We should consistently capitalize "Bash" as `Bash`.

cli/cmd/completion.go
```diff
@@ -0,0 +1,281 @@
...
+
+  ### Load the lokoctl completion code for bash into the current shell.
+  source <(lokoctl completion bash)
+
+  ### Generate Bash completion file and load it for every shell.
+  lokoctl completion bash > ~/.bash_lokoctl_completion
+  echo "source ~/.bash_lokoctl_completion" >> ~/.bashrc && source ~/.bashrc
+`
+
+const zshCompDesc = `  Generate the autocompletion script for lokoctl for the zsh shell.
```
johananl, line 56:
```suggestion
const zshCompDesc = `  Generate an autocompletion script for lokoctl for the zsh shell.
```

johananl, line 56: We don't have an equivalent line in the long description for `lokoctl completion bash`.

cli/cmd/completion.go
```diff
@@ -0,0 +1,281 @@
...
+
+  ### Load the lokoctl completion code for zsh into the current shell.
```
johananl, line 58: We seem to be using "autocompletion" and "completion" randomly. We should settle on one and use it consistently. Same for "code" vs "script" (I prefer "code").

cli/cmd/completion.go
```diff
@@ -0,0 +1,281 @@
...
+	alias _complete=_bash_comp
+	emulate -L sh
+	setopt kshglob noshglob braceexpand
+	source "$@"
+}
+__lokoctl_type() {
+	# -t is not supported by zsh
+	if [ "$1" == "-t" ]; then
+		shift
+		# fake Bash 4 to disable "complete -o nospace". Instead
```
johananl, line 78: Does this text come from upstream? If so, please ignore this comment as well as similar comments in this file.

- The comment should probably begin with an upper case letter.
- The last sentence doesn't have a period.
- I find the phrasing in this comment confusing.

cli/cmd/completion.go
```diff
@@ -0,0 +1,281 @@
...
+		shift
+	done
+	if [[ "$1" == -- ]]; then
+		shift
+	fi
+	for w in "${completions[@]}"; do
+		if [[ "${w}" = "$1"* ]]; then
+			# Use printf instead of echo because it is possible that
+			# the value to print is -n, which would be interpreted
+			# as a flag to echo
```
johananl, line 104: Missing period.

cli/cmd/completion.go
```diff
@@ -0,0 +1,281 @@
...
+			printf "%s\n" "${w}"
+		fi
+	done
+}
+__lokoctl_compopt() {
+	true # don't do anything. Not supported by bashcompinit in zsh
```
johananl, line 110: - Caps.
- Period.

cli/cmd/completion.go
```diff
@@ -0,0 +1,281 @@
...
+}
+__lokoctl_ltrim_colon_completions()
+{
+	if [[ "$1" == *:* && "$COMP_WORDBREAKS" == *:* ]]; then
+		# Remove colon-word prefix from COMPREPLY items
```
johananl, line 115: Period.

cli/cmd/completion.go
```diff
@@ -0,0 +1,281 @@
...
+	cur="${COMP_WORDS[COMP_CWORD]}"
+	prev="${COMP_WORDS[${COMP_CWORD}-1]}"
+	words=("${COMP_WORDS[@]}")
+	cword=("${COMP_CWORD[@]}")
+}
+__lokoctl_filedir() {
+	local RET OLD_IFS w qw
+	__debug "_filedir $@ cur=$cur"
+	if [[ "$1" = \~* ]]; then
+		# somehow does not work. Maybe, zsh does not call this at all
```
johananl, line 133: - Caps.
- Period.

cli/cmd/completion.go
```diff
@@ -0,0 +1,281 @@
...
+__lokoctl_quote() {
+	if [[ $1 == \'* || $1 == \"* ]]; then
+		# Leave out first character
+		printf %q "${1:1}"
+	else
+		printf %q "$1"
+	fi
+}
+autoload -U +X bashcompinit && bashcompinit
+# use word boundary patterns for BSD or GNU sed
```
johananl, line 170: - Caps.
- Period.

cli/cmd/completion.go
```diff
@@ -0,0 +1,281 @@
...
+				COMPREPLY+=("${qw}/")
+			else
+				COMPREPLY+=("${qw}")
+			fi
+		fi
+	done
+}
+__lokoctl_quote() {
+	if [[ $1 == \'* || $1 == \"* ]]; then
+		# Leave out first character
```
johananl, line 163: Period.

johananl (commented):
Thanks @knrt10.

I'm still seeing small inconsistencies. I've highlighted some of them inline. We also seem to be using multiple variations of "completion/autocompletion" and "script/code". Let's settle on one phrasing :wink:

cli/cmd/completion.go
knrt10: For docs, otherwise, they look wierd

cli/cmd/completion.go
knrt10: Yes, it's from upstream.

knrt10:
Updated

cli/cmd/completion.go
```diff
@@ -0,0 +1,283 @@
...
+  # Generate a Bash completion file and load it for every shell.
+  lokoctl completion bash > ~/.bash_lokoctl_completion
+  echo "source ~/.bash_lokoctl_completion" >> ~/.bashrc && source ~/.bashrc
+
+  # Set the lokoctl completion code for zsh to autoload on startup.
+  lokoctl completion zsh > "${fpath[1]}/_lokoctl" && exec $SHELL`
+
+const bashCompDesc = `  Generate the completion code for lokoctl for the Bash shell.
+
+  If running Bash 3.2 that is included with macOS, install Bash completion using Homebrew.
```
invidian, line 44: I'd be consistent with those comments.
```suggestion
  ### If running Bash 3.2 that is included with macOS, install Bash completion using Homebrew.
```

invidian (changes requested):
Just one nit, otherwise looks fine to me.

cli/cmd/completion.go
```diff
@@ -0,0 +1,283 @@
...
+  source <(lokoctl completion zsh)
+
+  # Generate a Bash completion file and load it for every shell.
+  lokoctl completion bash > ~/.bash_lokoctl_completion
+  echo "source ~/.bash_lokoctl_completion" >> ~/.bashrc && source ~/.bashrc
+
+  # Set the lokoctl completion code for zsh to autoload on startup.
+  lokoctl completion zsh > "${fpath[1]}/_lokoctl" && exec $SHELL`
+
+const bashCompDesc = `  Generate the completion code for lokoctl for the Bash shell.
```
iaguis, line 42: This looks pretty ugly in the documentation.

Doesn't it make more sense to have this as `Example` and add a small `Description` like we do for the `completion` command? The instructions below feel more like examples and it would make it more consistent too. We can also remove the triple `#`.

cli/cmd/completion.go
```diff
@@ -0,0 +1,283 @@
...
+
+  ### Load the lokoctl completion code for Bash into the current shell.
+  source <(lokoctl completion bash)
+
+  ### Generate a Bash completion file and load it for every shell.
+  lokoctl completion bash > ~/.bash_lokoctl_completion
+  echo "source ~/.bash_lokoctl_completion" >> ~/.bashrc && source ~/.bashrc
+`
+
+const zshCompDesc = `  Generate the completion code for lokoctl for the zsh shell.
```
iaguis, line 58: See my previous comment.

cli/cmd/completion.go
```diff
@@ -0,0 +1,283 @@
...
+}
+autoload -U +X bashcompinit && bashcompinit
+# Use word boundary patterns for BSD or GNU sed.
+LWORD='[[:<:]]'
+RWORD='[[:>:]]'
+if sed --help 2>&1 | grep -q 'GNU\|BusyBox'; then
+	LWORD='\<'
+	RWORD='\>'
+fi
+__lokoctl_convert_bash_to_zsh() {
```
iaguis, line 179: Can we add a todo somewhere to get rid of this when Cobra is released?

iaguis (changes requested):
Some small comments.

iaguis (approved):
lgtm

invidian (approved):
LGTM

knrt10:
tested for both zsh and bash. Merging now

PR #885, opened 2020-08-31
components/metallb: Update to v0.1.0-789-g85b7a46a

johananl:
This PR updates the `metallb` component to a more recent "snapshot" of the [peer autodiscovery feature](https://github.com/metallb/metallb/pull/593/). A main purpose of the update is to stabilize the BGP sessions we use on Packet.

As part of this update I've copy-pasted the manifests from the upstream ones as much as possible to make future updates easier. In some cases this resulted in fields being **reordered**.

On preliminary testing things look good to me, however I've opened the PR to get reviews ASAP since this update is kind of urgent. We should examine the changes carefully, and ideally at least one more person should manually verify things look sane before we merge this.


invidian (dismissed):
LGTM. Do we have an e2e test to cover ingress requests on Packet which would test that automatically?

I don't have an experience with setting up BGP on Packet, so testing may take me a bit longer.

johananl:
> LGTM. Do we have an e2e test to cover ingress requests on Packet which would test that automatically?
> 
> I don't have an experience with setting up BGP on Packet, so testing may take me a bit longer.

Doesn't look like we have that. I can look into adding that.

iaguis:
I've tested this manually and it worked for me :+1:

invidian (approved):

johananl:
Thanks @invidian. I assume you were able to test and things went fine?

johananl:
I'm taking a quick look at adding an integration test to verify ingress on Packet.

invidian:
>Thanks @invidian. I assume you were able to test and things went fine?

Yes, sorry for the delay, it works nicely :+1:

PR #886, opened 2020-09-01
Introduce platform.PostApplyHook interface and implement it for AKS cluster

### LINKED ISSUE #855: AKS: prometheus-operator gets installed before default storage class object is created
This causes the CI to fail, as `prometheus-operator` component never converges.

This is related to #559, however, to solve this one, we should perhaps make sure, that before AKS installation converges, default storage class becomes available. We could for example install the storage class on AKS explicitly via chart, to avoid waiting for the AKS to converge.

invidian:
See commit messages for the details.

Closes #855

### DIFF
--- a/pkg/platform/aks/aks.go
+++ b/pkg/platform/aks/aks.go
@@ -17,15 +17,21 @@
 package aks
 
 import (
+	"context"
 	"fmt"
 	"os"
 	"path/filepath"
 	"text/template"
+	"time"
 
 	"github.com/hashicorp/hcl/v2"
 	"github.com/hashicorp/hcl/v2/gohcl"
 	"github.com/mitchellh/go-homedir"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	v1 "k8s.io/client-go/kubernetes/typed/storage/v1"
 
+	"github.com/kinvolk/lokomotive/pkg/k8sutil"
 	"github.com/kinvolk/lokomotive/pkg/platform"
 	"github.com/kinvolk/lokomotive/pkg/terraform"
 )
@@ -307,6 +313,48 @@ func (c *config) Initialize(ex *terraform.Executor) error {
 	terraformRootDir := terraform.GetTerraformRootDir(assetDir)
 
 	return createTerraformConfigFile(c, terraformRootDir)
+}
+
+const (
+	retryInterval = 5 * time.Second
+	timeout       = 10 * time.Minute
+)
+
+// PostApplyHook implements platform.PlatformWithPostApplyHook interface and defines hooks
+// which should be executed after AKS cluster is created.
+func (c *config) PostApplyHook(kubeconfig []byte) error {
+	client, err := k8sutil.NewClientset(kubeconfig)
+	if err != nil {
+		return fmt.Errorf("creating clientset from kubeconfig: %w", err)
+	}
+
+	return waitForDefaultStorageClass(client.StorageV1().StorageClasses())
+}
+
+// waitForDefaultStorageClass waits until the default storage class appears on a given cluster.
+// If it doesn't appear within a defined time range, an error is returned.
+func waitForDefaultStorageClass(sci v1.StorageClassInterface) error {
+	// AKS still uses annotation with .beta.
+	defaultStorageClassAnnotation := "storageclass.beta.kubernetes.io/is-default-class"
+
+	if err := wait.PollImmediate(retryInterval, timeout, func() (done bool, err error) {
+		scs, err := sci.List(context.TODO(), metav1.ListOptions{})
+		if err != nil {
+			return false, fmt.Errorf("getting storage classes: %w", err)
+		}
+
+		for _, sc := range scs.Items {
+			if v, ok := sc.ObjectMeta.Annotations[defaultStorageClassAnnotation]; ok && v == "true" {
+				return true, nil
+			}
+		}
+
+		return false, nil
+	}); err != nil {
+		return fmt.Errorf("waiting for the default storage class to be configured: %w", err)
+	}
+
+	return nil
 }
 
 // createTerraformConfigFiles create Terraform config files in given directory.
--- a/pkg/platform/platform.go
+++ b/pkg/platform/platform.go
@@ -74,6 +74,13 @@ type Platform interface {
 	Meta() Meta
 }
 
+// PlatformWithPostApplyHook runs code after Terraform finishes applying. This allows
+// running sanity checks on the newly created cluster. Implementing this
+// interface is optional for platforms.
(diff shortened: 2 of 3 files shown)


pkg/platform/platform.go
```diff
@@ -74,6 +74,13 @@ type Platform interface {
 	Meta() Meta
```
knrt10, line 74: Some nits on the commit message. You used `interface` twice at the end of line 3. Also comma on line 3. I think it should be `period`.

pkg/platform/aks/aks.go
```diff
@@ -309,6 +315,48 @@ func (c *config) Initialize(ex *terraform.Executor) error {
...
+// waitForDefaultStorageClass wait until default storage class appears on a given cluster.
+// If it doesn't within defined time range, error is returned.
+func waitForDefaultStorageClass(sci v1.StorageClassInterface) error {
+	// AKS still uses annotation with .beta.
+	defaultStorageClassAnnotation := "storageclass.beta.kubernetes.io/is-default-class"
+
+	if err := wait.PollImmediate(retryInterval, timeout, func() (done bool, err error) {
+		scs, err := sci.List(context.TODO(), metav1.ListOptions{})
+		if err != nil {
+			return false, fmt.Errorf("getting storage classes %w", err)
```
knrt10, line 343:
```suggestion
			return false, fmt.Errorf("getting storage classes: %w", err)
```

knrt10 (commented):
Just a NIT, rest LGTM.

pkg/platform/aks/aks.go
invidian: Oops, fixed thanks :)

pkg/platform/platform.go
invidian: I _think_ I fixed it :smile: Please have a look again.

knrt10 (dismissed):
+1

pkg/platform/aks/aks.go
```diff
@@ -309,6 +315,48 @@ func (c *config) Initialize(ex *terraform.Executor) error {
...
+	client, err := k8sutil.NewClientset(kubeconfig)
+	if err != nil {
+		return fmt.Errorf("creating clientset from kubeconfig: %w", err)
+	}
+
+	return waitForDefaultStorageClass(client.StorageV1().StorageClasses())
+}
+
+// waitForDefaultStorageClass wait until default storage class appears on a given cluster.
+// If it doesn't within defined time range, error is returned.
```
iaguis, line 335:
```suggestion
// waitForDefaultStorageClass waits until the default storage class appears on a given cluster.
// If it doesn't appear within a defined time range, an error is returned.
```

pkg/platform/platform.go
```diff
@@ -74,6 +74,13 @@ type Platform interface {
...
 }
 
+// PostApplyHook is an interface, which platforms may optionally implement. If they do so,
+// after Terraform finishes applying, kubeconfig content will be passed to this function and
+// executed. This allows running sanity checks on the newly created cluster.
```
iaguis, line 79: I'd simplify this a bit.

```suggestion
// PostApplyHook runs code after Terraform finishes applying. This allows
// running sanity checks on the newly created cluster. Implementing this
// interface is optional for platforms.
```

iaguis (commented):
Looks OK but I'd like to have @johananl's review before merging.

knrt10 (approved):

johananl:
I'm fine with this implementation for now if it solves the problem for us. Once we refactor the platform interface I think it could make more sense to have a method called `PostApplyHook()` (or even `PostApplyHooks()`?) **on the platform interface** which optionally returns the hook(s) to be executed. Doing so would make things more explicit compared to implementing another interface.

iaguis (approved):

invidian:
>I'm fine with this implementation for now if it solves the problem for us. Once we refactor the platform interface I think it could make more sense to have a method called PostApplyHook() (or even PostApplyHooks()?) on the platform interface which optionally returns the hook(s) to be executed. Doing so would make things more explicit compared to implementing another interface.

Since the hook is optional, I think it's nice to have it separate, as it allows us to avoid adding boilerplate code to all platforms. This is not something every platform MUST implement, like e.g. generating Terraform code or Terraform plan (this could actually be optional too).

johananl:
> > I'm fine with this implementation for now if it solves the problem for us. Once we refactor the platform interface I think it could make more sense to have a method called PostApplyHook() (or even PostApplyHooks()?) on the platform interface which optionally returns the hook(s) to be executed. Doing so would make things more explicit compared to implementing another interface.
> 
> Since the hook is optional, I think it's nice to have it separate, as it allows us to avoid adding boilerplate code to all platforms. This is not something every platform MUST implement, like e.g. generating Terraform code or Terraform plan (this could actually be optional too).

The price you're paying is clarity: having an interface which can be optionally implemented leaves things implicit and more "magical" as far as the caller is concerned. It also doesn't make sense to me conceptually at the abstraction level: A platform is a type of a hook? :thinking: This is very different from saying "AWS is a type of platform".

I have no problem returning `nil` or an empty slice when no hooks are necessary on all platforms. To me this is much better than bending the abstraction to avoid "duplicating" a single line per platform.

invidian:
>I have no problem returning nil or an empty slice when no hooks are necessary on all platforms. To me this is much better than bending the abstraction to avoid "duplicating" a single line per platform.

Well, that's exectly the boilerplate I'm talking about. And it's not single line, but at least 4 :smile: 

>The price you're paying is clarity: Having an interface which can be optionally implemented leaves things implicit and more "magical" as far as the caller is concerned. It also doesn't make sense to me conceptually at the abstraction level: "A platform is a type of hook?" thinking This is very different from saying "AWS is a type of platform". Go interfaces are built for certain purposes and I don't think this use case is one of them. I think a hint which supports this is that you had to type-assert the interface rather than just executing a generic method.

Yeah, I agree that abstraction wise this is not great. (Actually, initially my interface was named `PlatformPostApplyHook`, but then linter complains, that you do `platform.PlatformPostApplyHook`.

And what do you think about:
```go
type PlatformWithPostApplyHook interface {
  Platform
  PostApplyHook(kubeconfig []byte) error
}
```

I think it provides much better context then, as it extends the `Platform` interface and optionally ships this extra method. Name is something we should discuss of course :)

iaguis:
I think this discussion doesn't belong to this PR :)

invidian:
>I think this discussion doesn't belong to this PR :)

Well, I consider now changing this newly introduced interface to the version from https://github.com/kinvolk/lokomotive/pull/886#issuecomment-684885334 :smile:

johananl:
For the sake of moving this forward @invidian, please do whatever makes sense to you which solves the problem *for now*, and I'll do my best to integrate whatever you come up with into the new platform interface in a way which makes sense. We can continue the discussion there.

knrt10 (approved):

iaguis (approved):

PR #909, opened 2020-09-04
cli/cmd: fix re-installing controlplane components in custom namespaces

invidian:
Since 474abab61cfc7bf33aeb91ea8682f5f7228d986d, we support also
installing controlplane components in namespaces other than kube-system.

On fresh installation, those namespaces are being created by bootkube.

However, in scenario when user upgrades from the previous release, we
also need to ensure that custom namespace exists, so this commit does
that by instructing Helm to use CreateNamespace feature.


surajssd (approved):
Upgrade worked fine.

PR #911, opened 2020-09-07
Add encryption to in-cluster pod traffic

knrt10:
Calico v3.15+ support wireguard and if enabled it encrypts in-cluster
pod traffic across nodes.

### Release Notes

- delete `default FelixConfiguration` while upgrading from 0.4.1
- `network_mtu` changed from CNI network MTU to Physical network MTU.
- `bpfLogLevel` will be briefly set from `Info` to `Off`.
- Following ports will be opened on all hosts for the upgrade time between `calico` chart upgrade and `calico-host-protection` chart:
  
```
  - 22/tcp
  - 68/udp
  - 179/tcp
  - 2379/tcp
  - 2380/tcp
  - 5473/tcp
  - 6443/tcp
  - 6666/tcp
  - 6667/tcp
```

### DIFF
--- a/assets/charts/control-plane/calico-host-protection/templates/host-protection.yaml
+++ b/assets/charts/control-plane/calico-host-protection/templates/host-protection.yaml
@@ -1,15 +1,3 @@
-# General
-# =======
----
-# Disable default failsafe inbound rules
-apiVersion: crd.projectcalico.org/v1
-kind: FelixConfiguration
-metadata:
-  name: default
-spec:
-  failsafeInboundHostPorts: []
-  bpfLogLevel: Info
-
 # Firewall policy
 # ===============
 ---
--- a/assets/charts/control-plane/calico/templates/config.yaml
+++ b/assets/charts/control-plane/calico/templates/config.yaml
@@ -13,7 +13,7 @@ data:
   # - Otherwise, if VXLAN or BPF mode is enabled, set to your network MTU - 50
   # - Otherwise, if IPIP is enabled, set to your network MTU - 20
   # - Otherwise, if not using any encapsulation, set to your network MTU.
-  veth_mtu: "{{ .Values.calico.networkMTU }}"
+  veth_mtu: "{{ sub .Values.calico.networkMTU 20 }}" # IP in IP overhead.
   # The CNI network configuration to install on each node. The special
   # values in this config will be automatically populated.
   cni_network_config: |-
--- a/assets/charts/control-plane/calico/templates/config.yaml
+++ b/assets/charts/control-plane/calico/templates/config.yaml
@@ -13,7 +13,11 @@ data:
   # - Otherwise, if VXLAN or BPF mode is enabled, set to your network MTU - 50
   # - Otherwise, if IPIP is enabled, set to your network MTU - 20
   # - Otherwise, if not using any encapsulation, set to your network MTU.
+  {{- if .Values.calico.encryptPodTraffic }}
+  veth_mtu: "{{ sub .Values.calico.networkMTU 60 }}" # Wireguard overhead.
+  {{- else }}
   veth_mtu: "{{ sub .Values.calico.networkMTU 20 }}" # IP in IP overhead.
+  {{- end}}
   # The CNI network configuration to install on each node. The special
   # values in this config will be automatically populated.
   cni_network_config: |-
--- a/assets/charts/control-plane/calico/templates/felixconfig.yaml
+++ b/assets/charts/control-plane/calico/templates/felixconfig.yaml
@@ -7,3 +7,4 @@ spec:
   failsafeInboundHostPorts: {{ .Values.calico.failsafeInboundHostPorts }}
   {{- end }}
   bpfLogLevel: Info
+  wireguardEnabled: {{ .Values.calico.encryptPodTraffic }}
--- a/assets/charts/control-plane/calico/templates/felixconfig.yaml
+++ b/assets/charts/control-plane/calico/templates/felixconfig.yaml
@@ -1,0 +1,9 @@
+apiVersion: crd.projectcalico.org/v1
+kind: FelixConfiguration
+metadata:
+  name: default
+spec:
+  {{- if hasKey .Values.calico "failsafeInboundHostPorts"  }}
+  failsafeInboundHostPorts: {{ .Values.calico.failsafeInboundHostPorts }}
+  {{- end }}
+  bpfLogLevel: Info
--- a/assets/charts/control-plane/calico/values.yaml
+++ b/assets/charts/control-plane/calico/values.yaml
@@ -13,3 +13,5 @@ calico:
   podCIDR: 10.2.0.0/16
   networkEncapsulation: "ipipMode: Always"
   blockedMetadataCIDRs: []
+  # Lokomotive specific change.
+  # failsafeInboundHostPorts:
--- a/assets/charts/control-plane/calico/values.yaml
+++ b/assets/charts/control-plane/calico/values.yaml
@@ -15,3 +15,4 @@ calico:
   blockedMetadataCIDRs: []
   # Lokomotive specific change.
   # failsafeInboundHostPorts:
+  encryptPodTraffic: false
--- a/assets/terraform-modules/aws/flatcar-linux/kubernetes/variables.tf
+++ b/assets/terraform-modules/aws/flatcar-linux/kubernetes/variables.tf
@@ -184,3 +184,9 @@ variable "kube_apiserver_extra_flags" {
   type        = list(string)
   default     = []
 }
+
+variable "encrypt_pod_traffic" {
+  description = "Enable in-cluster pod traffic encryption."
+  type        = bool
+  default     = false
+}
--- a/assets/terraform-modules/aws/flatcar-linux/kubernetes/variables.tf
+++ b/assets/terraform-modules/aws/flatcar-linux/kubernetes/variables.tf
@@ -111,9 +111,8 @@ variable "asset_dir" {
 }
 
 variable "network_mtu" {
-  description = "CNI interface MTU. Use 8981 if using instances types with Jumbo frames."
+  description = "Physical Network MTU. Use 9001 if using instances types with Jumbo frames."
   type        = number
-  default     = 1480
 }
 
 variable "host_cidr" {
--- a/assets/terraform-modules/bare-metal/flatcar-linux/kubernetes/variables.tf
+++ b/assets/terraform-modules/bare-metal/flatcar-linux/kubernetes/variables.tf
@@ -83,9 +83,8 @@ variable "asset_dir" {
 }
 
 variable "network_mtu" {
-  description = "CNI interface MTU"
+  description = "Physical Network MTU."
   type        = number
-  default     = 1480
 }
 
 variable "network_ip_autodetection_method" {
--- a/assets/terraform-modules/bare-metal/flatcar-linux/kubernetes/variables.tf
+++ b/assets/terraform-modules/bare-metal/flatcar-linux/kubernetes/variables.tf
@@ -183,3 +183,9 @@ variable "kube_apiserver_extra_flags" {
   type        = list(string)
   default     = []
 }
+
+variable "encrypt_pod_traffic" {
+  description = "Enable in-cluster pod traffic encryption."
+  type        = bool
+  default     = false
+}
--- a/assets/terraform-modules/bootkube/conditional.tf
+++ b/assets/terraform-modules/bootkube/conditional.tf
@@ -22,6 +22,7 @@ resource "local_file" "calico" {
         port     = protoport.port
       }
     ] : null
+    encrypt_pod_traffic = var.encrypt_pod_traffic
   })
   filename = "${var.asset_dir}/charts/kube-system/calico.yaml"
 }
--- a/assets/terraform-modules/bootkube/conditional.tf
+++ b/assets/terraform-modules/bootkube/conditional.tf
@@ -15,6 +15,13 @@ resource "local_file" "calico" {
     pod_cidr                        = var.pod_cidr
     enable_reporting                = var.enable_reporting
     blocked_metadata_cidrs          = var.blocked_metadata_cidrs
+    failsafe_inbound_host_ports = var.failsafe_inbound_host_ports != null ? [
+      for protoport in var.failsafe_inbound_host_ports :
+      {
+        protocol = protoport.protocol
+        port     = protoport.port
+      }
+    ] : null
   })
   filename = "${var.asset_dir}/charts/kube-system/calico.yaml"
 }
--- a/assets/terraform-modules/bootkube/resources/charts/calico.yaml
+++ b/assets/terraform-modules/bootkube/resources/charts/calico.yaml
@@ -16,3 +16,13 @@ calico:
   - ${cidr}
   %{~ endfor ~}
   %{~ endif ~}
+  %{~ if failsafe_inbound_host_ports != null && failsafe_inbound_host_ports != [] ~}
+  failsafeInboundHostPorts:
+  %{~ for protoport in failsafe_inbound_host_ports ~}
+  - protocol: ${protoport.protocol}
+    port: ${protoport.port}
+  %{~ endfor ~}
+  %{~ endif ~}
+  %{~ if failsafe_inbound_host_ports == [] ~}
+  failsafeInboundHostPorts: []
+  %{~ endif ~}
--- a/assets/terraform-modules/bootkube/resources/charts/calico.yaml
+++ b/assets/terraform-modules/bootkube/resources/charts/calico.yaml
@@ -26,3 +26,4 @@ calico:
   %{~ if failsafe_inbound_host_ports == [] ~}
   failsafeInboundHostPorts: []
   %{~ endif ~}
+  encryptPodTraffic: ${encrypt_pod_traffic}
--- a/assets/terraform-modules/bootkube/variables.tf
+++ b/assets/terraform-modules/bootkube/variables.tf
@@ -179,3 +179,9 @@ variable "failsafe_inbound_host_ports" {
   type        = list(any)
   default     = null
 }
+
+variable "encrypt_pod_traffic" {
+  description = "Enable in-cluster pod traffic encryption."
+  type        = bool
+  default     = false
+}
--- a/assets/terraform-modules/bootkube/variables.tf
+++ b/assets/terraform-modules/bootkube/variables.tf
@@ -173,3 +173,9 @@ variable "enable_tls_bootstrap" {
   description = "Enable TLS Bootstrap for Kubelet."
   type        = bool
 }
+
+variable "failsafe_inbound_host_ports" {
+  description = "UDP/TCP/SCTP protocol/port pairs to allow incoming traffic on regardless of the security policy."
+  type        = list(any)
+  default     = null
+}
--- a/assets/terraform-modules/packet/flatcar-linux/kubernetes/bootkube.tf
+++ b/assets/terraform-modules/packet/flatcar-linux/kubernetes/bootkube.tf
@@ -47,4 +47,5 @@ module "bootkube" {
 
   # We install calico-host-protection chart on Packet which ships GNPs, so we can disable failsafe ports in Calico.
   failsafe_inbound_host_ports = []
+  encrypt_pod_traffic         = var.encrypt_pod_traffic
 }
--- a/assets/terraform-modules/packet/flatcar-linux/kubernetes/bootkube.tf
+++ b/assets/terraform-modules/packet/flatcar-linux/kubernetes/bootkube.tf
@@ -44,4 +44,7 @@ module "bootkube" {
 
   bootstrap_tokens     = var.enable_tls_bootstrap ? concat([local.controller_bootstrap_token], var.worker_bootstrap_tokens) : []
   enable_tls_bootstrap = var.enable_tls_bootstrap
+
+  # We install calico-host-protection chart on Packet which ships GNPs, so we can disable failsafe ports in Calico.
+  failsafe_inbound_host_ports = []
 }
--- a/docs/configuration-reference/platforms/aws.md
+++ b/docs/configuration-reference/platforms/aws.md
@@ -107,6 +107,8 @@ cluster "aws" {
 
   enable_tls_bootstrap = true
 
+  encrypt_pod_traffic = true
+
   disk_size = var.disk_size
 
   disk_type = var.disk_type
@@ -219,6 +221,7 @@ worker_pool "my-worker-pool" {
 | `region`                      | AWS region to use for deploying the cluster.                                                                                                                                               | "eu-central-1"  |    string    |  false   |
 | `enable_aggregation`          | Enable the Kubernetes Aggregation Layer.                                                                                                                                                   |      true       |     bool     |  false   |
 | `enable_tls_bootstrap`        | Enable TLS bootstraping for Kubelet.                                                                                                                                                       |      true       |     bool     |  false   |
+| `encrypt_pod_traffic`         | Enable in-cluster pod traffic encryption. If true `network_mtu` is reduced by 60 to make room for the encryption header.                                                                   |      false      |     bool     |  false   |
 | `disk_size`                   | Size of the EBS volume in GB.                                                                                                                                                              |       40        |    number    |  false   |
 | `disk_type`                   | Type of the EBS volume (e.g. standard, gp2, io1).                                                                                                                                          |      "gp2"      |    string    |  false   |
 | `disk_iops`                   | IOPS of the EBS volume (e.g 100).                                                                                                                                                          |        0        |    number    |  false   |
--- a/docs/configuration-reference/platforms/aws.md
+++ b/docs/configuration-reference/platforms/aws.md
@@ -113,7 +113,7 @@ cluster "aws" {
 
   disk_iops = var.disk_iops
 
-  network_mtu = 1480
+  network_mtu = 1500
 
   host_cidr = ""
 
@@ -222,7 +222,7 @@ worker_pool "my-worker-pool" {
 | `disk_size`                   | Size of the EBS volume in GB.                                                                                                                                                              |       40        |    number    |  false   |
 | `disk_type`                   | Type of the EBS volume (e.g. standard, gp2, io1).                                                                                                                                          |      "gp2"      |    string    |  false   |
 | `disk_iops`                   | IOPS of the EBS volume (e.g 100).                                                                                                                                                          |        0        |    number    |  false   |
-| `network_mtu`                 | CNI interface MTU. Use 8981 if using instances types with Jumbo frames.                                                                                                                    |      1480       |    number    |  false   |
+| `network_mtu`                 | Physical Network MTU. When using instance types with Jumbo frames, use 9001.                                                                                                               |      1500       |    number    |  false   |
 | `host_cidr`                   | CIDR IPv4 range to assign to EC2 nodes.                                                                                                                                                    |  "10.0.0.0/16"  |    string    |  false   |
 | `pod_cidr`                    | CIDR IPv4 range to assign Kubernetes pods.                                                                                                                                                 |  "10.2.0.0/16"  |    string    |  false   |
 | `service_cidr`                | CIDR IPv4 range to assign Kubernetes services.                                                                                                                                             |  "10.3.0.0/16"  |    string    |  false   |
--- a/docs/configuration-reference/platforms/baremetal.md
+++ b/docs/configuration-reference/platforms/baremetal.md
@@ -96,6 +96,8 @@ cluster "bare-metal" {
 
   enable_tls_bootstrap = true
 
+  encrypt_pod_traffic = true
+
   oidc {
     issuer_url     = var.oidc_issuer_url
     client_id      = var.oidc_client_id
@@ -153,6 +155,7 @@ os_version = var.custom_default_os_version
 | `os_version`                | Flatcar Container Linux version to install. Version such as "2303.3.1" or "current".                                                                                         |    "current"     |    string    |  false   |
 | `os_channel`                | Flatcar Container Linux channel to install from ("flatcar-stable", "flatcar-beta", "flatcar-alpha", "flatcar-edge").                                                         | "flatcar-stable" |    string    |  false   |
 | `enable_tls_bootstrap`      | Enable TLS bootstraping for Kubelet.                                                                                                                                         |       true       |     bool     |  false   |
+| `encrypt_pod_traffic`       | Enable in-cluster pod traffic encryption. If true `network_mtu` is reduced by 60 to make room for the encryption header.                                                     |      false       |     bool     |  false   |
 | `oidc`                      | OIDC configuration block.                                                                                                                                                    |        -         |    object    |  false   |
 | `oidc.issuer_url`           | URL of the provider which allows the API server to discover public signing keys. Only URLs which use the https:// scheme are accepted.                                       |        -         |    string    |  false   |
 | `oidc.client_id`            | A client id that all tokens must be issued for.                                                                                                                              |    "gangway"     |    string    |  false   |
--- a/docs/configuration-reference/platforms/baremetal.md
+++ b/docs/configuration-reference/platforms/baremetal.md
@@ -76,7 +76,7 @@ cluster "bare-metal" {
     "testlabel" = ""
   }
 
-  network_mtu = 1480
+  network_mtu = 1500
 
   controller_domains = var.controller_domains
 
@@ -145,7 +145,7 @@ os_version = var.custom_default_os_version
 | `matchbox_client_key_path`  | Path to the server TLS key file.                                                                                                                                             |        -         |    string    |   true   |
 | `matchbox_endpoint`         | Matchbox API endpoint.                                                                                                                                                       |        -         |    string    |   true   |
 | `matchbox_http_endpoint`    | Matchbox HTTP read-only endpoint. Example: "http://matchbox.example.com:8080"                                                                                                |        -         |    string    |   true   |
-| `network_mtu`               | CNI interface MTU.                                                                                                                                                           |       1480       |    number    |  false   |
+| `network_mtu`               | Physical Network MTU.                                                                                                                                                        |       1500       |    number    |  false   |
 | `worker_names`              | Ordered list of worker names. Example: ["node2", "node3"]                                                                                                                    |        -         | list(string) |   true   |
 | `worker_macs`               | Ordered list of worker identifying MAC addresses. Example ["52:54:00:b2:2f:86", "52:54:00:c3:61:77"]                                                                         |        -         | list(string) |   true   |
 | `worker_domains`            | Ordered list of worker FQDNs. Example ["node2.example.com", "node3.example.com"]                                                                                             |        -         | list(string) |   true   |
--- a/docs/configuration-reference/platforms/packet.md
+++ b/docs/configuration-reference/platforms/packet.md
@@ -111,6 +111,8 @@ cluster "packet" {
 
   enable_tls_bootstrap = true
 
+  encrypt_pod_traffic = true
+
   enable_reporting = false
 
   network_ip_autodetection_method = "first-found"
@@ -221,6 +223,7 @@ node_type = var.custom_default_worker_type
 | `node_private_cidr`                   | Private IPv4 CIDR of the nodes used to allow inter-node traffic. Example "10.0.0.0/8"                                                                                                                                                                                             |        -        |    string    |   true   |
 | `enable_aggregation`                  | Enable the Kubernetes Aggregation Layer.                                                                                                                                                                                                                                          |      true       |     bool     |  false   |
 | `enable_tls_bootstrap`                | Enable TLS bootstraping for Kubelet.                                                                                                                                                                                                                                              |      true       |     bool     |  false   |
+| `encrypt_pod_traffic`                 | Enable in-cluster pod traffic encryption. If true `network_mtu` is reduced by 60 to make room for the encryption header.                                                                                                                                                          |      false      |     bool     |  false   |
 | `network_mtu`                         | Physical Network MTU.                                                                                                                                                                                                                                                             |      1500       |    number    |  false   |
 | `pod_cidr`                            | CIDR IPv4 range to assign Kubernetes pods.                                                                                                                                                                                                                                        |  "10.2.0.0/16"  |    string    |  false   |
 | `service_cidr`                        | CIDR IPv4 range to assign Kubernetes services.                                                                                                                                                                                                                                    |  "10.3.0.0/16"  |    string    |  false   |
--- a/docs/configuration-reference/platforms/packet.md
+++ b/docs/configuration-reference/platforms/packet.md
@@ -100,7 +100,7 @@ cluster "packet" {
 
   cluster_domain_suffix = "cluster.local"
 
-  network_mtu = 1480
+  network_mtu = 1500
 
   tags {
     key1 = "value1"
@@ -221,7 +221,7 @@ node_type = var.custom_default_worker_type
 | `node_private_cidr`                   | Private IPv4 CIDR of the nodes used to allow inter-node traffic. Example "10.0.0.0/8"                                                                                                                                                                                             |        -        |    string    |   true   |
 | `enable_aggregation`                  | Enable the Kubernetes Aggregation Layer.                                                                                                                                                                                                                                          |      true       |     bool     |  false   |
 | `enable_tls_bootstrap`                | Enable TLS bootstraping for Kubelet.                                                                                                                                                                                                                                              |      true       |     bool     |  false   |
-| `network_mtu`                         | CNI interface MTU                                                                                                                                                                                                                                                                 |      1480       |    number    |  false   |
+| `network_mtu`                         | Physical Network MTU.                                                                                                                                                                                                                                                             |      1500       |    number    |  false   |
 | `pod_cidr`                            | CIDR IPv4 range to assign Kubernetes pods.                                                                                                                                                                                                                                        |  "10.2.0.0/16"  |    string    |  false   |
 | `service_cidr`                        | CIDR IPv4 range to assign Kubernetes services.                                                                                                                                                                                                                                    |  "10.3.0.0/16"  |    string    |  false   |
 | `cluster_domain_suffix`               | Cluster's DNS domain.                                                                                                                                                                                                                                                             | "cluster.local" |    string    |  false   |
--- a/pkg/platform/aws/aws.go
+++ b/pkg/platform/aws/aws.go
@@ -83,6 +83,7 @@ type config struct {
 	DisableSelfHostedKubelet bool              `hcl:"disable_self_hosted_kubelet,optional"`
 	OIDC                     *oidc.Config      `hcl:"oidc,block"`
 	EnableTLSBootstrap       bool              `hcl:"enable_tls_bootstrap,optional"`
+	EncryptPodTraffic        bool              `hcl:"encrypt_pod_traffic,optional"`
 	KubeAPIServerExtraFlags  []string
 }
 
--- a/pkg/platform/aws/aws.go
+++ b/pkg/platform/aws/aws.go
@@ -108,6 +108,7 @@ func NewConfig() *config {
 		Region:             "eu-central-1",
 		EnableAggregation:  true,
 		EnableTLSBootstrap: true,
+		NetworkMTU:         platform.NetworkMTU,
 	}
 }
 
--- a/pkg/platform/aws/template.go
+++ b/pkg/platform/aws/template.go
@@ -97,6 +97,11 @@ module "aws-{{.Config.ClusterName}}" {
   {{- end }}
 
   enable_tls_bootstrap    = {{ .Config.EnableTLSBootstrap }}
+
+  {{- if .Config.EncryptPodTraffic }}
+  encrypt_pod_traffic = {{.Config.EncryptPodTraffic}}
+  {{- end }}
+
   worker_bootstrap_tokens = [
     {{- range $index, $pool := .Config.WorkerPools }}
     module.worker-pool-{{ $index }}.worker_bootstrap_token,
--- a/pkg/platform/baremetal/template.go
+++ b/pkg/platform/baremetal/template.go
@@ -30,6 +30,10 @@ module "bare-metal-{{.ClusterName}}" {
   # Enable TLS Bootstrap
   enable_tls_bootstrap = {{ .EnableTLSBootstrap }}
 
+  {{- if .EncryptPodTraffic }}
+  encrypt_pod_traffic = {{ .EncryptPodTraffic }}
+  {{- end }}
+
   # configuration
   cached_install     = "{{.CachedInstall}}"
   k8s_domain_name    = "{{.K8sDomainName}}"
--- a/pkg/platform/packet/packet.go
+++ b/pkg/platform/packet/packet.go
@@ -121,6 +121,7 @@ func NewConfig() *config {
 	return &config{
 		EnableAggregation:  true,
 		EnableTLSBootstrap: true,
+		NetworkMTU:         platform.NetworkMTU,
 	}
 }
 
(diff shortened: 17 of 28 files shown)


assets/charts/control-plane/calico-host-protection/templates/host-protection.yaml
```diff
@@ -9,6 +9,7 @@ metadata:
 spec:
   failsafeInboundHostPorts: []
   bpfLogLevel: Info
+  wireguardEnabled: true
```
invidian, line 12: @iaguis I think we should make it configurable?

knrt10: @invidian name of variable `enable_cluster_encryption` seems fine?

invidian: Maybe `encrypt_pods_traffic`? As IMO encrypting cluster could also mean encrypting etcd data, encrypting disks on nodes etc.

pkg/platform/packet/packet.go
```diff
@@ -94,6 +94,7 @@ type config struct {
 	DisableSelfHostedKubelet bool              `hcl:"disable_self_hosted_kubelet,optional"`
 	OIDC                     *oidc.Config      `hcl:"oidc,block"`
 	EnableTLSBootstrap       bool              `hcl:"enable_tls_bootstrap,optional"`
+	EncryptPodsTraffic       bool              `hcl:"encrypt_pods_traffic,optional"`
```
invidian, line 97: It seems to me that this option should be available on all platforms?

invidian (changes requested):

pkg/platform/packet/packet.go
knrt10: What I found was calico-host-protection is getting applied for packet only.

johananl: @knrt10 true. We don't need Calico's host protection on platforms which provide a VPC functionality (e.g. AWS). This isn't related to encryption though.

knrt10: Yes, I have removed the felix config from calico-host-protection, and added it to calico charts, so that all encryption can be enabled for all clusters

assets/terraform-modules/aws/flatcar-linux/kubernetes/variables.tf
```diff
@@ -185,3 +185,9 @@ variable "kube_apiserver_extra_flags" {
   type        = list(string)
   default     = []
 }
+
+variable "encrypt_pods_traffic" {
```
johananl, line 189:
```suggestion
variable "encrypt_pod_traffic" {
```

"Pods traffic" sounds weird to me. I'd say "pod traffic" even when talking about multiple pods.

iaguis:
We should also change the MTU size when enabling this as mentioned [here](https://docs.projectcalico.org/archive/v3.15/networking/mtu#determine-mtu-size).

Would encryption be picked up by the Calico chart on lokoctl apply? Would the MTU?

assets/charts/control-plane/calico/templates/felixconfig.yaml
```diff
@@ -0,0 +1,11 @@
+# General
+# =======
+---
+# Disable default failsafe inbound rules
+apiVersion: crd.projectcalico.org/v1
+kind: FelixConfiguration
+metadata:
+  name: default
+spec:
+  failsafeInboundHostPorts: []
```
invidian, line 10: This option should be enabled only if we have some GNPs in place, otherwise cluster will loose all the connectivity I think. It should be optional, hidden by bootkube module variable, which Packet will set to true.

assets/charts/control-plane/calico/templates/felixconfig.yaml
```diff
@@ -0,0 +1,11 @@
+# General
+# =======
+---
+# Disable default failsafe inbound rules
```
invidian, line 4: I think those lines can be removed all together, as they do not make sense without the context of Calico host protection file.

assets/charts/control-plane/calico/values.yaml
```diff
@@ -13,3 +13,5 @@ calico:
   podCIDR: 10.2.0.0/16
   networkEncapsulation: "ipipMode: Always"
   blockedMetadataCIDRs: []
+  # Lokomotive specific change
+  encryptPodsTraffic:
```
invidian, line 17:
```suggestion
  # Lokomotive specific change.
  encryptPodsTraffic: false
```

docs/configuration-reference/platforms/aks.md
```diff
@@ -68,6 +68,7 @@ cluster "aks" {
   resource_group_name   = var.resource_group_name
   application_name      = var.application_name
   manage_resource_group = var.manage_resource_group
+  encrypt_pod_traffic = true
```
invidian, line 71: I don't think this will work on AKS.

invidian (changes requested):

assets/charts/control-plane/calico/values.yaml
knrt10: should this be false, do we want encrption by default or not?

invidian: I don't think we want to encrypt by default. @iaguis ?

invidian:
>Would encryption be picked up by the Calico chart on lokoctl apply? Would the MTU?

If encryption is enabled and Calico does not align MTU by default, then I guess we can do it automatically?

assets/charts/control-plane/calico/values.yaml
knrt10: As this is experimental feature, I think this should be false by default.

assets/charts/control-plane/calico/templates/felixconfig.yaml
knrt10: @invidian Do you think this is in scope of this PR? I have no problem adding it as a different commit, just wanted to know.

invidian: Yes, that should be done as part of moving the configuration to bootkube module.

docs/configuration-reference/platforms/aws.md
```diff
@@ -107,6 +107,8 @@ cluster "aws" {
 
   enable_tls_bootstrap = true
 
+  encrypt_pod_traffic = true
```
invidian, line 110: I think docs commit can go together with the implementation.

assets/terraform-modules/packet/flatcar-linux/kubernetes/bootkube.tf
```diff
@@ -42,7 +42,8 @@ module "bootkube" {
   # metadata.packet.net should always resolve to 192.80.8.124.
   blocked_metadata_cidrs = ["192.80.8.124/32"]
 
-  bootstrap_tokens     = var.enable_tls_bootstrap ? concat([local.controller_bootstrap_token], var.worker_bootstrap_tokens) : []
-  enable_tls_bootstrap = var.enable_tls_bootstrap
-  encrypt_pod_traffic  = var.encrypt_pod_traffic
+  bootstrap_tokens            = var.enable_tls_bootstrap ? concat([local.controller_bootstrap_token], var.worker_bootstrap_tokens) : []
+  enable_tls_bootstrap        = var.enable_tls_bootstrap
+  encrypt_pod_traffic         = var.encrypt_pod_traffic
+  failsafe_inbound_host_ports = []
```
invidian, line 48: Can we have a comment here please, explaining the value?
```suggestion
  # We install calico-host-protection chart on Packet which ships GNPs, so we can disable failsafe ports in Calico.
  failsafe_inbound_host_ports = []
```

assets/charts/control-plane/calico/templates/felixconfig.yaml
```diff
@@ -3,6 +3,6 @@ kind: FelixConfiguration
 metadata:
   name: default
 spec:
-  failsafeInboundHostPorts: []
+  failsafeInboundHostPorts: {{ .Values.calico.failsafeInboundHostPorts }}
```
invidian, line 6: I think this can be merged together with first commit.

invidian (changes requested):
Looking good.

assets/charts/control-plane/calico/values.yaml
```diff
@@ -13,3 +13,5 @@ calico:
   podCIDR: 10.2.0.0/16
   networkEncapsulation: "ipipMode: Always"
   blockedMetadataCIDRs: []
+  # Lokomotive specific change.
+  failsafeInboundHostPorts: []
```
invidian, line 17: Shouldn't this here be `~` (null) if we always render it in `failsafeInboundHostPorts: {{ .Values.calico.failsafeInboundHostPorts }}` ?

As we do not override it in `assets/terraform-modules/bootkube/resources/charts/calico.yaml` if value is `null`, it will still be set to `[]`.

invidian (commented):
Just one more question, as I'm not sure if we don't change the existing behavior with failsafe ports.

assets/charts/control-plane/calico/values.yaml
knrt10: Except for Packet, with this implementation if the value is `null` nothing will be applied, because only 2 conditions are checked `%{~ if failsafe_inbound_host_ports != null && failsafe_inbound_host_ports != [] ~}` and `%{~ if failsafe_inbound_host_ports == [] ~}1

invidian: Yes, but this file (`assets/charts/control-plane/calico/values.yaml`) defines default values for the chart, which won't be overwritten. That means the following command should not return anything:
```sh
$ helm template calico ./assets/charts/control-plane/calico | grep failsafeInboundHostPorts
  failsafeInboundHostPorts: []
```

However, it currently does.

knrt10: This is for packet right?

knrt10: aah, I got it what you mean

knrt10: yes, this should be null

invidian (changes requested):
I get the following error while trying this PR:
```
$ ~/repos/kinvolk/lokomotive/lokoctl cluster apply --verbose
ERRO[0000] cluster.lokocfg:33,3-22: Unsupported argument; An argument named "encrypt_pod_traffic" is not expected here.  args="[]" command="lokoctl cluster apply"
FATA[0000] Errors found while loading cluster configuration  args="[]" command="lokoctl cluster apply"
```

Has it been tested?

knrt10:
My bad forgot to add variable for platforms. I was testing it without adding variable.

invidian:
```
Sep 08 15:19:48 ip-10-0-12-134 bootkube-start[3010]: [  282.810139] bootkube[6]: %!(EXTRA []interface {}=[])kube-system/calico: beginning wait for [15 60000000000] resources with timeout of %!v(MISSING)
Sep 08 15:19:51 ip-10-0-12-134 bootkube-start[3010]: [  285.731238] bootkube[6]: kube-system/calico: creating [1] resource(s)
Sep 08 15:19:51 ip-10-0-12-134 bootkube-start[3010]: [  285.785543] bootkube[6]: kube-system/calico: creating [12] resource(s)
Sep 08 15:19:51 ip-10-0-12-134 bootkube-start[3010]: [  285.964715] bootkube[6]: Error: FelixConfiguration.crd.projectcalico.org "default" is invalid: spec.failsafeInboundHostPorts: Invalid value: "null": spec.failsafeInboundHostPorts in body must be of type array: "null"
Sep 08 15:19:51 ip-10-0-12-134 bootkube-start[3010]: [  285.965807] bootkube[6]: Tearing down temporary bootstrap control plane...
Sep 08 15:19:51 ip-10-0-12-134 bootkube-start[3010]: [  285.980467] bootkube[6]: Error: FelixConfiguration.crd.projectcalico.org "default" is invalid: spec.failsafeInboundHostPorts: Invalid value: "null": spec.failsafeInboundHostPorts in body must be of type array: "null"
Sep 08 15:19:51 ip-10-0-12-134 bootkube-start[3010]: [  285.981511] bootkube[6]: Error: FelixConfiguration.crd.projectcalico.org "default" is invalid: spec.failsafeInboundHostPorts: Invalid value: "null": spec.failsafeInboundHostPorts in body must be of type array: "null"
```

Still doesn't work it seems.

I think we should include this flag in the CI, to have at least some sanity checks.

knrt10:
I think we cannot use it as null, it should be [] only, what do you think @invidian? Because as the error suggests, it should be of type array only. And I think using [] should not create any problem.

invidian:
>I think we cannot use failsafeInboundHostPorts as null or empty, it should be [] only, what do you think @invidian? Because as the error suggests, it should be of type array only. And I think using [] should not create any problem.

We can just not render the field if it's `nil`, would that work in helm template?

knrt10:
I think it should, will try it

assets/charts/control-plane/calico/templates/felixconfig.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: crd.projectcalico.org/v1
+kind: FelixConfiguration
+metadata:
+  name: default
+spec:
+  {{- if .Values.calico.failsafeInboundHostPorts }}
```
invidian, line 6:
```suggestion
  {{- if hasKey .Values.calico "failsafeInboundHostPorts" }}
```

assets/charts/control-plane/calico/values.yaml
```diff
@@ -13,3 +13,6 @@ calico:
   podCIDR: 10.2.0.0/16
   networkEncapsulation: "ipipMode: Always"
   blockedMetadataCIDRs: []
+  # Lokomotive specific change.
+  failsafeInboundHostPorts:
```
invidian, line 17:
```suggestion
  #failsafeInboundHostPorts:
```

invidian (changes requested):
I think suggested changes should work as expected:
```
0 ✓ (140ms) 09:39:19 invidian@dellxps15mateusz ~/repos/kinvolk/lokomotive (pr/911)*$ helm template calico ./assets/charts/control-plane/calico | grep failsafeInboundHostPorts
1 ✗ (94.6ms) 09:39:28 invidian@dellxps15mateusz ~/repos/kinvolk/lokomotive (pr/911)*$ cat values.yaml
calico:
  failsafeInboundHostPorts: []
0 ✓ (50.4ms) 09:39:31 invidian@dellxps15mateusz ~/repos/kinvolk/lokomotive (pr/911)*$ helm template calico -f values.yaml ./assets/charts/control-plane/calico | grep failsafeInboundHostPorts
  failsafeInboundHostPorts: []
```

assets/charts/control-plane/calico/templates/felixconfig.yaml
knrt10: Thanks for the suggestion @invidian, but with this change, `[]` for packet it's not getting applied

knrt10: I am confused because `helm template` is getting rendered correctly also, 

```bash
➜ helm get values calico -n kube-system                                                                  
USER-SUPPLIED VALUES:
calico:
  blockedMetadataCIDRs:
  - 192.80.8.124/32
  cniImage: calico/cni:v3.15.2
  controllersImage: calico/kube-controllers:v3.15.2
  enableReporting: false
  encryptPodsTraffic: true
  failsafe_inbound_host_ports: []
  flexvolDriverImage: calico/pod2daemon-flexvol:v3.15.2
  image: calico/node:v3.15.2
  ipipEnabled: true
  networkEncapsulation: 'ipipMode: Always'
  networkIpAutodetectionMethod: can-reach=10.0.0.1
  networkMTU: 1480
  podCIDR: 10.2.0.0/16
  vxlanEnabled: false
```

it shows that `failsafe_inbound_host_ports` value is correctly getting rendered

assets/terraform-modules/bootkube/resources/charts/calico.yaml
```diff
@@ -16,3 +16,14 @@ calico:
   - ${cidr}
   %{~ endfor ~}
   %{~ endif ~}
+  %{~ if failsafe_inbound_host_ports != null && failsafe_inbound_host_ports != [] ~}
+  failsafe_inbound_host_ports:
```
invidian, line 20:
```suggestion
  failsafeInboundHostPorts:
```

assets/terraform-modules/bootkube/resources/charts/calico.yaml
```diff
@@ -16,3 +16,14 @@ calico:
...
+  %{~ for protoport in failsafe_inbound_host_ports ~}
+  - protocol: ${protoport.protocol}
+    port: ${protoport.port}
+  %{~ endfor ~}
+  %{~ endif ~}
+  %{~ if failsafe_inbound_host_ports == [] ~}
+  failsafe_inbound_host_ports: []
```
invidian, line 27:
```suggestion
  failsafeInboundHostPorts: []
```

invidian (changes requested):

assets/terraform-modules/bootkube/resources/charts/calico.yaml
knrt10: 🤦

assets/terraform-modules/bootkube/resources/charts/calico.yaml
knrt10: 🤦

knrt10:
@invidian tested and updated.

invidian (approved):
LGTM, though we don't have a way to test it right now because of Flatcar bug affecting Alpha channel, where wireguard module is implemented. 

@iaguis could you have a look on it so we can merge it?

invidian (dismissed):

docs/configuration-reference/platforms/packet.md
```diff
@@ -221,7 +223,8 @@ node_type = var.custom_default_worker_type
...
-| `network_mtu`                         | CNI interface MTU                                                                                                                                                                                                                                                                 |      1480       |    number    |  false   |
+| `encrypt_pod_traffic`                 | Enable in-cluster pod traffic.                                                                                                                                                                                                                                                    |      false      |     bool     |  false   |
+| `network_mtu`                         | CNI interface MTU. Use 1440 if `encrypt_pod_traffic` = true                                                                                                                                                                                                                       |      1480       |    number    |  false   |
```
iaguis, line 227:
```suggestion
| `network_mtu`                         | CNI interface MTU. Use 1440 if `encrypt_pod_traffic` = true.                                                                                                                                                                                                                      |      1480       |    number    |  false   |
```

docs/configuration-reference/platforms/aws.md
```diff
@@ -219,10 +221,11 @@ worker_pool "my-worker-pool" {
...
 | `disk_type`                   | Type of the EBS volume (e.g. standard, gp2, io1).                                                                                                                                          |      "gp2"      |    string    |  false   |
 | `disk_iops`                   | IOPS of the EBS volume (e.g 100).                                                                                                                                                          |        0        |    number    |  false   |
-| `network_mtu`                 | CNI interface MTU. Use 8981 if using instances types with Jumbo frames.                                                                                                                    |      1480       |    number    |  false   |
+| `network_mtu`                 | CNI interface MTU. Use 1440 if `encrypt_pod_traffic = true`. Use 8981 if using instances types with Jumbo frames and `encrypt_pod_traffic = false` else use 8941.                          |      1480       |    number    |  false   |
```
iaguis, line 228:
```suggestion
| `network_mtu`                 | CNI interface MTU. Use 1440 if `encrypt_pod_traffic = true`. When using instance types with Jumbo frames, use 8981 (or 8941 if `encrypt_pod_traffic = true`).                          |      1480       |    number    |  false   |
```

docs/configuration-reference/platforms/aws.md
```diff
@@ -219,10 +221,11 @@ worker_pool "my-worker-pool" {
 | `region`                      | AWS region to use for deploying the cluster.                                                                                                                                               | "eu-central-1"  |    string    |  false   |
 | `enable_aggregation`          | Enable the Kubernetes Aggregation Layer.                                                                                                                                                   |      true       |     bool     |  false   |
 | `enable_tls_bootstrap`        | Enable TLS bootstraping for Kubelet.                                                                                                                                                       |      true       |     bool     |  false   |
+| `encrypt_pod_traffic`         | Enable in-cluster pod traffic.                                                                                                                                                             |      false      |     bool     |  false   |
```
iaguis, line 224:
```suggestion
| `encrypt_pod_traffic`         | Enable in-cluster pod traffic. If true you need to reduce the default `network_mtu`, see its documentation.                                                                                                                                                             |      false      |     bool     |  false   |
```

docs/configuration-reference/platforms/packet.md
```diff
@@ -221,7 +223,8 @@ node_type = var.custom_default_worker_type
...
 | `enable_tls_bootstrap`                | Enable TLS bootstraping for Kubelet.                                                                                                                                                                                                                                              |      true       |     bool     |  false   |
-| `network_mtu`                         | CNI interface MTU                                                                                                                                                                                                                                                                 |      1480       |    number    |  false   |
+| `encrypt_pod_traffic`                 | Enable in-cluster pod traffic.                                                                                                                                                                                                                                                    |      false      |     bool     |  false   |
```
iaguis, line 226:
```suggestion
| `encrypt_pod_traffic`                 | Enable in-cluster pod traffic.  If true you need to reduce the default `network_mtu`, see its documentation.                                                                                                                                                                                                                                                       |      false      |     bool     |  false   |
```

docs/configuration-reference/platforms/baremetal.md
```diff
@@ -149,7 +151,8 @@ os_version = var.custom_default_os_version
...
 | `os_channel`                | Flatcar Container Linux channel to install from ("flatcar-stable", "flatcar-beta", "flatcar-alpha", "flatcar-edge").                                                         | "flatcar-stable" |    string    |  false   |
-| `enable_tls_bootstrap`      | Enable TLS bootstraping for Kubelet.                                                                                                                                         |      true        |     bool     |  false   |
+| `enable_tls_bootstrap`      | Enable TLS bootstraping for Kubelet.                                                                                                                                         |       true       |     bool     |  false   |
+| `encrypt_pod_traffic`       | Enable in-cluster pod traffic.                                                                                                                                               |      false       |     bool     |  false   |
```
iaguis, line 155: We don't expose the MTU here, I created an issue for this: https://github.com/kinvolk/lokomotive/issues/976

iaguis (dismissed):
Some changes.

I was going to suggest documenting the nodes should have kernel Wireguard support, but since the next Flatcar Stable (the only channel without it) is due next week and will include this I think it's OK.

iaguis:
Also, I tested this on Flatcar Alpha on a machine not affected by the [pod-to-pod connectivity bug](https://github.com/flatcar-linux/Flatcar/issues/183) and it worked fine for me.

docs/configuration-reference/platforms/aws.md
invidian: I thought we could subtract it automatically, so they user don't run into troubles if they don't set it up?

knrt10:
@iaguis @invidian updated

docs/configuration-reference/platforms/aws.md
iaguis: That would be ideal but I'd be ok with having the user set this manually for now.

iaguis: If it's easy to do we should just do it of course.

docs/configuration-reference/platforms/baremetal.md
knrt10: created PR for it #977

docs/configuration-reference/platforms/aws.md
invidian: @knrt10 can you try implementing it? I think we could put it at Terraform template level, something like:
```
{{ if .Config.EncruptPodTraffic }}
network_mtu = {{ .Config.NetworkMTU }} - 40 # Wireguard overhead
{{- else }}
network_mtu = {{ .Config.NetworkMTU }}
{{- end }}
```

knrt10: Done, but it had to be implemented little differently, because we were setting default value in terraform, instead of that did that in go code, and it works

invidian (dismissed):
LGTM, thanks @knrt10

assets/charts/control-plane/calico/values.yaml
```diff
@@ -13,3 +13,6 @@ calico:
   podCIDR: 10.2.0.0/16
   networkEncapsulation: "ipipMode: Always"
   blockedMetadataCIDRs: []
+  # Lokomotive specific change.
+  # failsafeInboundHostPorts:
+  encryptPodsTraffic: false
```
iaguis, line 18: Can we name this `encryptPodTraffic` to match the name of the variable in terraform?

docs/configuration-reference/platforms/aws.md
```diff
@@ -219,10 +221,11 @@ worker_pool "my-worker-pool" {
 | `region`                      | AWS region to use for deploying the cluster.                                                                                                                                               | "eu-central-1"  |    string    |  false   |
 | `enable_aggregation`          | Enable the Kubernetes Aggregation Layer.                                                                                                                                                   |      true       |     bool     |  false   |
 | `enable_tls_bootstrap`        | Enable TLS bootstraping for Kubelet.                                                                                                                                                       |      true       |     bool     |  false   |
+| `encrypt_pod_traffic`         | Enable in-cluster pod traffic. If true you need to reduce the default `network_mtu`, see its documentation.                                                                                |      false      |     bool     |  false   |
```
iaguis, line 224: This doesn't apply anymore. Also, the first sentence doesn't make sense. How about this?

```suggestion
| `encrypt_pod_traffic`         | Enable in-cluster pod traffic encryption. If true `network_mtu` is reduced by 40 to make room for the encryption header.                                                                                |      false      |     bool     |  false   |
```

Please change all the other platforms as well.

docs/configuration-reference/platforms/aws.md
```diff
@@ -219,10 +221,11 @@ worker_pool "my-worker-pool" {
...
 | `disk_type`                   | Type of the EBS volume (e.g. standard, gp2, io1).                                                                                                                                          |      "gp2"      |    string    |  false   |
 | `disk_iops`                   | IOPS of the EBS volume (e.g 100).                                                                                                                                                          |        0        |    number    |  false   |
-| `network_mtu`                 | CNI interface MTU. Use 8981 if using instances types with Jumbo frames.                                                                                                                    |      1480       |    number    |  false   |
+| `network_mtu`                 | CNI interface MTU. Use 1440 if `encrypt_pod_traffic = true`. When using instance types with Jumbo frames, use 8981 (or 8941 if `encrypt_pod_traffic = true`).                              |      1480       |    number    |  false   |
```
iaguis, line 228: This doesn't apply anymore. How about this?

```suggestion
| `network_mtu`                 | CNI interface MTU. When using instance types with Jumbo frames, use 8981.                              |      1480       |    number    |  false   |
```

docs/configuration-reference/platforms/baremetal.md
```diff
@@ -145,14 +147,15 @@ os_version = var.custom_default_os_version
...
 | `matchbox_endpoint`         | Matchbox API endpoint.                                                                                                                                                       |        -         |    string    |   true   |
 | `matchbox_http_endpoint`    | Matchbox HTTP read-only endpoint. Example: "http://matchbox.example.com:8080"                                                                                                |        -         |    string    |   true   |
-| `network_mtu`               | CNI interface MTU.                                                                                                                                                           |       1480       |    number    |  false   |
+| `network_mtu`               | CNI interface MTU. Use 1440 if `encrypt_pod_traffic` = true.                                                                                                                 |       1480       |    number    |  false   |
```
iaguis, line 150:
```suggestion
| `network_mtu`               | CNI interface MTU.                                                                                                                 |       1480       |    number    |  false   |
```

docs/configuration-reference/platforms/baremetal.md
```diff
@@ -145,14 +147,15 @@ os_version = var.custom_default_os_version
...
 | `os_version`                | Flatcar Container Linux version to install. Version such as "2303.3.1" or "current".                                                                                         |    "current"     |    string    |  false   |
 | `os_channel`                | Flatcar Container Linux channel to install from ("flatcar-stable", "flatcar-beta", "flatcar-alpha", "flatcar-edge").                                                         | "flatcar-stable" |    string    |  false   |
 | `enable_tls_bootstrap`      | Enable TLS bootstraping for Kubelet.                                                                                                                                         |       true       |     bool     |  false   |
+| `encrypt_pod_traffic`       | Enable in-cluster pod traffic.  If true you need to reduce the default `network_mtu`, see its documentation.                                                                 |      false       |     bool     |  false   |
```
iaguis, line 158: Change this as mentioned above.

docs/configuration-reference/platforms/packet.md
```diff
@@ -221,7 +223,8 @@ node_type = var.custom_default_worker_type
...
 | `enable_tls_bootstrap`                | Enable TLS bootstraping for Kubelet.                                                                                                                                                                                                                                              |      true       |     bool     |  false   |
-| `network_mtu`                         | CNI interface MTU                                                                                                                                                                                                                                                                 |      1480       |    number    |  false   |
+| `encrypt_pod_traffic`                 | Enable in-cluster pod traffic.  If true you need to reduce the default `network_mtu`, see its documentation.                                                                                                                                                                      |      false      |     bool     |  false   |
```
iaguis, line 226: Change this as mentioned above.

docs/configuration-reference/platforms/packet.md
```diff
@@ -221,7 +223,8 @@ node_type = var.custom_default_worker_type
...
+| `network_mtu`                         | CNI interface MTU. Use 1440 if `encrypt_pod_traffic` = true.                                                                                                                                                                                                                      |      1480       |    number    |  false   |
```
iaguis, line 227:
```suggestion
| `network_mtu`                         | CNI interface MTU.                                                                                                                                                                                                                      |      1480       |    number    |  false   |
```

iaguis (dismissed):

docs/configuration-reference/platforms/baremetal.md
invidian: Perhaps we should note, that MTU will be reduced by wireguard overhead if this is true?

invidian (dismissed):

assets/terraform-modules/bare-metal/flatcar-linux/kubernetes/variables.tf
```diff
@@ -85,7 +85,6 @@ variable "asset_dir" {
 variable "network_mtu" {
   description = "CNI interface MTU"
   type        = number
-  default     = 1480
```
ipochi, line 88: Why is the default value removed ?

invidian: So we're able to modify it from the template to account for wireguard overhead.

ipochi (dismissed):
Great work @knrt10 

LGTM :tada:

assets/charts/control-plane/calico-host-protection/templates/host-protection.yaml
```diff
@@ -1,15 +1,3 @@
-# General
-# =======
----
-# Disable default failsafe inbound rules
-apiVersion: crd.projectcalico.org/v1
-kind: FelixConfiguration
-metadata:
-  name: default
```
invidian, line 8: Hm, as we move this object from `calico-host-protection` chart to `calico`, I guess we need to require from the user to remove this object manually before rolling an upgrade. i wonder if it has some implications like weakened rules for upgrade time (as failsafe port won't be emptied) OR that networking breaks for some reason.

I think we should test that.

knrt10: Tested the upgrade process and it involves deleting `default FelixConfiguration` while upgrading.

invidian: I suggest we add "Release notes" section in PR description then, like in https://github.com/kinvolk/lokomotive/pull/946.

invidian: By looking at the config we set and default values in Calico, following changes will be done during the upgrade:
- `bpfLogLevel` will be briefly set from `Info` to `Off`.
- Following ports will be opened on all hosts:
  ```
  - protocol: tcp
    port: 22
  - protocol: udp
    port: 68
  - protocol: tcp
    port: 179
  - protocol: tcp
    port: 2379
  - protocol: tcp
    port: 2380
  - protocol: tcp
    port: 5473
  - protocol: tcp
    port: 6443
  - protocol: tcp
    port: 6666
  - protocol: tcp
    port: 6667
  ```

I _think_ if we include that in release notes, it should be fine, given that this is only one time thing during the upgrade. Thoughts @iaguis @johananl ?

knrt10: @invidian done

ipochi: Isn't it becoming a lot of hassle for a released deemed as a `maintainence` release ?

invidian: @ipochi what do you mean? The PR is almost complete, why wouldn't we merge it?

ipochi: I am not opposed to merging it.

The step of removal of felix configuration from its current chart of `calico-host-protection`, can this be avoided and not be touched for the scope of this PR ? Would that mean less hassles on the upgrade side in this release ?

invidian: >The step of removal of felix configuration from its current chart of calico-host-protection, can this be avoided and not be touched for the scope of this PR ? Would that mean less hassles on the upgrade side in this release ?

Yes, it could possibly be avoided, but it will be messy. I think we should clean it up like this PR does. The upgrade steps are not so bad.

ipochi: Ok, I have no objections if others are okay with it.

iaguis: A couple of things.

* @ipochi not sure what you mean about a `maintenance` release. But if we include this in the next release it needs to be a minor release. Otherwise we'd be releasing a patch release that includes new features which is something we don't want.
* I find the syntax used when mentioning what ports are going to be opened during the upgrade weird, I'd use something like:

    * 22/tcp
    * 68/udp
    * ...

Otherwise I think it's ok to have these instructions for upgrades and have a small window of open ports if we warn about it. If we want to provide a more secure option, we could provide a temporary GlobalNetworkPolicy covering those ports and ask the user to apply it before the upgrade and remove it after.

knrt10: @iaguis updated description

invidian: >Otherwise I think it's ok to have these instructions for upgrades and have a small window of open ports if we warn about it. If we want to provide a more secure option, we could provide a temporary GlobalNetworkPolicy covering those ports and ask the user to apply it before the upgrade and remove it after.

Just a note: temporary GNP won't work here, as safePorts bypasses it anyway as far as I understand.

pkg/platform/baremetal/template.go
```diff
@@ -30,6 +30,13 @@ module "bare-metal-{{.ClusterName}}" {
   # Enable TLS Bootstrap
   enable_tls_bootstrap = {{ .EnableTLSBootstrap }}
 
+  {{- if .EncryptPodTraffic }}
+  encrypt_pod_traffic = {{ .EncryptPodTraffic }}
+  network_mtu = {{ .NetworkMTU }} - 40 # Wireguard overhead
```
johananl, line 35: I don't think this makes sense. IIUC we have two possible values for the `network_mtu` knob:

1. Physical network MTU minus 20 (when IP-IP is used)
2. Physical network MTU minus 40 (when WireGuard is used)

We currently expose a knob called `network_mtu` which has a default value of `1480`. I'm assuming that right now this knob means "the actual MTU to use" as opposed to "the MTU of the **physical** network".

I think we want to do one of the following:

- Decide that `network_mtu` means "the actual MTU to use" and then when WireGuard is used we explicitly set it to `1440`.
- Decide that `network_mtu` means "the MTU of the physical network", set the default to `1500` (for Ethernet) and internally calculate the final value as `network_mtu - 20` for IP-IP and `network_mtu - 60` for WireGuard.

Subtracting 40 from the **IP-IP** value doesn't make sense IMO.

johananl:
I think we should make it a habit to include a "how to test" section in PRs. I just tried using this feature using Flatcar `stable` which doesn't work because the kernel is too old.

assets/terraform-modules/aws/flatcar-linux/kubernetes/variables.tf
```diff
@@ -185,3 +184,9 @@ variable "kube_apiserver_extra_flags" {
   type        = list(string)
   default     = []
 }
+
+variable "encrypt_pod_traffic" {
+  description = "Enable in-cluster pod traffic."
```
johananl, line 189: This description doesn't mention encryption. If you fix this, please fix all occurrences.

assets/terraform-modules/bootkube/conditional.tf
```diff
@@ -15,6 +15,14 @@ resource "local_file" "calico" {
     pod_cidr                        = var.pod_cidr
     enable_reporting                = var.enable_reporting
     blocked_metadata_cidrs          = var.blocked_metadata_cidrs
+    failsafe_inbound_host_ports = var.failsafe_inbound_host_ports != null ? [
```
johananl, line 18: I don't understand this change.

assets/terraform-modules/bootkube/resources/charts/calico.yaml
```diff
@@ -16,3 +16,14 @@ calico:
   - ${cidr}
   %{~ endfor ~}
   %{~ endif ~}
+  %{~ if failsafe_inbound_host_ports != null && failsafe_inbound_host_ports != [] ~}
```
johananl, line 19: I don't understand this change.

assets/terraform-modules/bootkube/variables.tf
```diff
@@ -173,3 +173,15 @@ variable "enable_tls_bootstrap" {
   description = "Enable TLS Bootstrap for Kubelet."
   type        = bool
 }
+
+variable "failsafe_inbound_host_ports" {
+  description = "UDP/TCP/SCTP protocol/port pairs that Felix will allow incoming traffic to host endpoints on irrespective of the security policy."
```
johananl, line 178:
```suggestion
  description = "UDP/TCP/SCTP protocol/port pairs to allow incoming traffic on regardless of the security policy."
```

- Avoid the future tense.
- Simplify the phrasing and make it less ambiguous.

Also, what format is expected here? `tcp/123` for example?

johananl (commented):
The encryption seems to work fine provided that you don't use `stable`. Nice job @knrt10.

I've left a couple of comments.

docs/configuration-reference/platforms/packet.md
```diff
@@ -111,6 +111,8 @@ cluster "packet" {
 
   enable_tls_bootstrap = true
 
+  encrypt_pod_traffic = true
```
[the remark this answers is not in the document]
knrt10: I think it's not possible because calico values are set by bootkube module.

assets/terraform-modules/bootkube/conditional.tf
knrt10: Previously we were just setting `failsafeInboundHostPorts` in calico-host-protection chart as []. Which should not be case, for more information https://github.com/kinvolk/lokomotive/pull/911#discussion_r484406367

assets/terraform-modules/bootkube/resources/charts/calico.yaml
knrt10: explained above

assets/terraform-modules/bootkube/variables.tf
knrt10: Format here is 

```yaml
- protocol: tcp
  port: 22
```

pkg/platform/baremetal/template.go
invidian: @johananl is right, except that Wireguard use 60 bytes header: https://docs.projectcalico.org/networking/mtu#determine-mtu-size. It would be good to clarify the meaning of this flag. I think we could refine it as physical network MTU and then substract 20 from it by default, which preserves existing behavior.

knrt10: Updated

pkg/platform/aws/template.go
```diff
@@ -97,6 +94,14 @@ module "aws-{{.Config.ClusterName}}" {
   {{- end }}
 
   enable_tls_bootstrap    = {{ .Config.EnableTLSBootstrap }}
+
+  {{- if .Config.EncryptPodTraffic }}
+  encrypt_pod_traffic = {{.Config.EncryptPodTraffic}}
+  network_mtu = {{ .Config.NetworkMTU }} - 40 # Wireguard overhead
```
invidian, line 100: Wireguard overhead is 60 bytes, not 40.

docs/configuration-reference/platforms/aws.md
```diff
@@ -219,10 +221,11 @@ worker_pool "my-worker-pool" {
...
 | `disk_type`                   | Type of the EBS volume (e.g. standard, gp2, io1).                                                                                                                                          |      "gp2"      |    string    |  false   |
 | `disk_iops`                   | IOPS of the EBS volume (e.g 100).                                                                                                                                                          |        0        |    number    |  false   |
-| `network_mtu`                 | CNI interface MTU. Use 8981 if using instances types with Jumbo frames.                                                                                                                    |      1480       |    number    |  false   |
+| `network_mtu`                 | CNI interface MTU. When using instance types with Jumbo frames, use 8981.                                                                                                                  |      1480       |    number    |  false   |
```
invidian, line 228: This change seems unrelated to the adding of encryption support.

assets/terraform-modules/aws/flatcar-linux/kubernetes/variables.tf
```diff
@@ -113,7 +113,6 @@ variable "asset_dir" {
 variable "network_mtu" {
   description = "CNI interface MTU. Use 8981 if using instances types with Jumbo frames."
```
invidian, line 114: If we change the meaning of this variable, from "CNI interface MTU" to "MTU of physical network of the servers", we should change it accordingly. Also value for jumbo frames should be adjusted. Also, how do we support enabling jumbo frames? Perhaps this can be removed from the description?

pkg/platform/aws/template.go
```diff
@@ -100,6 +97,9 @@ module "aws-{{.Config.ClusterName}}" {
 
   {{- if .Config.EncryptPodTraffic }}
   encrypt_pod_traffic = {{.Config.EncryptPodTraffic}}
+  network_mtu = {{ .Config.NetworkMTU }} - 40 # Wireguard overhead
+  {{- else }}
+  network_mtu = {{ .Config.NetworkMTU }} - 20 # IP in IP overhead
```
invidian, line 102: I think this change of meaning of `network_mtu` parameter should be done in separate commit, before we add encryption support.

Also, I wonder at which point we should reduce the MTU. Maybe it should be done on the Helm template level, so the mutation is closer to the point where it's used?

pkg/platform/aws/aws.go
```diff
@@ -109,6 +109,7 @@ func NewConfig() *config {
 		Region:             "eu-central-1",
 		EnableAggregation:  true,
 		EnableTLSBootstrap: true,
+		NetworkMTU:         1500, //nolint:gomnd
```
invidian, line 112: It seems this could be a constant. Perhaps we could put it in `platform` package?

invidian (changes requested):

docs/configuration-reference/platforms/aws.md
knrt10: It was suggested by Iago or Johannes for fixing the sentences I guess. Should I add it in a separate commit?

invidian: Yes, please.

assets/terraform-modules/aws/flatcar-linux/kubernetes/variables.tf
```diff
@@ -111,9 +111,8 @@ variable "asset_dir" {
 }
 
 variable "network_mtu" {
-  description = "CNI interface MTU. Use 8981 if using instances types with Jumbo frames."
+  description = "Physical Network MTU. Use 8981 if using instances types with Jumbo frames."
```
invidian, line 114: Should this be 9000 for Jumbo frames?

pkg/platform/platform.go
```diff
@@ -27,6 +27,9 @@ import (
 	"github.com/kinvolk/lokomotive/pkg/version"
 )
 
+// NetworkMTU is size of header for calico MTU.
```
invidian, line 30:
```suggestion
// NetworkMTU is the default host network MTU.
```

assets/charts/control-plane/calico/templates/config.yaml
```diff
@@ -13,7 +13,11 @@ data:
   # - Otherwise, if VXLAN or BPF mode is enabled, set to your network MTU - 50
   # - Otherwise, if IPIP is enabled, set to your network MTU - 20
   # - Otherwise, if not using any encapsulation, set to your network MTU.
+  {{- if .Values.calico.encryptPodTraffic }}
+  veth_mtu: "{{ sub .Values.calico.networkMTU 60 }}" # Wireguard overhead
```
invidian, line 17:
```suggestion
  veth_mtu: "{{ sub .Values.calico.networkMTU 60 }}" # Wireguard overhead.
```

invidian (changes requested):
Just some last nits, otherwise LGTM.

assets/terraform-modules/aws/flatcar-linux/kubernetes/variables.tf
knrt10: I think 9001 for AWS

docs/configuration-reference/platforms/aws.md
```diff
@@ -113,7 +113,7 @@ cluster "aws" {
 
   disk_iops = var.disk_iops
 
-  network_mtu = 1480
```
invidian, line 116: I think this commit could be merged with the one changing the code.

docs/configuration-reference/platforms/aws.md
invidian, line 110: Docs can be committed with the implementation.

invidian (approved):
Just some commit structure nits, no blockers from merging really.

Nice work @knrt10, LGTM :heavy_check_mark:

iaguis (approved):
LGTM. We should mention in the release notes that we've changed the meaning of the `network_mtu` parameter.

knrt10:
Thank you all for the reviews. Merging now.

docs/configuration-reference/platforms/packet.md
```diff
@@ -221,7 +223,8 @@ node_type = var.custom_default_worker_type
...
 | `enable_tls_bootstrap`                | Enable TLS bootstraping for Kubelet.                                                                                                                                                                                                                                              |      true       |     bool     |  false   |
-| `network_mtu`                         | CNI interface MTU                                                                                                                                                                                                                                                                 |      1480       |    number    |  false   |
+| `encrypt_pod_traffic`                 | Enable in-cluster pod traffic encryption. If true `network_mtu` is reduced by 60 to make room for the encryption header.                                                                                                                                                          |      false      |     bool     |  false   |
```
arianvp, line 226: `network_mtu` stays the same right?  It's just that the `mtu` _inside_ the tunnel gets reduced by `60`

invidian: Yes, value configured by the user stays the same. However, if you enable encryption, then it gets reduced by additional `40`, as IPIP overhead is `20` and wireguard encryption is `60`.

arianvp: got it. thanks

PR #925, opened 2020-09-09
feat: add severity labels to MetalLB alerts

niels-s:
Add the severity label "critical" to the MetalLB alerts, using the same
approach as the alerts configured by the prometheus-operator.

When any of these alerts are triggered, this could impact any
load balancer/ingress, therefore labeling them as critical.

### DIFF
--- a/pkg/components/metallb/manifests.go
+++ b/pkg/components/metallb/manifests.go
@@ -948,24 +948,32 @@ spec:
     - alert: MetalLBNoBGPSession
       expr: metallb_bgp_session_up != 1
       for: 2m
+      labels:
+        severity: critical
       annotations:
         description: '{{ $labels.instance }}: MetalLB has not established a BGP session for more than 2 minutes.'
         summary: '{{ $labels.instance }}: MetalLB has not established BGP session.'
     - alert: MetalLBConfigStale
       expr: metallb_k8s_client_config_stale_bool != 0
       for: 2m
+      labels:
+        severity: critical
       annotations:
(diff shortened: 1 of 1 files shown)


pkg/components/metallb/manifests.go
```diff
@@ -948,24 +948,32 @@ spec:
     - alert: MetalLBNoBGPSession
       expr: metallb_bgp_session_up != 1
       for: 2m
+      labels:
+        severity: critical
```
invidian, line 952: Hm, I see 2 conventions around those labels. One is to use `page` and another to use `critical`. What consumes this label later on?

niels-s: Do you have a reference to the `page` level, because I can't find any references to it when [searching this repo](https://github.com/kinvolk/lokomotive/search?q=page&unscoped_q=page) ? When I check our current lokomotive cluster, I don't find any severity labels with `page`, only `none`, `warning` or `critical`.
`kubectl get prometheusrule --all-namespaces -oyaml | grep "page"`

note: we don't have all the components installed so could be we are missing some alert rules.

invidian: Ah yes, I should've grep the code before asking. Sorry, I'm not familiar with alert manager conventions. But it seems prometheus-operator ships many `severity: critical`, so it seems reasonable to just follow that.

invidian (approved):
LGTM

surajssd (approved):
LGTM

PR #931, opened 2020-09-09
component/metallb: Fix controller tolerations

johananl:
This fixes a bug which causes MetalLB speaker tolerations to be applied to controller pods, too.

Tests should be added as part of https://github.com/kinvolk/lokomotive/pull/927.


invidian (approved):

surajssd (approved):
LGTM

PR #935, opened 2020-09-10
FLUO: convert to Helm chart

invidian:
So all components are consistently sourced from Helm charts.

Signed-off-by: Mateusz Gozdek <mateusz@kinvolk.io>

Issue #817, opened 2020-08-20
Remove specific assets paths from assets generation configuration


invidian:

knrt10:
@invidian is this still relevant?

invidian:
>@invidian is this still relevant?

Yes.

PR #936, opened 2020-09-10
pkg/assets: cleanup exported API

### LINKED ISSUE #796: Revamp assets API
Our current API for doing operation on assets looks like following:
| assets                                                                                                                 | walkers                                                                                                                |
|------------------------------------------------------------------------------------------------------------------------|------------------------------------------------------------------------------------------------------------------------|
| Selection_219 | Selection_220 |

It would be better, if we could have something like following to be consumed by components and platforms code:
```
assets.Read(path, extensions string) (map[string]string, error)
assets.Copy(sourcePath, destinationPath string) error
```

Also, while doing that, we can keep #795 in mind.

invidian:
~**Includes commits from #935**~

This PR finishes cleaning up API of `pkg/assets` package, so end result is, that only exported function we have is `Extract()`.

See commit messages for more details.

Closes #796

### DIFF
--- a/pkg/assets/assets_generate.go
+++ b/pkg/assets/assets_generate.go
@@ -16,19 +16,36 @@
 package main
 
 import (
-	"log"
+	"net/http"
+	"time"
+
+	// External dependencies should also be added to assets.go file.
+	"github.com/prometheus/alertmanager/pkg/modtimevfs"
+	"github.com/shurcooL/httpfs/union"
+	"github.com/shurcooL/vfsgen"
 
 	"github.com/kinvolk/lokomotive/pkg/assets"
 )
 
 func main() {
-	dirs := map[string]string{
-		assets.TerraformModulesSource: "../../assets/terraform-modules",
-		assets.ControlPlaneSource:     "../../assets/charts/control-plane",
-		assets.ComponentsSource:       "../../assets/charts/components",
+	directoriesToEmbed := map[string]http.FileSystem{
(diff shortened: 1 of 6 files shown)


invidian:
ping @surajssd @johananl

surajssd (dismissed):
LGTM, can you rebase on latest master and trigger the CI again before merging?

invidian:
Ping @johananl

pkg/assets/assets_generate.go
```diff
@@ -16,19 +16,36 @@
...
 	"github.com/kinvolk/lokomotive/pkg/assets"
 )
 
 func main() {
-	dirs := map[string]string{
-		assets.TerraformModulesSource: "../../assets/terraform-modules",
-		assets.ControlPlaneSource:     "../../assets/charts/control-plane",
-		assets.ComponentsSource:       "../../assets/charts/components",
+	directoriesToEmbed := map[string]http.FileSystem{
+		assets.TerraformModulesSource: http.Dir("../../assets/terraform-modules"),
```
knrt10, line 32: @invidian can this https://github.com/kinvolk/lokomotive/issues/817 be also done as part of this PR?

invidian: I'd do it separately to reduce the scope of this PR. This is an cleanup PR, not enhancement one.

surajssd (approved):
looks ok

PR #953, opened 2020-09-11
Add a coding style guide

johananl:
This PR adds an initial code style guide for Lokomotive. The guide currently focuses mainly on Go code, however we could add guidelines for other types of code (e.g. HCL, Markdown).

I find it useful to maintain such a guide as I find myself commenting the same things over and over again in PRs and would love to have a document I could link to. I think that following a common and agreed-upon guide could make code reviews more efficient.

The document should be agreed upon by the team and is expected to evolve over time. The current state is by no means a comprehensive list of guidelines. Rather, it currently contains a small set of initial guidelines which I believe already have consensus in the team so we might as well formalize them.

### DIFF
--- a/docs/development/coding-style-guide.md
+++ b/docs/development/coding-style-guide.md
@@ -1,0 +1,280 @@
+# Lokomotive coding style guide
+
+This document describes the coding style that's enforced in the Lokomotive codebase. The guidelines
+in this document should be followed by Lokomotive maintainers and contributors alike. The guide is
+structured in a way which makes it easy to link to a specific guideline, for example from a PR
+comment.
+
+## Guidelines
+
+This section describes the entire set of coding style guidelines which should be followed when
+working on the Lokomotive codebase.
+
+Each "topic" should reside on its own under a heading so that it's easy to unambiguously link to a
+specific guideline during code reviews.
+
+The list of guidelines is meant to evolve over time. If you think a new guideline should be
+enforced, suggest changes to this document in a PR.
+
+### Error string formatting
+
+Error strings are typically *chained*:
+
+```
+Deploying cluster: reading component config: parsing HCL: unexpected character '%'
+```
+
+Remember this fact when constructing error strings and think about how the final string might look
+to the user.
+
+#### Wrapping errors
+
+When "wrapping" an error in another error to provide context, the `%w` formatting verb should be
+used following the [Go 1.13 recommendations](https://blog.golang.org/go1.13-errors) to error
+handling:
+
+`fmt.Errorf("parsing config: %w", err)`
+
+#### Error string text
+
+Don't include words which describe failure when *wrapping* errors. This includes words such as
+"failed", "error", "could not" etc. It's enough to mention a failure in the **beginning** of the
+error string, i.e. when logging the error, and if necessary also at the **end** of the error
+string, i.e. when creating the error value at the deepest level of the call stack. There is no need
+to repeat the fact that something failed in every intermediate "piece" of the error string.
+
+To demonstrate this, consider the following:
+
+```
+Error deploying cluster: error reading component config: failed to parse HCL: unexpected character '%' while parsing HCL
+```
+
+vs.
+
+```
+Error deploying cluster: reading component config: parsing HCL: unexpected character '%'
+```
+
+The 2nd example provides the same information as the 1st one without repetition.
+
+When formatting an error string, provide just the *context*, i.e. **what we were doing** when the
+error occurred (or the erroneous condition if the error is the last one in the "chain").
+
+Good:
+
+- `fmt.Errorf("parsing config: %w", err)`
+- `fmt.Errorf("deploying cluster: %w", err)`
+- `fmt.Errorf("reversing time: %w", err)`
+- `fmt.Errorf("space probe is offline")`
+
+Bad:
+
+- `fmt.Errorf("parsing config failed: %w", err)`
+- `fmt.Errorf("error encountered while deploying cluster: %w", err)`
+- `fmt.Errorf("could not reverse time: %w", err)`
+
+#### Capitalizing error strings
+
+Don't capitalize error strings when *returning* an error (remember - errors are chained).
+Capitalize the **beginning** of an error string when **logging** an error.
+
+Good:
+
+- `fmt.Errorf("parsing config: %w", err)`
+- `log.Errorf("Deploying cluster: %v", err)`
+
+Bad:
+
+- `fmt.Errorf("Parsing config: %w", err)`
+- `log.Errorf("deploying cluster: %v", err)`
+
+### Comments
+
+Follow the standard Go [guidelines](https://blog.golang.org/godoc) for comments in Go code. *In
+addition*, follow the guidelines below.
+
+#### Whitespace
+
+Include a space between `//` and the beginning of the comment.
+
+Good:
+
+```go
+// This is a comment.
+```
+
+Bad:
+
+```go
+//This is a comment.
+```
+
+#### Capitalization
+
+Capitalize the first letter of a Go comment unless the official Go guidelines require you to do
+otherwise.
+
+Good:
+
+```go
+// A very important constant.
+const this = "that"
+```
+
+Bad:
+
+```go
+// a very important constant.
+const this = "that"
+```
+
+#### Periods
+
+As a general rule, end all comments with a period. Possible exceptions are *very* short comments
+(up to 2-3 words which don't form a self-contained sentence) which are typically **inline**,
+however generally you should format comments as full sentences ending with a period.
+
+Good:
+
+```go
+// Foo does this and also that.
+func Foo() {
+    ...
+    n := 3 // Default
+}
+```
+
+Bad:
+
+```go
+// Foo does this and also that
+func Foo() {
+    ...
+}
+```
+
+#### "nolint" comments
+
+"nolint" comments are a special type of comment used to tell the linter to ignore a specific
+linting rule. They are formatted differently from "regular" Go comments.
+
+Format "nolint" comments according to the following example:
+
+```go
+//nolint:funlen,gosec
+func foo() {
+    ...
+}
+```
+
+- **Don't** put a space between `//` and `nolint`.
+- Don't capitalize "nolint" comments.
+- Don't end "nolint" comments with a period.
+- Don't put spaces after commas when multiple linting rules are ignored.
+
+When adding a "nolint" comment for a function which has a doc comment, put the "nolint" line below
+the doc comment after a "separator" comment line:
+
+```go
+// Foo does this and that.
+//
+//nolint:funlen,gosec
+func Foo() {
+    ...
+}
+```
+
+### Table-driven tests
+
+Whenever possible, use [table-driven tests](https://github.com/golang/go/wiki/TableDrivenTests).
+Table-driven tests are compact, readable and easy to maintain compared to test cases that are
+described imperatively.
+
+It is usually possible to construct most tests in a way which allows the various test cases to be
+described *declaratively* in a Go struct. Most tests can be broken down to "steps" similar to the
+following:
+
+1. Set some initial state.
+1. Call the function we're testing with specific arguments.
+1. Ensure the return value matches a desired result.
+
+Each step in the list above can correspond to a field in the Go struct which describes the test
+cases. There are exceptions to this, but in general try to group similar test cases in one test
+function. You can then share the test execution logic for all test cases as in the example below.
+
+Test cases which require a different *execution logic* (e.g. a different sequence of function
+calls) should reside in separate test functions.
+
+**Example**
+
+```go
+func add(a, b int) int {
+	return a + b
+}
+
+func TestAdd(t *testing.T) {
+	tests := []struct {
+		desc string
+		a    int
+		b    int
+		want int
+	}{
+		{
+			desc: "Positive numbers",
+			a:    1,
+			b:    3,
+			want: 4,
+		},
+		{
+			desc: "Negative and positive numbers",
+			a:    1,
+			b:    -3,
+			want: -2,
+		},
+		{
+			desc: "Zero and a positive number",
+			a:    0,
+			b:    5,
+			want: 5,
+		},
+		{
+			desc: "Failing test",
+			a:    2,
+			b:    2,
+			want: 5,
+		},
+	}
+
+	for _, test := range tests {
+		// Pin local var to avoid scope issues.
+		test := test
+
+		t.Run(test.desc, func(t *testing.T) {
+			sum := add(test.a, test.b)
+
+			if sum != test.want {
+				t.Fatalf("Unexpected result: got %d, want %d", sum, test.want)
+			}
+		})
+	}
+}
+```
+
+Using a structure as in the example above provides an easy-to-read output:
+
+```
+=== RUN   TestAdd
+=== RUN   TestAdd/Positive_numbers
+=== RUN   TestAdd/Negative_and_positive_numbers
+=== RUN   TestAdd/Zero_and_a_positive_number
+=== RUN   TestAdd/Failing_test
+    main_test.go:46: Unexpected result: got 4, want 5
+--- FAIL: TestAdd (0.00s)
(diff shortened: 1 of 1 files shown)


invidian:
@johananl did you see #883?

docs/development/coding-style-guide.md
```diff
@@ -0,0 +1,246 @@
...
+
+```go
+// a very important constant.
+const this = "that"
+```
+
+#### Periods
+
+As a general rule, end all comments with a period. Possible exceptions are *very* short comments
+(up to 2-3 words which don't form a self-contained sentence) which are typically **inline**.
```
invidian, line 116: I would suggest to make all comments a statements. It is not difficult and then you can always use periods.

docs/development/coding-style-guide.md
```diff
@@ -0,0 +1,246 @@
...
+#### "nolint" comments
+
+"nolint" comments are a special type of comment used to tell the linter to ignore a specific
+linting rule. They are formatted differently from "regular" Go comments.
+
+Format "nolint" comments according to the following example:
+
+```go
+//nolint:funlen,gosec
+func foo() {
```
invidian, line 146: As mentioned in #883, do we prefer nolint on top of the function, rather than inline with function, like `func foo() { //nolint:funlen` ?

docs/development/coding-style-guide.md
```diff
@@ -0,0 +1,246 @@
...
+    ...
+}
+```
+
+- **Don't** put a space between `//` and `nolint`.
+- Don't capitalize "nolint" comments.
+- Don't end "nolint" comments with a period.
+- Don't put spaces after commas when multiple linting rules are ignored.
+
+### Table-driven tests
```
invidian, line 156: Should we mention here, that table-driven tests should share logic at the end of the test? Otherwise it doesn't make much sense to use table-driven approach.

invidian (commented):
Great start @johananl!

johananl:
> @johananl did you see #883?

Nope, thanks for the pointer. I'm deliberately avoiding adding *all* the guidelines we know about from the start. Rather, I wanted to start by merging something simple and encourage the team to add more guidelines as the need arises.

docs/development/coding-style-guide.md
johananl: What do you mean by "statements"? Can you provide an example?

docs/development/coding-style-guide.md
johananl: I find the inline version a bit strange, but no strong opinion here.

docs/development/coding-style-guide.md
johananl: I thought this was clear from the example. I can make it more explicit though.

docs/development/coding-style-guide.md
invidian: Oh, I meant "sentences" :smile: sorry for the confusion!

docs/development/coding-style-guide.md
invidian: Yeah, I think that would be useful. Thanks :)

johananl: Done.

docs/development/coding-style-guide.md
invidian: https://golangci-lint.run/usage/false-positives/ suggests on top, so I suggest we should point here that having it inline is a bad example:
```
0 ✓ (134ms) 16:40:28 invidian@dellxps15mateusz ~/repos/kinvolk/lokomotive (invidian/more-cleanup)$ git grep -A1 nolint | grep func  | wc -l
26
0 ✓ (78.8ms) 16:40:30 invidian@dellxps15mateusz ~/repos/kinvolk/lokomotive (invidian/more-cleanup)$ git grep nolint | grep func | wc -l
7
```

I also think how does it play with function comment, like:
```go
// Foo does bar.
//
//nolint:funlen
func Foo() {
```

vs

```go
// Foo does bar
func Foo() { //nolint:funlen
```

I _think_ first case is still slightly better (less packed, so more readable).

docs/development/coding-style-guide.md
johananl: What constitutes a sentence? For example, are the following sentences?

- `Deploy cluster.`
- `Default value.`
- `Workaround - see <issue>.`

Sometimes a "full" sentence (i.e. a sentence which includes a subject, and object and a verb) can be superfluous in code comments.

docs/development/coding-style-guide.md
johananl: I also prefer the first example.

johananl: How is it now (just pushed)?

docs/development/coding-style-guide.md
invidian: Hmm. So maybe we can keep the exception about short comments but encourage to phrase comments as sentences when possible?

For example:
`Workaround - see <issue>.` -> `Workaround for foo, see <issue> for more details.`

johananl: SGTM.

docs/development/coding-style-guide.md
invidian: Looks great, thanks!

invidian (approved):
LGTM :heavy_check_mark:

iaguis (dismissed):
Thanks! LGTM.

surajssd (approved):
LGTM

knrt10 (approved):
Thank you for adding this. +1

docs/development/coding-style-guide.md
```diff
@@ -0,0 +1,262 @@
...
+Don't include words which describe failure in error strings. This includes words such as "failed",
+"error", "could not" etc. It is clear that an error indicates a failure. Instead, provide just the
+*context*, i.e. **what we were doing** when the error occurred (or the erroneous condition if the
+error is the last one in the "chain").
+
+Good:
+
+- `fmt.Errorf("parsing config: %w", err)`
+- `fmt.Errorf("deploying cluster: %w", err)`
+- `fmt.Errorf("reversing time: %w", err)`
```
rata, line 49: I don't really know what that is. In fact, the "bad" alternative was more useful for me as at least there I'm slightly more confident that to "reverse time" is a thing.

Maybe this fact signals that error messages like this are a little cryptic (at least if you are not familiar with the concepts, as I guess several users won't be familiar with what HCL is)?

docs/development/coding-style-guide.md
```diff
@@ -0,0 +1,262 @@
...
+```
+
+Bad:
+
+```go
+// a very important constant.
+const this = "that"
+```
+
+#### Periods
```
rata, line 113: I find a title like this more clear.

```suggestion
#### End comments with periods
```

I understand this might not be consistent with the rest, but titles start to get cryptic from my POV. I think it is useful to look at a glance and understand where to check something. A clearer title helps, IMHO.

docs/development/coding-style-guide.md
```diff
@@ -0,0 +1,262 @@
...
+- `fmt.Errorf("parsing config: %w", err)`
+- `fmt.Errorf("deploying cluster: %w", err)`
+- `fmt.Errorf("reversing time: %w", err)`
+- `fmt.Errorf("space probe is offline")`
+
+Bad:
+
+- `fmt.Errorf("parsing config failed: %w", err)`
+- `fmt.Errorf("error encountered while deploying cluster: %w", err)`
+- `fmt.Errorf("could not reverse time: %w", err)`
```
rata, line 56: I have mixed feeling with this.

What is the reasoning to prefer those over these? I mean, I understand what was mentioned here: https://github.com/kinvolk/lokomotive/pull/953/files#diff-a4a1697930a32405cd5efdc7aa09b71fR40-R43. But I don't see any real reason to prefer that and when I look at the examples it doesn't seem clearly better.

For example, this line: 
```
 Deploying cluster: reading component config: parsing HCL: unexpected character '%'
```

Seems it will be more clear if it is shown like this:

```
 Deploying cluster: reading component config: parsing HCL config failed: unexpected character '%'
```

If the error clearly says what failed (e.g. "parsing HCL config failed: <err>"), IMHO seems easier to understand rather if it just says what it was doing (e.g. "parsing HCL confing: <err>").

But as I said, I have mixed feeling with this. Sometimes I like the brevity, so I'm unsure.

If no one else has concerns with this, just mark this comment as solved :)

docs/development/coding-style-guide.md
```diff
@@ -0,0 +1,262 @@
...
+
+### Table-driven tests
+
+Whenever possible, use [table-driven tests](https://github.com/golang/go/wiki/TableDrivenTests).
+Table-driven tests are compact, readable and easy to maintain compared to test cases that are
+described imperatively.
+
+It is usually possible to construct most tests in a way which allows the various test cases to be
+described *declaratively* in a Go struct. Most tests can be broken down to "steps" similar to the
+following:
```
rata, line 177: Is this paragraph and all that comes after really needed?

docs/development/coding-style-guide.md
```diff
@@ -0,0 +1,262 @@
...
+This section describes the entire set of coding style guidelines which should be followed when
+working on the Lokomotive codebase.
+
+Each "topic" should reside on its own under a heading so that it's easy to unambiguously link to a
+specific guideline during code reviews.
+
+The list of guidelines is meant to evolve over time. If you think a new guideline should be
+enforced, suggest changes to this document in a PR.
+
+### Error string formatting
```
ipochi, line 19: I'd also add about not using `` , rather use `%q` directive.

ipochi (approved):
minor suggestion but LGTM.

Great start @johananl

rata (commented):
Mostly LGTM.

The only thing I'm unsure about is the errors. I'm unsure if errors won't become cryptic or if we will simply have ton of review comments on error fromatting without much benefit.

Is there any other project using that style rule for errors? I guess from the blog post example maybe the go project itself?

docs/development/coding-style-guide.md
invidian: @rata where is HCL mentioned here?

I think the main motivation for removing `failed` and similar keywords is to avoid cluttering the code with them, as you would need to include them in every single error message. And `fmt.Errorf` already implies that something has failed.

Perhaps to provide some extra context here, it would be good to have an example how chained errors looks like later on. e.g.
`Applying cluster failed: parsing configuration: illegal character at line 8`.
vs
`Applying cluster failed: parsing configuration failed: failed to parse illegal character at line 8`.

Maybe it would make more sense to do something like:
`Applying cluster failed. Cause: parsing configuration: illegal character at line 8`.

BTW, now that I think about it, perhaps we should also write something down, that those error traces should be uniquely identifiable, so ideally single function shouldn't return the same error in 5 different places, as then it makes it difficult to trace what exactly happened.

docs/development/coding-style-guide.md
johananl: Can add that, although `%q` formats using *double quotes* and sometimes people want single quotes.

johananl:
>The only thing I'm unsure about is the errors. I'm unsure if errors won't become cryptic or if we will simply have ton of review comments on error fromatting without much benefit.

Not sure I got your point. What change are you suggesting? Removing the guidelines about errors?

>Is there any other project using that style rule for errors? I guess from the blog post example maybe the go project itself?

There is https://github.com/golang/go/wiki/CodeReviewComments#error-strings if you're looking for an official recommendation. But it also just makes sense IMO because error strings are chained. These guidelines produce consistently-formatted, readable errors.

docs/development/coding-style-guide.md
invidian: I think we can add that later, to describe that if you mention a context of error, `%q` should be used or something. Current PR is not opinionated on that as far as I read.

docs/development/coding-style-guide.md
johananl: @rata "reversing time" was included as an example only. I'd be extremely careful with adding actual time reversal functionality to Lokomotive, and in any case this should be done in a major release.

The intention behind the error formatting guidelines is to avoid having things like the following:

`Deploying cluster: Could not deploy cluster: Failed to deploy cluster due to an error: Error while running Terraform: Terraform execution failed: Error - no such file or directory`

rata: @invidian I don't think that makes much sense. It's a log, not a core dump :-D

@johananl Ohhh, I get the time reverse now! :joy:. I got the intention to avoid those kind of errors, was unsure if the whole error was better one way or the other. But I guess this seems fine. Thanks :)

docs/development/coding-style-guide.md
rata: Nit-picking, I think using infinitive might help. So, instead of `parsing config` just `parse config`, etc. In fact, in the blog and quickly checking go source code, seems that past and infinitive are quite commonly used.

But maybe this is being too restrictive on how errors should look like with no benefit. As you prefer.

As I said in the previous comment, if no one else has concerns with this, just mark this as solved :)

rata:
> Not sure I got your point. What change are you suggesting? Removing the guidelines about errors?

Yes, that can be simpler. But no problem, we can remove it later if we find that it creates overhead and little benefit :)

johananl:
> > Not sure I got your point. What change are you suggesting? Removing the guidelines about errors?
> 
> Yes, that can be simpler. But no problem, we can remove it later if we find that it creates overhead and little benefit :)

I've added only things which I've found myself constantly commenting on in PRs. That was actually the trigger for me starting this document. IMO *not* having guidelines for error strings is what's causing overhead.

docs/development/coding-style-guide.md
johananl: Yes, I think @ipochi wanted us to *add* a guideline for `%q`.

docs/development/coding-style-guide.md
johananl: I've made it clearer why we shouldn't "clutter" error strings with "failed", "couldn't" etc.

docs/development/coding-style-guide.md
johananl: Remember this heading is under `### Comments`, i.e. it's about periods in comments specifically. I wouldn't use the imperative tense ("do this and that"), but I can change it to "Periods at ends of comments" if you think it's better.

docs/development/coding-style-guide.md
johananl: My logic here is the following:

An error indicates to the user that something failed. More precisely, **one specific thing** failed, and caused a chain of dependent things to fail as a result. For example, we couldn't read a file, and therefore we couldn't construct a config, and therefore we couldn't deploy a cluster.

Given the above, including "failed, "couldn't" etc. in *every* piece of the chain is redundant. Consider the following two cases:

```
Error deploying cluster: error reading component config: failed to parse HCL: unexpected character '%' while parsing HCL
```

vs.

```
Error deploying cluster: reading component config: parsing HCL: unexpected character '%'
```

In my opinion there is no need, for example, to say `error reading component config`. It is very clear that we didn't manage to read the config, because if we did there wouldn't be an error...

The pieces of an error string provide **context** to an error as sort of a condensed "story". We don't need to repeat "error", "failed", "this and that broke" etc. to provide context.

docs/development/coding-style-guide.md
johananl: My intention here was to emphasize that we should try to describe tests declaratively using the table-driven tests approach. Which part do you think could be unnecessary?

invidian (approved):

docs/development/coding-style-guide.md
johananl: I'm resolving this for the sake of making progress. @ipochi please feel free to add/modify things in a new PR.

docs/development/coding-style-guide.md
johananl: I'm resolving this as I would like to get things merged. @rata please feel free to suggest changes in a new PR - this is a living document.

johananl:
Thanks for everybody who chimed in. I will merge this since we don't seem to have any major disagreements around the content.
Please feel free to propose new additions or modify existing guidelines in new PRs.

PR #966, opened 2020-09-15
Makefile: complain if assets dir has empty directories

invidian:
As running locally 'make update-assets' will include them, but as git do
not include empty directories, after fresh clone it will be removed, so
the assets won't be consistent.

This find will ensure, that if there are empty directories found,
update-assets will fail with clear message to the developer.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,6 +45,8 @@ all: build build-test test lint codespell
 
 .PHONY: update-assets
 update-assets:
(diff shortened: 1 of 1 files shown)


knrt10:
Should we check for empty files too? Because we have 5

```bash
find assets/ -type f -empty | wc -l                                                                                                        
       5
```

invidian:
>Should we check for empty files too? Because we have 5

Empty files are fine, as they show up in `git status` still. We also have them, as otherwise `helm lint` complains, that `values.yaml` file is not present.

knrt10 (approved):
+1

PR #990, opened 2020-09-18
components/metallb: Add missing autodiscovery labels

### LINKED ISSUE #987: components/metallb: Add all supported labels to autodiscovery config
Right now we configure the MetalLB autodiscovery to look just for the labels required for establishing a BGP session: local ASN, remote ASN and peer address.

https://github.com/kinvolk/lokomotive/blob/5ba9d5cd3bf38854c38f538a1cac97bb1214b982/pkg/components/metallb/manifests.go#L516-L518

In order to allow customizing all [supported](https://github.com/metallb/metallb/blob/85b7a46ab5ca3a47ab15b6eb5497a8ac01ffe7d5/internal/config/config.go#L75-L81) BGP parameters, we should specify labels for them in the autodiscovery config following the current convention. It is then up to the user to decide whether to apply a specific label to a worker pool.

Once https://github.com/metallb/metallb/pull/593 is merged we may want to look into exposing the autodiscovery config to the user in the component config.

johananl:
In order to allow users to control all supported BGP parameters by
specifying labels in worker pool config, we need to specify mappings
for them.

Fixes #987.

To test this change, deploy a cluster using a config similar to the following:

```hcl
cluster "packet" {
  asset_dir        = "./assets"
  cluster_name     = "my-cluster"
  controller_count = 1
  controller_type  = "t1.small.x86"
  dns {
    provider = "route53"
    zone     = "example.com"
  }
  facility    = "ams1"
  project_id = "fake"
  ssh_pubkeys       = ["ssh-rsa ..."]
  management_cidrs  = ["0.0.0.0/0"]
  node_private_cidr = "10.1.1.0/25"

  worker_pool "pool-1" {
    count     = 2
    node_type = "t1.small.x86"
    # Example: configure hold time and peer port by setting autodiscovery labels
    labels    = "metallb.lokomotive.io/hold-time=3s,metallb.lokomotive.io/peer-port=1179"
  }
}

component "metallb" {
  address_pools = {
    default = ["147.1.2.3/32"]
  }
}

```

### DIFF
--- a/pkg/components/metallb/component_test.go
+++ b/pkg/components/metallb/component_test.go
@@ -265,6 +265,10 @@ component "metallb" {
     my-asn: metallb.lokomotive.io/my-asn
     peer-asn: metallb.lokomotive.io/peer-asn
     peer-address: metallb.lokomotive.io/peer-address
+    peer-port: metallb.lokomotive.io/peer-port
+    src-address: metallb.lokomotive.io/src-address
+    hold-time: metallb.lokomotive.io/hold-time
+    router-id: metallb.lokomotive.io/router-id
 address-pools:
 - name: default
   protocol: bgp
--- a/pkg/components/metallb/manifests.go
+++ b/pkg/components/metallb/manifests.go
@@ -516,6 +516,10 @@ data:
         my-asn: metallb.lokomotive.io/my-asn
         peer-asn: metallb.lokomotive.io/peer-asn
         peer-address: metallb.lokomotive.io/peer-address
+        peer-port: metallb.lokomotive.io/peer-port
+        src-address: metallb.lokomotive.io/src-address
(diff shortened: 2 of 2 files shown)


invidian (approved):
LGTM

iaguis (approved):
lgtm

surajssd (changes requested):
Please add an e2e test to verify these labels are taking effect?

johananl:
> Please add an e2e test to verify these labels are taking effect?

Since we're already verifying the component renders these labels in a unit test, what would be checking by adding an e2e test for this? AFAICT such a test would verify that the *API server* actually writes the ConfigMap we provide. Do you see value in that?

surajssd (approved):
Ah I didn't realise that the unit test was updated as well. Cool LGTM. Sorry for the noise.

PR #996, opened 2020-09-22
docs/metallb: Clarify address_pools knob

johananl:
This PR clarifies the proper usage of the `address_pools` knob of the `metallb` component.

### DIFF
--- a/docs/configuration-reference/components/metallb.md
+++ b/docs/configuration-reference/components/metallb.md
@@ -50,8 +50,8 @@ MetalLB component configuration example:
 ```tf
 component "metallb" {
   address_pools = {
-    default = ["147.63.8.20/32"]
-    special_addresses = ["147.85.47.16/29", "147.85.47.24/29"]
+    pool1 = ["147.63.8.20/32"]
+    pool2 = ["147.85.47.16/29", "147.85.47.24/29"]
   }
   controller_node_selectors = {
     "kubernetes.io/hostname" = "worker3"
@@ -85,7 +85,7 @@ Example:
 
 | Argument                    | Description                                                                                | Default | Type                                                                                                           | Required |
 |-----------------------------|--------------------------------------------------------------------------------------------|:-------:|:---------------------------------------------------------------------------------------------------------------|:--------:|
-| `address_pools`             | A map which allows specifying one or more CIDRs which MetalLB can use to expose services.  |    -    | object({default = list(string), special_addresses = list(string)})                                             |   true   |
+| `address_pools`             | A map which allows specifying one or more CIDRs which MetalLB can use to expose services.  |    -    | map(list(string))                                                                                              |   true   |
 | `controller_node_selectors` | A map with specific labels to run MetalLB controller pods selectively on a group of nodes. |    -    | map(string)                                                                                                    |  false   |
 | `speaker_node_selectors`    | A map with specific labels to run MetalLB speaker pods selectively on a group of nodes.    |    -    | map(string)                                                                                                    |  false   |
 | `controller_toleration`     | Specify one or more tolerations for controller pods.                                       |    -    | list(object({key = string, effect = string, operator = string, value = string, toleration_seconds = string })) |  false   |
--- a/docs/configuration-reference/components/metallb.md
+++ b/docs/configuration-reference/components/metallb.md
@@ -100,6 +100,7 @@ To apply the MetalLB component:
 ```bash
 lokoctl component apply metallb
 ```
+
 ## Deleting
 
 To destroy the component:
--- a/docs/configuration-reference/components/metallb.md
+++ b/docs/configuration-reference/components/metallb.md
@@ -77,6 +77,28 @@ component "metallb" {
 
 MetalLB will use the specified CIDR for exposing services of type `LoadBalancer`.
 
+### Advanced IP allocation
+
+By default, MetalLB uses all specified address pools to allocate IP addresses to services. To
+request an address from a specific pool, set the `metallb.universe.tf/address-pool` annotation for
+the relevant service:
+
+```yaml
+apiVersion: v1
+kind: Service
+metadata:
+  name: nginx
+  annotations:
+    metallb.universe.tf/address-pool: pool2
+spec:
+  ports:
+  - port: 80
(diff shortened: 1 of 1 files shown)


knrt10 (approved):
Thanks for the PR. +1

docs/configuration-reference/components/metallb.md
```diff
@@ -50,8 +50,8 @@ MetalLB component configuration example:
 ```tf
 component "metallb" {
   address_pools = {
-    default = ["147.63.8.20/32"]
-    special_addresses = ["147.85.47.16/29", "147.85.47.24/29"]
+    pool1 = ["147.63.8.20/32"]
```
invidian, line 53: How can you choose which pool will be used? Maybe we could put some link in the description?

johananl: This is covered by the official MetalLB docs. TL;DR: you specify an annotation on the relevant service. I've included a "for more info" note with a link to the official docs.

invidian: Hm, as a person who is not very familiar with MetalLB, having some more specific links and description of what this flag does would be useful.

Could we link to https://metallb.universe.tf/usage/example/? Or maybe give an example of setting this annotation to some component service, if that's currently supported?

ipochi: +1 on @invidian's comment about linking an example.

But truth be told, couple of lines of addition in our documentation would make much more sense that merely linking the example from metallb docs.


Example: Annotate the Service with the name of the address pool to obtain an IP address within the specified range:
```
metallb.universe.tf/address-pool: pool1
```

docs/configuration-reference/components/metallb.md
```diff
@@ -77,6 +77,9 @@ component "metallb" {
 
 MetalLB will use the specified CIDR for exposing services of type `LoadBalancer`.
 
+>NOTE: The official MetalLB [documentation](https://metallb.universe.tf/) contains information
```
ipochi, line 80: This links to the main website. Is that the intention ? or to a specific page ?

docs/configuration-reference/components/metallb.md
johananl: The question is - if we start including examples for MetalLB **usage**, why document just this case? There are many more params which can be tweaked and we would be duplicating the MetalLB docs.

invidian: We should document this case, because `metallb.universe.tf/address-pool` Service labels is directly related to `address_pools` feature and key field in it IMO.

docs/configuration-reference/components/metallb.md
johananl: I've deliberately linked to the main page. What other page did you have in mind?

docs/configuration-reference/components/metallb.md
johananl: But there is more info regarding address pools. The MetalLB docs explain how allocation works, how to disable auto-allocation etc. Any info we include there would be duplicating the MetalLB docs.

invidian: I'm fine with some more specific link to MetalLB documentation too. We don't necessarily need to duplicate the content here. But linking to https://metallb.universe.tf/ is not very helpful, as the documentation is broad.

ipochi: We can do a TL;DR version in our documentation and link to metallb docs for more comprehensive documentation.

johananl: My point is - if we do what you're proposing @invidian @ipochi, why do it just for pool selection? What's special about this feature? There is a bunch of MetalLB functionality which isn't covered in the Lokomotive docs.

invidian: All other knobs we expose for MetalLB are generic to any Kubernetes workload (selector and tolerations). Only  `address_pools` is MetalLB specific, hence the request to have a better documentation for it.

docs/configuration-reference/components/metallb.md
ipochi: I meant if the link to the main website was deliberate or accidental. If deliberate, no issues :+1:

docs/configuration-reference/components/metallb.md
johananl: Fine, I'll add some info.

johananl: Done.

johananl:
I've addressed the feedback.

invidian (dismissed):
LGTM, thanks @johananl

docs/configuration-reference/components/metallb.md
```diff
@@ -77,6 +77,46 @@ component "metallb" {
...
+    targetPort: 80
+  selector:
+    app: nginx
+  type: LoadBalancer
+```
+
+To exempt a pool from automatic IP allocation, set `auto-assign` on the pool to `false`:
+
+```yaml
+address-pools:
```
iaguis, line 105: Where is this located? In the MetalLB config map? In that case we should mention where this should go.

invidian: ping @johananl

johananl: > Where is this located? In the MetalLB config map? In that case we should mention where this should go.

I assumed it was obvious. I guess it isn't :-) Fixing.

johananl: Actually, wait. We aren't exposing any knobs to control `auto-assign`.

I've added info about selecting which pool to use following this: https://github.com/kinvolk/lokomotive/pull/996#discussion_r492876707

I guess it doesn't make sense to mention disabling auto assignment as long as we don't expose this functionality.

johananl: I've removed the section about `auto-assign`. We can discuss adding support for it in a new issue if there is demand for it (but there is a bunch of knobs we don't support at the moment).

invidian (approved):
LGTM, thanks @johananl

knrt10 (approved):
+1. Nice work.

iaguis (approved):

PR #1010, opened 2020-09-24
packet: Read BGP peer address from metadata service

### LINKED ISSUE #1009: components/metallb: Ensure correct discovery of BGP peer address
In some Packet facilities the BGP address is the node's gateway address, whereas in others it is a **loopback** address that's reachable via the gateway. Right now the latter case seems to break our MetalLB component.

Right now we do the following to discover the peer address:

https://github.com/kinvolk/lokomotive/blob/3008e5a1cd7cc478482a7c2b365e8a02ae25fc95/assets/terraform-modules/packet/flatcar-linux/kubernetes/workers/cl/worker.yaml.tmpl#L90

Packet exposes the correct BGP addresses to use generically on their API as well as the metadata service:

```
curl https://api.packet.net/devices/${INSTANCE_UUID}/bgp/neighbors

curl https://api.packet.net/devices/${INSTANCE_UUID} | jq -r .bgp_neighbors[0]

curl https://metadata.packet.net/metadata | jq -r .bgp_neighbors[0]
```

We can likely use the above instead of reading the gateway from the routing table.

johananl:
In some Packet facilities the BGP peer address isn't the same as the gateway address allocated for a host. Rather, it is a loopback address that's reachable via the gateway.

The Packet metadata service now exposes BGP info to hosts, so we can query the metadata service for the BGP peer address.

The source address is explicitly specified since when the peer address is a loopback address, the source IP addresses which ends up getting selected by the kernel is the node's *public* address which doesn't work. In cases where the peer address is the gateway address there is no harm in explicitly specifying the source.

Fixes #1009.

### DIFF
--- a/assets/terraform-modules/packet/flatcar-linux/kubernetes/workers/cl/worker.yaml.tmpl
+++ b/assets/terraform-modules/packet/flatcar-linux/kubernetes/workers/cl/worker.yaml.tmpl
@@ -34,6 +34,20 @@ systemd:
         ExecStart=/bin/sh -c 'while ! /usr/bin/grep '^[^#[:space:]]' /etc/resolv.conf > /dev/null; do sleep 1; done; /opt/wait-for-dns ${dns_zone} ${cluster_name}-private 3600'
         [Install]
         RequiredBy=kubelet.service
+    %{~ if bgp_node_labels != "" ~}
+    - name: bgp-metadata.service
+      enable: true
+      contents: |
+        [Unit]
+        Description=Write BGP metadata to disk
+        Before=kubelet.service
+        [Service]
+        Type=oneshot
+        RemainAfterExit=true
+        ExecStart=/opt/bgp-metadata
+        [Install]
+        RequiredBy=kubelet.service
+    %{~ endif ~}
     - name: coreos-metadata.service
       enable: true
       contents: |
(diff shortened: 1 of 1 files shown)


invidian (dismissed):
LGTM, I wonder if we should also add tests for it, so we can detect some issues in the future. Thanks for quick reaction @johananl !

johananl:
> LGTM, I wonder if we should also add tests for it, so we can detect some issues in the future. Thanks for quick reaction @johananl !

Ideally such a test would be a "blackbox" ingress test (i.e. sending traffic via MetalLB). *However*, there doesn't seem to be a trivial way to handle EIP allocation so we would have to build something for it. Additionally, such a test wouldn't have caught this specific issue because only specific facilities were affected.

johananl:
There is a race condition with enabling BGP.

assets/terraform-modules/packet/flatcar-linux/kubernetes/workers/cl/worker.yaml.tmpl
```diff
@@ -82,12 +82,9 @@ systemd:
         ExecStartPre=/usr/bin/bash -c "grep 'certificate-authority-data' /etc/kubernetes/kubeconfig | awk '{print $2}' | base64 -d > /etc/kubernetes/ca.crt"
         ExecStartPre=-/usr/bin/rkt rm --uuid-file=/var/cache/kubelet-pod.uuid
         ExecStartPre=/etc/kubernetes/configure-kubelet-cgroup-driver
-        # TODO: Workaround until https://github.com/coreos/afterburn/pull/358
-        # makes it into Flatcar. Then we can read COREOS_PACKET_IPV4_PRIVATE_GATEWAY_0
-        # from /run/metadata/flatcar and disable the template conditionals below.
         ExecStart=/bin/sh -c \
             "%{~ if bgp_node_labels != "" ~}
-            BGP_PEER_ADDRESS=$(ip route | grep '10.0.0.0/8' | awk {'print $3'}); \
+            BGP_PEER_ADDRESS=$(curl -s https://metadata.packet.net/metadata | jq -r .bgp_neighbors[0].peer_ips[0]); \
```
rata, line 87: What if curl time out or hangs for a very long time? I think the content of the variable might not be a valid peer and we probably should fail?

I think adding these parameters might help (is not nice, though):  `curl --retry-delay 1 --retry 60 --retry-connrefused --retry-max-time 60 --connect-timeout 20`

(Taken from https://github.com/kinvolk/vpn/blob/923a75d6713690e811fb003a1bf01a9e3398f307/bird/startup.sh#L9, this repo is not public for other seeing it. Sorry :()

That will only help to not hang, but we still need to handle the failure scenario.

rata, line 87: Why the first one of the peers?

rata (approved):
Added two comments about failure scenarios, but if this is needed ASAP we can handle those in a following PR. As you consider it is best.

assets/terraform-modules/packet/flatcar-linux/kubernetes/workers/cl/worker.yaml.tmpl
johananl: Good point. We're facing another problem though: sometimes `curl` runs before the BGP session got created by Terraform, in which case the value becomes `null`. I'm trying to avoid having to wrap this in a retry script.

johananl: Because MetalLB doesn't support multiple node peers. I've updated the commit message.

johananl: I've added a "wait for" script which handles retries. I've added `--connect-timeout` though.

johananl: Actually we may want some flags in the actual command which gets the data. Fixing.

johananl:
> There is a race condition with enabling BGP.

Fixed using a retry script.

invidian:
>Additionally, such a test wouldn't have caught this specific issue because only specific facilities were affected.

We randomize the locations to get best machines availability, so it would occur in some cases at least.

>Ideally such a test would be a "blackbox" ingress test (i.e. sending traffic via MetalLB). However, there doesn't seem to be a trivial way to handle EIP allocation (in combination with multiple parallel CI runs) so we would have to build something for it. 

Right... Given that it is not trivial to test, we should merge this to fix issue and perhaps raise the priority of testing it.

assets/terraform-modules/packet/flatcar-linux/kubernetes/workers/cl/worker.yaml.tmpl
```diff
@@ -421,6 +434,35 @@ storage:
...
+      mode: 0544
+      contents:
+        inline: |
+          #!/bin/bash
+          set -o pipefail
+          max_attempts=3600
+          echo "Polling metadata service for BGP metadata"
+          counter=0
+          while [[ $counter -lt $max_attempts ]]; do
+              out=$(curl -s -f --connect-timeout 5 \
```
invidian, line 448: What if BGP is not enabled for some reason on the host? Are we able to differentiate somehow situation when it's not enabled completely from not _yet_ enabled?

invidian (approved):
Just one thought, otherwise LGTM

assets/terraform-modules/packet/flatcar-linux/kubernetes/workers/cl/worker.yaml.tmpl
johananl: Oh, good catch. We do have conditionals in the kubelet template for when BGP is disabled, however we must also disable this script, otherwise kubelet will never start.

johananl:
Marking as "do not merge" until https://github.com/kinvolk/lokomotive/pull/1010#discussion_r494599881 is resolved.

assets/terraform-modules/packet/flatcar-linux/kubernetes/workers/cl/worker.yaml.tmpl
invidian: Ah, hehe. I meant more like a bug in Terraform code. I forgot that we actually have an option to disable BGP for workers :)

rata: In case it helps, it is here: https://github.com/kinvolk/lokomotive/blob/master/pkg/platform/packet/template.go#L207

assets/terraform-modules/packet/flatcar-linux/kubernetes/workers/cl/worker.yaml.tmpl
rata: Oh, right, I forgot about that :). Can you create an issue to do this in the future, then? So we don't forget :)

assets/terraform-modules/packet/flatcar-linux/kubernetes/workers/cl/worker.yaml.tmpl
```diff
@@ -421,6 +434,35 @@ storage:
...
+                  https://metadata.packet.net/metadata | jq -r .bgp_neighbors[0].peer_ips[0])
+              ret=$?
+              if [[ $ret -ne 0 ]]; then
+                  echo "Non-zero exit code: $ret"
+              elif [[ "$out" = "" ]]; then
+                  echo "Empty response"
+              elif [[ "$out" = "null" ]]; then
+                  echo "Null response"
+              else
+                  echo "BGP metadata is available!"
```
rata, line 458: Wouldn't it be simpler to write the metadata info to an env file, so that the kubelet just uses that (like Enviromnent=<file created>) instead of repeating the curl?

Furthermore, the following curl can return an error, for example, and in that case we are not handling it correctly

invidian: I guess it make sense, but we need to write it to some tmpfs, so it gets removed on boot, right?

rata: No, if the file is removed the kubelet service won't work after node restart. We want to preserve that file.

invidian: Yeah, but on next boot it will be freshly recreated. Otherwise kubelet will start without waiting for file, so we may get stale data.

assets/terraform-modules/packet/flatcar-linux/kubernetes/workers/cl/worker.yaml.tmpl
johananl: https://github.com/kinvolk/lokomotive/issues/1014

johananl:
@rata @invidian do you guys think the deployment should fail if BGP is enabled bug we couldn't get the peer address? In other words, do we let the kubelet start without the BGP-related labels? Doing so would mean the user will get a working cluster but won't be able to use MetalLB.

invidian:
>@rata @invidian do you guys think the deployment should fail if BGP is enabled but we couldn't get the peer address? In other words, do we let the kubelet start without the BGP-related labels? Doing so would mean the user will get a working cluster but won't be able to use MetalLB.

I think it should fail, yes. I guess that would assume mis-configured project which has BGP disabled?

assets/terraform-modules/packet/flatcar-linux/kubernetes/workers/cl/worker.yaml.tmpl
johananl: >Otherwise kubelet will start without waiting for file, so we may get stale data.

I don't think there is ever a case where a node's BGP address changes.

johananl:
> > @rata @invidian do you guys think the deployment should fail if BGP is enabled but we couldn't get the peer address? In other words, do we let the kubelet start without the BGP-related labels? Doing so would mean the user will get a working cluster but won't be able to use MetalLB.
> 
> I think it should fail, yes. I guess that would assume mis-configured project which has BGP disabled?

For example. Could also be other things we haven't foreseen. The current behavior is to fail. I agree it's better to fail loudly than quietly here.

assets/terraform-modules/packet/flatcar-linux/kubernetes/workers/cl/worker.yaml.tmpl
johananl: Added a conditional around enabling `wait-for-bgp.service`.

assets/terraform-modules/packet/flatcar-linux/kubernetes/workers/cl/worker.yaml.tmpl
johananl: Yes, I've considered that but decided to go with something simpler since the whole thing is a workaround which should hopefully be gone soon. It *is* safer though to go with the file approach. I'll refactor.

johananl: Done.

johananl:
I've refactored the script to poll for BGP metadata and write the peer address to `/run/metadata/bgp`. The kubelet unit now uses this file as an environment file when BGP is enabled.

I've tested a manual deployment with BGP both enabled and disabled and this seems to work well.

assets/terraform-modules/packet/flatcar-linux/kubernetes/workers/cl/worker.yaml.tmpl
```diff
@@ -421,6 +431,41 @@ storage:
...
+                  if [[ ! "$out" =~ ^[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}$ ]]; then
```
invidian, line 457: Hmm, maybe https://stackoverflow.com/a/5284410/2974814 ? :smile:

invidian (approved):
LGTM

assets/terraform-modules/packet/flatcar-linux/kubernetes/workers/cl/worker.yaml.tmpl
johananl: Not sure what you're suggesting. Could you explain?

johananl:
@invidian @rata I would like to merge this ASAP if there are no further concerns since we have a broken state in `master` right now. Are there any more blockers? I suggest that for non-blockers we open new issues and iterate later.

assets/terraform-modules/packet/flatcar-linux/kubernetes/workers/cl/worker.yaml.tmpl
invidian: If you insist on validating obtained data, perhaps we should do it right. So using `^((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)(\.|$)){4}$`, as current regex allows values like:
```
256.0.0.0
1.1.1.1.1
```

Which are not correct.

I was laughing, because validating IP addresses with regex is always fun :)

johananl: I don't "insist" on anything, just thought it's better to add at least a minimal validation to catch at least cases where the output is obviously not an IP address. This isn't comprehensive and doesn't handle IPv6 but it's better than nothing.

`1.1.1.1.1` doesn't match on my testing. `256.0.0.0` indeed matches, but so do `0.0.0.0` and `127.0.0.1` which are valid IPs which don't make sense as a peer address...

I'm happy to use a tighter regex if we have one which works reliably.

johananl: I've replaced the regex with `^((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)(\.|$)){4}$` and am testing.

iaguis (dismissed):
lgtm

invidian (approved):

invidian (changes requested):
Rebase required.

johananl:
> Rebase required.

Thanks.

P.S. GitHub tells me that :wink:

invidian:
>Thanks.

> P.S. GitHub tells me that wink

I know, but it's still showing the PR status in the list as "Waiting for review", which is not really correct :smile:

johananl:
Testing looks good. Need another LGTM.

iaguis (approved):

invidian (approved):

PR #1026, opened 2020-09-28
docs/rook-ceph-storage: Use correct apply command

johananl:
The doc tells the user to update the `rook-ceph` component config and then re-apply the `rook` component, which doesn't make sense.


invidian (approved):

surajssd (approved):
LGTM

Issue #1033, opened 2020-09-30
Fix find-updates script


knrt10:
Since the charts have been moved to assets directory, so the script needs to be updated.

PR #1034, opened 2020-09-30
Fix scripts/find-updates..sh

knrt10:
Some charts were moved to assets directory.

Add httpbin and fluo updates check

TLS provider is removed from packet so fixed that too.

closes: #1033


invidian (approved):
LGTM

iaguis (approved):

PR #1047, opened 2020-10-06
docs/KEYS: switch to personal GPG key

iaguis:
I lost access to the previous private key
(0xBECE2BD4B68D2071255A7A4E3689AD7F72EBB080) so I've added a
iago@kinvolk.io identity to my personal key
(0xB7DDBE7001408D7D7FD079D27E26368261BAD4F7) and I'll be using that key
from now on.

Additionally, I've added the new key to keys.openpgp.org
(https://keys.openpgp.org/search?q=iago%40kinvolk.io) which verifies the
email address associated with the key.


invidian (approved):
OK

Issue #502, opened 2020-05-28
De-duplicate timeout and retry intervals in test/components/*


invidian:
Right now each component test is in separate package and they all define the timeouts and retry intervals for checking if deployment converged. The values are usually 5 seconds and 5 minutes respectively. We should just define those values in e.g. `test/components` package and re-use them, if no specific value is needed.

PR #1049, opened 2020-10-07
test/*: de-duplicate value timeout and retryInterval

knrt10:
timeout and retryInterval were the same for most of the components and was
duplicated.

closes: #502

### DIFF
--- a/test/calico/metadata_access_test.go
+++ b/test/calico/metadata_access_test.go
@@ -21,7 +21,6 @@ import (
 	"context"
 	"os"
 	"testing"
-	"time"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -52,9 +51,6 @@ spec:
     - --timeout
     - "5"
 `
-
-	retryInterval = 1 * time.Second
-	timeout       = 5 * time.Minute
 )
 
 func TestNoMetadataAccessRandomPod(t *testing.T) { //nolint:funlen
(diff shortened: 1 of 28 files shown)


invidian (changes requested):
Hm, it seems there is more occurrences, even with 5sec/5min values. 

I also think 9 minute tests could also get a const, like `TimeoutSlow` or something.

knrt10:
There are 2 different values for timeout, 1 is 7min and other is 9 min. What should be the naming convention?

TimeoutSlow and TimeoutSlower?

invidian:
>There are 2 different values for timeout, 1 is 7min and other is 9 min. What should be the naming convention?

7 minutes can either be left as-is (as there is no duplication) or just increased to 9 minutes.

test/calico/metadata_access_test.go
```diff
@@ -54,7 +54,6 @@ spec:
 `
 
 	retryInterval = 1 * time.Second
```
invidian, line 56: It should be fine to retry after 5 seconds, we can get rid of this too.

test/components/cert-manager/cert-manager_test.go
```diff
@@ -47,7 +46,7 @@ func TestCertManagerDeployments(t *testing.T) {
 	for _, test := range testCases {
 		t.Run(fmt.Sprintf("cert-manager deployment:%s", test.deployment), func(t *testing.T) {
 			t.Parallel()
-			testutil.WaitForDeployment(t, client, namespace, test.deployment, time.Second*5, time.Minute*5)
+			testutil.WaitForDeployment(t, client, namespace, test.deployment, testutil.RetryInterval, testutil.Timeout) //nolint:goscope
```
invidian, line 49: It seems we should add `test := test` after for-range, rather than `//nolint:goscope`.

test/components/util/util.go
```diff
@@ -40,6 +40,15 @@ import (
 	"k8s.io/client-go/transport/spdy"
 )
 
+const (
+	// RetryInterval is time for test to retry.
+	RetryInterval = time.Second * 5
+	// Timeout is time after which tests stops and fails.
+	Timeout = time.Minute * 5
+	// TimeoutSlower is time after which tests stops and fails.
+	TimeoutSlower = time.Minute * 9
```
invidian, line 49: `Slower` sounds weird to me. I think just `Slow` might be better, so it indicates that this test case is slow, so it should get higher timeout. If we add `Slower`, we could also have `Slowest` etc. I think `Slow` is more clear.
```suggestion
	TimeoutSlow = time.Minute * 9
```

invidian (changes requested):
Nice work @knrt10

invidian (approved):
LGTM

invidian:
Rebase needed.

invidian (changes requested):
Rebase needed.

invidian (approved):

surajssd (approved):
LGTM

PR #1076, opened 2020-10-14
pkg/components: add Inspektor Gadget component

### LINKED ISSUE #1063: Add Inspektor Gadget component
We should add an [Inspektor Gadget](https://github.com/kinvolk/inspektor-gadget) component so users can inspect the cluster and use the Traceloop UI when they have [Headlamp](https://github.com/kinvolk/headlamp/) installed.

iaguis:
Fixes #1063

### DIFF
--- a/assets/charts/components/inspektor-gadget/templates/psp.yaml
+++ b/assets/charts/components/inspektor-gadget/templates/psp.yaml
@@ -1,0 +1,24 @@
+apiVersion: policy/v1beta1
+kind: PodSecurityPolicy
+metadata:
+  name: inspektor-gadget
+  annotations:
+    seccomp.security.alpha.kubernetes.io/allowedProfileNames: 'docker/default'
+    seccomp.security.alpha.kubernetes.io/defaultProfileName:  'docker/default'
+spec:
+  privileged: true
+  volumes:
+  - 'hostPath'
+  hostNetwork: true
+  hostPID: true
+  runAsUser:
+    rule: 'RunAsAny'
+  seLinux:
+    rule: 'RunAsAny'
+  supplementalGroups:
+    rule: 'RunAsAny'
+  fsGroup:
+    rule: 'RunAsAny'
+  readOnlyRootFilesystem: false
+  allowedHostPaths:
+  - pathPrefix: "/"
--- a/assets/charts/components/inspektor-gadget/values.yaml
+++ b/assets/charts/components/inspektor-gadget/values.yaml
@@ -1,0 +1,42 @@
+# Default values for inspektor-gadget.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+image:
+  repository: kinvolk/gadget
+  pullPolicy: IfNotPresent
+  # Overrides the image tag whose default is the chart appVersion.
+  tag: 202007010134320f732c # v0.2.0
+
+imagePullSecrets: []
+nameOverride: ""
+fullnameOverride: ""
+
+serviceAccount:
+  # Specifies whether a service account should be created
+  create: true
+  # Annotations to add to the service account
+  annotations: {}
+  # The name of the service account to use.
+  # If not set and create is true, a name is generated using the fullname template
+  name: ""
+
+podAnnotations: {}
+
+resources: {}
+  # We usually recommend not to specify default resources and to leave this as a conscious
+  # choice for the user. This also increases chances charts run on environments with little
+  # resources, such as Minikube. If you do want to specify resources, uncomment the following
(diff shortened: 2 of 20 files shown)


assets/charts/components/inspektor-gadget/templates/psp.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: policy/v1beta1
+kind: PodSecurityPolicy
+metadata:
+  name: inspektor-gadget-restricted
```
invidian, line 4: `-restricted`? From what? :joy: This is almost the most permissive policy I've seen :smile: 
```suggestion
  name: inspektor-gadget
```

assets/charts/components/inspektor-gadget/values.yaml
```diff
@@ -0,0 +1,42 @@
+# Default values for inspektor-gadget.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+image:
+  repository: docker.io/kinvolk/gadget
```
invidian, line 6: `docker.io` is default ?
```suggestion
  repository: kinvolk/gadget
```

invidian (dismissed):
Just some nits, otherwise LGTM, but I'm curious why CI is failing.

assets/charts/components/inspektor-gadget/templates/psp.yaml
iaguis: at least it can't use Secrets? :joy:

iaguis:
I've changed the namespace to `gadget`, for some reasoning see https://github.com/kinvolk/inspektor-gadget/pull/140#issuecomment-709934072.

> I'm curious why CI is failing.

I was using `WaitForDeployment()` instead of `ẀaitForDaemonSet()` :facepalm:

iaguis:
> I've changed the namespace to `gadget`, for some reasoning see [kinvolk/inspektor-gadget#140 (comment)](https://github.com/kinvolk/inspektor-gadget/pull/140#issuecomment-709934072).

After thinking about it some more and discussing with others I think what makes most sense is to keep IG in the `kube-system` namespace so it works out of the box with the current `kubectl gadget` plugin. Since it will have its own PSP this should be OK. If/when upstream IG changes its namespace to `gadget` we can change it too.

I changed the PR accordingly.

invidian:
Odd that CI didn't get triggered.

docs/configuration-reference/components/inspektor-gadget.md
```diff
@@ -0,0 +1,68 @@
...
+```
+
+## Attribute reference
+
+Table of all the arguments accepted by the component.
+
+Example:
+
+| Argument           | Description                                                                                                                   | Default       | Type   | Required |
+|--------------------|-------------------------------------------------------------------------------------------------------------------------------|--:-:----------|--:-:---|--:-:-----|
```
invidian, line 50: `:-:` does not render properly on GitHub.

invidian (dismissed):
Just one broken thing, otherwise LGTM.

assets/charts/components/inspektor-gadget/values.yaml
```diff
@@ -0,0 +1,42 @@
+# Default values for inspektor-gadget.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+image:
+  repository: kinvolk/gadget
```
knrt10, line 6: Unrelated but we should also push it to quay

test/components/inspektor-gadget/inspektorgadget_test.go
```diff
@@ -0,0 +1,35 @@
...
+
+	testutil "github.com/kinvolk/lokomotive/test/components/util"
+)
+
+const (
+	daemonset = "inspektor-gadget"
+	namespace = "kube-system"
+)
+
+func TestInspektorGadgetDeployment(t *testing.T) {
```
knrt10, line 31: TestInspektorGadgetDaemonSet

knrt10 (changes requested):
One small NIT, rest LGTM

assets/charts/components/inspektor-gadget/values.yaml
iaguis: Filed an issue upstream: https://github.com/kinvolk/inspektor-gadget/issues/141

knrt10: cool

knrt10 (approved):
Looks great. +1

iaguis:
I'll take @invidian's lgtm and @knrt10's one and merge this.

PR #1080, opened 2020-10-14
scripts/find-updates.sh: Fix providers

knrt10:

Regression from https://github.com/kinvolk/lokomotive/pull/824/


invidian (approved):
OK, though we could find more reliable way of obtaining those.

Issue #1077, opened 2020-10-14
Prevent deleting of component namespace if any other release object exists in this namespace


knrt10:
### Problem Statement

Currently, if the user has more than 1 component installed in same namespace and they try to delete a single component using **delete-namespace** flag, it does no checks and deletes the namespace. This issue occured as **headlamp** component is installed in **lokomotive-system** namespace. And deleting it will result in the deletion of **admission-webhook.**

### Offline discussion suggestions

We wanted a solution where we can install multiple components in same namespace.

- Install these components in dedicated namespace
- Track components installed in cluster
- Check if namespace is `lokomotive-system` and throw error to user
- Create a Validation webhook and avoid deletion of `lokomotive-system` namespace
- Check if other helm release are installed in that namespace, if not delete the namespace **(accepted soultion)**

invidian:
One more option was:
- Only remove the namespace if it has no resources left, but that would require iterating over all resources reported by `kubectl api-resources --namespaced=true` which would probably be slow and maybe even problematic from API point of view.

PR #1093, opened 2020-10-19
Add check for multiple release in Namespace

knrt10:
User should know that before deleting a component that other helm
release  exists in a namespace.

closes: #1077

### DIFF
--- a/cli/cmd/component-delete.go
+++ b/cli/cmd/component-delete.go
@@ -82,6 +82,7 @@ func runDelete(cmd *cobra.Command, args []string) {
 
 	if !confirm && !askForConfirmation(confirmationMessage) {
 		contextLogger.Info("Components deletion cancelled.")
+
 		return
 	}
 
--- a/pkg/components/util/install.go
+++ b/pkg/components/util/install.go
@@ -184,7 +184,7 @@ func UninstallComponent(c components.Component, kubeconfig []byte, deleteNSBool 
 
 	cfg, err := HelmActionConfig(ns, kubeconfig)
 	if err != nil {
-		return fmt.Errorf("failed preparing helm client: %w", err)
+		return fmt.Errorf("preparing Helm client: %w", err)
 	}
 
 	history := action.NewHistory(cfg)
--- a/pkg/components/util/install.go
+++ b/pkg/components/util/install.go
@@ -205,7 +205,16 @@ func UninstallComponent(c components.Component, kubeconfig []byte, deleteNSBool 
 		}
 	}
 
+	releasesList, err := action.NewList(cfg).Run()
+	if err != nil {
+		return fmt.Errorf("listing Helm releases: %w", err)
+	}
+
 	if deleteNSBool {
+		if len(releasesList) > 0 {
+			return fmt.Errorf("namespace %q may have other components installed, cannot remove namespace", ns)
+		}
+
 		if err := deleteNS(ns, kubeconfig); err != nil {
 			return err
 		}
--- a/test/components/component_delete_multiple_release_test.go
+++ b/test/components/component_delete_multiple_release_test.go
@@ -1,0 +1,42 @@
+// Copyright 2020 The Lokomotive Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// +build aws aws_edge packet
+// +build disruptivee2e
+
+package components_test
+
+import (
+	"testing"
+
+	"github.com/kinvolk/lokomotive/pkg/components"
+	_ "github.com/kinvolk/lokomotive/pkg/components/headlamp"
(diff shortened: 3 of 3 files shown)


cli/cmd/component-delete.go
```diff
@@ -20,6 +20,7 @@ import (
```
invidian, line 20: Commit message has double space.
```diff
- User should know that before deleting a component that other helm release  exists in a namespace.
+ User should know that before deleting a component that other helm release exists in a namespace.
```

cli/cmd/component-delete.go
```diff
@@ -127,10 +128,32 @@ func componentNamesToObjects(componentNames []string) ([]components.Component, e
 	return c, nil
 }
 
+func checkMutlipleReleaseNamespace(ns string, kubeconfig []byte, deleteNamespace bool) error {
```
invidian, line 132: We should not add more code into `cli/cmd`. I'd make it part of `util.UninstallComponent`.

cli/cmd/component-delete.go
```diff
@@ -127,10 +128,32 @@ func componentNamesToObjects(componentNames []string) ([]components.Component, e
...
+	if err != nil {
+		return fmt.Errorf("preparing helm client: %w", err)
+	}
+
+	releasesList, err := action.NewList(actionConfig).Run()
+	if err != nil {
+		return fmt.Errorf("listing helm releases: %w", err)
+	}
+
+	if len(releasesList) > 1 && deleteNamespace {
```
invidian, line 143: We should check for `>0` after we uninstall the release I think.

cli/cmd/component-delete.go
```diff
@@ -127,10 +128,32 @@ func componentNamesToObjects(componentNames []string) ([]components.Component, e
 	return c, nil
 }
 
+func checkMutlipleReleaseNamespace(ns string, kubeconfig []byte, deleteNamespace bool) error {
+	actionConfig, err := util.HelmActionConfig(ns, kubeconfig)
+	if err != nil {
+		return fmt.Errorf("preparing helm client: %w", err)
```
invidian, line 135:
```suggestion
		return fmt.Errorf("preparing Helm client: %w", err)
```

cli/cmd/component-delete.go
```diff
@@ -127,10 +128,32 @@ func componentNamesToObjects(componentNames []string) ([]components.Component, e
...
+	}
+
+	releasesList, err := action.NewList(actionConfig).Run()
+	if err != nil {
+		return fmt.Errorf("listing helm releases: %w", err)
```
invidian, line 140:
```suggestion
		return fmt.Errorf("listing Helm releases: %w", err)
```

cli/cmd/component-delete.go
```diff
@@ -127,10 +128,32 @@ func componentNamesToObjects(componentNames []string) ([]components.Component, e
...
+	}
+
+	if len(releasesList) > 1 && deleteNamespace {
+		return fmt.Errorf("multiple release exists in: %q namespace", ns)
```
invidian, line 144:
```suggestion
		return fmt.Errorf("namespace %q may have other components installed, refusing to remove", ns)
```

invidian (changes requested):

pkg/components/util/install.go
```diff
@@ -205,7 +205,16 @@ func UninstallComponent(c components.Component, kubeconfig []byte, deleteNSBool
 		}
 	}
 
+	releasesList, err := action.NewList(cfg).Run()
```
invidian, line 208: Can we have e2e test installing 2 components into single namespace and then trying to remove them?

We could for example allow installing `httpbin` in separate namespace to make it easier, as this component should only create minimal amount of manifests (no not-namespaces resources etc).

invidian (changes requested):
LGTM, but I would add tests for it.

pkg/components/util/install.go
knrt10: We can test for, headlamp as it is already installed in `lokomotive-system` namespace.

invidian: Hm, but if we uninstall headlamp, then headlamp tests may fail. I guess we can run them as disruptive then.

invidian (approved):
LGTM

invidian (approved):

pkg/components/util/install.go
```diff
@@ -205,7 +205,16 @@ func UninstallComponent(c components.Component, kubeconfig []byte, deleteNSBool
...
+	if err != nil {
+		return fmt.Errorf("listing Helm releases: %w", err)
+	}
+
 	if deleteNSBool {
+		if len(releasesList) > 0 {
+			return fmt.Errorf("namespace %q may have other components installed, refusing to remove", ns)
```
ipochi, line 215: better phrasing would be 
```
namespace `xyz` has other components installed: [list the components installed in the namespace], cannot delete namespace.
```

ipochi (dismissed):
One suggestion, but other than that looks good.
Lovely work @knrt10 :tada: 

Glad my idea `listing helm releases in the namespace` worked out quite nicely :)

pkg/components/util/install.go
invidian: We cannot really differentiate possible user's Helm releases from our releases at the moment, hence I suggested following phrasing on purpose. We cannot reliably say what components are installed there, as we do not track this information at the moment.

ipochi: True, we don't track that and that's why we also cannot delete them. I dont see it a harm to list out the releases installed in the namespace(even if they were user installed helm releases ) to show why a namespace cannot be deleted.

ipochi (approved):
Thank you for the PR @knrt10 :tada: 

lgtm

invidian (approved):

PR #1095, opened 2020-10-19
Update openEBS to v2.2.0.

ipochi:
Updates OpenEBS from `1.12.0` to `v2.2.0.

### DIFF
--- a/assets/charts/components/openebs-operator/Chart.yaml
+++ b/assets/charts/components/openebs-operator/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v1
-appVersion: 1.12.0
+appVersion: 2.2.0
 description: Containerized Storage for Containers
 home: http://www.openebs.io/
 icon: https://raw.githubusercontent.com/cncf/artwork/master/projects/openebs/icon/color/openebs-icon-color.png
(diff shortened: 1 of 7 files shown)


invidian:
@ipochi is this PR ready for review (if so, can we mark it as one)?

ipochi:
> @ipochi is this PR ready for review? If so, can we mark it as one?

@invidian This wont be a part of 0.5 release for reasons mentioned above.
You are however free to review the upgrade guide.

There is some work pending on the migration section which won't be updated unless migration instructions are changed/updated/simplified on OpenEBS end. 
Current migration has operator installations manually, which are not part of the OpenEBS helm chart but will be in the next version. Migration is optional

Also for 0.5 release, are we updating components ? If so I can create another PR that only updates OpenEBS to 2.2.0 and and update guide(without migration section)

ipochi:
updated the PR by splitting into two.

This PR now only covers updating OpenEBS
Upgrade guide is now covered in #1163

surajssd (dismissed):

invidian (dismissed):

invidian (approved):

surajssd (approved):

PR #1018, opened 2020-09-25
cli/cmd: cleanups part 3

invidian:
~**Includes commits from #1015**~

Part of #603.

PR #1098, opened 2020-10-19
cli/cmd: cleanups part 4

invidian:
~**Includes commits from #1018, don't review commits from #1018 in this PR.**~

Final part of the `cli/cmd`cleanups. #603 does not gets closed yet, as we should still refactor newly added API in `cli/cmd/cluster` to not print output to the user and to not ask for confirmation, as this should be done by the CLI package.

Part of #603.

Goals of this PR:
- To remove usage of global variables (local and from other packages). `viper` and `cobra` packages heavily depends on it, so we leave that for the time being in `cli/cmd`. `cli/cmd/cluster` should not import them.
- To allow experimenting with release decoupling (#312) using Go modules. That requires Go package-level isolation in the code, which is achieved with this PR, as now `cli/cmd` depends only on `cli/cmd/cluster` from `github.com/kinvolk/lokomotive` module.
  

Non goals:
- To build a nice abstraction around cluster operations like applying etc.
- To provide API which can be used programatically.
- To remove UI elements from isolated code.


cli/cmd/cluster/cluster.go
```diff
@@ -37,7 +37,6 @@ import (
 	_ "github.com/kinvolk/lokomotive/pkg/platform/packet"
```
knrt10, line 37: Nit: For commit message. `anonymouse`

cli/cmd/cluster-apply.go
```diff
@@ -62,6 +63,8 @@ func runClusterApply(cmd *cobra.Command, args []string) {
 		upgradeKubelets: upgradeKubelets,
 		skipComponents:  skipComponents,
 		verbose:         verbose,
+		configPath:      viper.GetString("lokocfg"),
+		valuesPath:      viper.GetString("lokocfg-vars"),
```
ipochi: Would it not be better if we had a higher level struct which had all the options common to all sub commands ?

These two lines are repetitive for every sub command and will be if another one is added.

ipochi (changes requested):

johananl (commented):
Thanks @invidian. I like the minimalist `cli/cmd` :+1: 

From a (very) cursory look I have the following questions:

- Is there a reason to keep the "core" logic in `cli/cmd` at all? e.g. why not `pkg/cluster`?
- Why are there files such as `component-apply.go` in `cli/cmd/cluster`? Shouldn't they live somewhere else?
- While we're at it, why do we have `cmd` and `cli/cmd`? Maybe rename `cli/cmd` to `cli` to simplify and avoid confusion?

I can take a more in-depth look later on if needed.

cli/cmd/cluster/cluster.go
invidian: Thanks, fixed.

cli/cmd/cluster-apply.go
invidian: Probably, yeah. But I didn't want to play around here with deduplication and building some fancy abstraction. My goal was to separate this code from CLI code. It is all definitely not a desired state yet, but moving this around should make further improvements easier.

invidian:
>Is there a reason to keep the "core" logic in cli/cmd at all? e.g. why not pkg/cluster?

I keep it as part of `cli/cmd` for now, to avoid moving it further away. The code still does printing and handles confirmation, which I think should also be removed from it in the future.

I deliberately did not put this in `pkg`, as I think it is still messy and requires re-structuring. This PR mostly just introduce package level separation, removes globals etc so re-structuring can be done easier. Eventually, most of this should go to `pkg`, yes.

 >Why are there files such as component-apply.go in cli/cmd/cluster? Shouldn't they live somewhere else that's related to components?

`component apply` operation optionally depends on cluster configuration, as it may take the kubeconfig from it. If it would be fully independent, then yes, we could move it closer to components code.

>While we're at it, why do we have cmd and cli/cmd? Maybe rename cli/cmd to cli to simplify and avoid confusion?

Yes... We should probably have `cmd/lokoctl` and `cli/lokoctl`, as `cmd/X` is recommended for `main` packages for binaries and `cli/X` is a good place to handle CLI logic (which can be used in multiple binaries for example, or to be used programatically. But this is all out of scope for this PR.

cli/cmd/cluster/component.go
```diff
@@ -0,0 +1,47 @@
...
+	_ "github.com/kinvolk/lokomotive/pkg/components/rook"
+	_ "github.com/kinvolk/lokomotive/pkg/components/rook-ceph"
+	_ "github.com/kinvolk/lokomotive/pkg/components/velero"
+	_ "github.com/kinvolk/lokomotive/pkg/components/web-ui"
+
+	"github.com/kinvolk/lokomotive/pkg/components"
+)
+
+// AvailableComponents returns list of valid component names.
+func AvailableComponents() []string {
```
ipochi, line 45: nit: maybe `SupportedComponents` or `AllComponents` ?

ipochi (approved):
lgtm

cli/cmd/cluster/component.go
invidian: `SupportedComponents`  might be ambiguous, as it's unclear what kind of "support" we talk about here.

`AllComponents` is perhaps fine, as it's a synonym for `AvailableComponents`, but I'm not sure if it's worth changing :smile:

PR #1104, opened 2020-10-20
gangway: add a serviceaccount

### LINKED ISSUE #1103: Gangway gives an empty CA when outputting kubeconfig
When deploying Dex+Gangway to a new cluster and trying to log in, the CLI instructions show this:

```
echo "" \ > /tmp/ca-test-oidc.pem
```

In quotes should be the cluster CA but it is empty.

This happens because Gangway is using the default service account and since https://github.com/kinvolk/lokomotive/pull/704 we don't mount it in pods by default. This causes the CA file that Gangway expects (`/var/run/secrets/kubernetes.io/serviceaccount/ca.crt`) not to be in the pod file system.

iaguis:
Gangway relies on a service account (any service account) to be mounted
on the pod file system to figure out the cluster CA.

We've disabled automounting of the default service account with the
our admission wehbook and this was breaking Gangway.

This creates a dummy service account so the Gangway pod has access to
the cluster CA file.

Fixes #1103 

TODO:

- [x] Test

### DIFF
--- a/assets/charts/components/gangway/templates/deployment.yaml
+++ b/assets/charts/components/gangway/templates/deployment.yaml
@@ -20,6 +20,7 @@ spec:
         runAsNonRoot: true
         runAsUser: 65534
         runAsGroup: 65534
+      serviceAccountName: gangway
       initContainers:
       - name: download-theme
         image: alpine/git:1.0.7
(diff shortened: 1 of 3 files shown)


invidian (approved):

knrt10 (approved):
We could add `deploy.Spec.Template.Spec.ServiceAccountName` as a constant. But no need to change. +1. We need this merged ASAP as headlamp PR depends on it.

PR #1105, opened 2020-10-21
docs/configuration-reference: remove "Example:"

iaguis:
It makes no sense, what follows is not an example but a table with
arguments.


ipochi (approved):
Thanks for the PR @iaguis 

LGTM :tada:

invidian (approved):

PR #1106, opened 2020-10-21
velero: restic plugin region value required

knrt10:
This is required but the code was using it as optional. Also, some docs formatting changes.

### DIFF
--- a/docs/configuration-reference/components/velero.md
+++ b/docs/configuration-reference/components/velero.md
@@ -60,20 +60,20 @@ component "velero" {
 
   # openebs {
   #   credentials = file("cloud-credentails-file")
-  #		provider		= "aws"
-	#
+  #   provider		= "aws"
+  #
   #   backup_storage_location {
   #     provider = "aws"
   #     region 	 = "my-region"
   #     bucket 	 = "my-bucket"
-  #			name     = "my-backup-location"
+  #     name     = "my-backup-location"
   #   }
   #
   #   volume_snapshot_location {
-  #			bucket 	 = "my-bucket"
-  #			region 	 = "my-region"
-  #			provider = "aws"
-  #			name 		 = "my-snapshot-location"
+  #     bucket   = "my-bucket"
+  #     region   = "my-region"
+  #     provider = "aws"
+  #     name     = "my-snapshot-location"
   #     prefix   = "backup-prefix"
   #     local    = false
   #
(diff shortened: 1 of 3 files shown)


docs/configuration-reference/components/velero.md
```diff
@@ -60,20 +60,20 @@ component "velero" {
 
   # openebs {
   #   credentials = file("cloud-credentails-file")
-  #		provider		= "aws"
```
invidian, line 63: Seems unrelated, can we commit it separately please?

docs/configuration-reference/components/velero.md
```diff
@@ -150,7 +151,7 @@ Example:
...
 | `restic.backup_storage_location.bucket`              | Cloud storage bucket name for storing backups.                                                                              | -                                                 | string | true     |
 | `restic.backup_storage_location.name`                | Name for backup location object on the cluster.                                                                             | -                                                 | string | false    |
-| `restic.backup_storage_location.region`              | Cloud provider region for storing snapshots.                                                                                | `eu-west-1`                                       | string | false    |
+| `restic.backup_storage_location.region`              | Cloud provider region for storing snapshots.                                                                                | `eu-west-1`                                       | string | true     |
```
invidian, line 154: Should this value actually be required? If we list a default value for it?

invidian (changes requested):

docs/configuration-reference/components/velero.md
knrt10: That's what got me confused. Because it's not using the default value

pkg/components/velero/restic/restic.go
```diff
@@ -38,7 +38,7 @@ type BackupStorageLocation struct {
 	Provider string `hcl:"provider"`
 	Bucket   string `hcl:"bucket"`
 	Name     string `hcl:"name,optional"`
-	Region   string `hcl:"region,optional"`
```
ipochi, line 41: @knrt10 

`Region` cannot be required as its only required for aws provider in the `BackupStorageLocation` but when using (GCP)[https://github.com/vmware-tanzu/velero-plugin-for-gcp/blob/main/backupstoragelocation.md] or (Azure)[https://github.com/vmware-tanzu/velero-plugin-for-microsoft-azure/blob/main/backupstoragelocation.md] it is not required.

The idea of having default seems not right to me.

ipochi (changes requested):

ipochi:
This is not needed anymore @knrt10.

Could you please closing this.

PR #1108, opened 2020-10-21
docs: Add CoC and git parts from Kinvolk's contribution repository

joaquimrocha:


docs/CONTRIBUTING.md
```diff
@@ -11,6 +12,10 @@
 - [Updating dependencies](#updating-dependencies)
 - [Testing and linting requirements](#testing-and-linting-requirements)
 
+## Code of Conduct
+
+Please refer to the Kinvolk [Code of Conduct](https://github.com/kinvolk/contribution/blob/master/CODE_OF_CONDUCT.md).
```
invidian, line 17: I wonder if this will be picked up by https://github.com/kinvolk/lokomotive/community.

invidian (approved):
LGTM, just one Q

docs/CONTRIBUTING.md
joaquimrocha: Interesting. Probably it won't. But if that's the case we could add a `CODE_OF_CONDUCT.md` file just with the same text, like the Kubernetes Dashboard does: https://github.com/kubernetes/dashboard/blob/master/code-of-conduct.md

In either case, I think it's important to keep this section in the `contributing` doc.

ipochi (approved):
LGTM

docs/CONTRIBUTING.md
invidian: I'd add this file and keep the section then. We will need to periodically update it then.

vbatts: I see an "add" button on that community page. I think we can link it once merged

vbatts: hm. maybe

vbatts:
LGTM

PR #1132, opened 2020-10-28
Add `require_volume_annotation` for restic plugin

ipochi:
This PR is based on top of #1131 and would need to be merged first.

Adds a new field `require_volume_annotation` that was introduced in Velero v1.5.2.

If enabled the user will  have to manually add annotations to the pod when taking backups.

Default is `false`, i.e volumes will be backed up without the need for annotation.


assets/terraform-modules/aws/flatcar-linux/kubernetes/versions.tf
```diff
@@ -6,7 +6,7 @@ terraform {
   required_providers {
```
invidian, line 6: This shouldn't be here I suppose?

pkg/components/velero/component.go
```diff
@@ -201,14 +201,14 @@ func (c *component) getProvider() (provider, error) {
 		providers = append(providers, c.Restic)
 	}
 
-	if len(providers) > 1 {
```
invidian, line 204: It seems like something which should be committed separately. Why changing the order makes a difference?

docs/configuration-reference/components/velero.md
```diff
@@ -86,6 +86,8 @@ component "velero" {
   # restic {
   #   credentials = file("cloud-credentials-file")
   #
+  #   default_volumes_to_restic = true
```
invidian, line 89: I know this is upstream name for this flag, but perhaps we could come up with some better name for it. For example, we're already in `restic` block, so there is no need to add `to_restic` IMO.

Maybe we could have something like `require_volume_annotation`? And we set it to true by default?

invidian (dismissed):
Some nits

invidian:
I just noticed this PR includes commits from #1131, sorry that I put all review comments here. It's good to mention such things in PR description :)

docs/configuration-reference/components/velero.md
ipochi: I agree on the naming, but I am on the fenc with setting the value as `true` by default as its not the default behavior of Velero.

invidian: No no, `require_volume_annotation = true` would result in `default_volumes_to_restic = false`.

invidian:
There are conflicts which needs to be resolved.

invidian (changes requested):
There are conflicts which needs to be resolved.

ipochi:
@invidian yes, I found a bug in the existing code and resolving that.

docs/configuration-reference/components/velero.md
```diff
@@ -147,7 +149,8 @@ Table of all the arguments accepted by the component.
...
 | `restic`                                             | Configure Restic provider for Velero.                                                                                       | -                                                 | object | false    |
 | `restic.credentials`                                 | Content of cloud provider credentials.                                                                                      | -                                                 | string | true     |
-| `restic.backup_storage_location.provider`            | Cloud provider name for storing backups.                                                                                    | -                                                 | string | true     |
+| `restic.require_volume_annotation`                   | Backup all pod volumes without having to apply annotation on the pod when using restic.                                     | true                                              | bool   | false    |
```
invidian, line 152: Maybe something like:
```suggestion
| `restic.require_volume_annotation`                   | If false, all pod volumes will be backed-up automatically. If true, <X> annotation needs to be added to pods to mark which volumes should be backed up.                                     | true                                              | bool   | false    |
```

invidian (changes requested):
Just one nit, otherwise LGTM

ipochi:
@invidian There is an opt-out feature in Velero which should be added as annotation if volume is not be backed up by restic by default.

Should we flip the default value for `require_volume_annotation` to `false` i.e opt-out strategy where volume will not be considered for back up if the resource has the following annotation `backup.velero.io/backup-volumes-excludes` ?

invidian:
>Should we flip the default value for require_volume_annotation to false i.e opt-out strategy where volume will not be considered for back up if the resource has the following annotation backup.velero.io/backup-volumes-excludes ?

Sounds good :+1: If we have a place for it, we could mention it in our documentation for Velero component.

invidian (approved):
LGTM

surajssd (approved):
LGTM

PR #1133, opened 2020-10-28
Update terraform providers

ipochi:


assets/terraform-modules/bootkube/versions.tf
```diff
@@ -5,7 +5,7 @@ terraform {
   required_providers {
```
invidian, line 5: Why is there `10` in the commit message?

pkg/assets/generated_assets.go
```diff
@@ -5359,9 +5359,9 @@ var vfsgenAssets = func() http.FileSystem {
 		"/terraform-modules/aws/flatcar-linux/kubernetes/versions.tf": &vfsgen۰CompressedFileInfo{
```
invidian, line 5359: To be honest, I'm not a fan of committing assets generation separately, as it breaks the atomicity of the commits. Maybe we can just commit everything as one commit?

invidian (dismissed):
Looks good

invidian (approved):
:heart:

surajssd (approved):
Quick LGTM :laughing:

PR #1135, opened 2020-10-28
aws-ebs-csi-driver: update to v0.7.0

ipochi:

### DIFF
--- a/assets/charts/components/aws-ebs-csi-driver/values.yaml
+++ b/assets/charts/components/aws-ebs-csi-driver/values.yaml
@@ -85,5 +85,6 @@ serviceAccount:
   snapshot:
     annotations: {}
 
+# Lokomotive specific addition.
 # False if you want to use a different storage option, e.g. OpenEBS
 enableDefaultStorageClass: true
--- a/assets/charts/components/aws-ebs-csi-driver/Chart.yaml
+++ b/assets/charts/components/aws-ebs-csi-driver/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v1
-appVersion: "0.6.0"
+appVersion: "0.7.0"
 name: aws-ebs-csi-driver
 description: A Helm chart for AWS EBS CSI Driver
-version: 0.5.0
+version: 0.6.0
 kubeVersion: ">=1.13.0-0"
 home: https://github.com/kubernetes-sigs/aws-ebs-csi-driver
 sources:
--- a/assets/charts/components/aws-ebs-csi-driver/templates/controller.yaml
+++ b/assets/charts/components/aws-ebs-csi-driver/templates/controller.yaml
@@ -23,7 +23,6 @@ spec:
     spec:
       nodeSelector:
         kubernetes.io/os: linux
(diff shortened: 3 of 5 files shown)


assets/charts/components/aws-ebs-csi-driver/values.yaml
```diff
@@ -84,6 +84,3 @@ serviceAccount:
     annotations: {}
   snapshot:
     annotations: {}
-
-# False if you want to use a different storage option, e.g. OpenEBS
-enableDefaultStorageClass: true
```
invidian, line 89: This seems like our change. Should this be moved to component values?

ipochi: Is it our change ?

I didn't see the `XXX Lokomotive specific change` comment.

invidian: Then I'd move it to component code to avoid confusion.

invidian: ping @ipochi

ipochi: @invidian This is already part of the templated `values.yaml`

https://github.com/kinvolk/lokomotive/blob/4b4e249506b2ca85a7c897ba64c4ec0bb8116a62/pkg/components/aws-ebs-csi-driver/component.go#L32

invidian: I thought we just hardcode the value here. Sorry for the confusion. In this case, this should stay here as a documentation and a default value for value used by the Helm chart. It shouldn't be removed.

ipochi: @invidian 

I haven't removed it manually. Its not part of the helm chart.
I suppose it was added by us in the previous version of the chart, but its not present in 0.6.0 version of the upstream chart.

invidian: Yes, it seems like our addition used here:
```
$ git grep enableDefaultStorageClass
assets/charts/components/aws-ebs-csi-driver/templates/storageclass.yaml:{{- if .Values.enableDefaultStorageClass }}
```

Perhaps we can add `XXX Lokomotive specific change` then.

Actually, it seems entire storage class manifest is our addition. We should upstream it.

invidian: Opened an issue upstream: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/605.

invidian: And created a PR: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/606

ipochi: please re-review.

invidian (approved):
LGTM, thanks @ipochi

knrt10 (approved):

iaguis (approved):

PR #1149, opened 2020-10-30
velero: Add provider field and remove default configuration in restic plugin

ipochi:
Brings back the addition of `provider` field in the component configuration.
This is done so to be able to provide default configurations for the plugins.

Also it removes the default BackupStorageLocation config in restic
plugin.

Name is optional and its default value is `default` in Helm chart.
Provider is required and making it a default of `aws` doesn't make sense
as the user has to provide  value to it anyway.

Region is also specifc to the provider used. If not `aws` then there is
no need to input that value.

Release notes: 

Add `provider` field in the component configuration.
example:

```
component `velero` {
  .
  .
  provider = "openebs"
 
  openebs {
    .
    .
  }
}
```

### DIFF
--- a/docs/configuration-reference/components/velero.md
+++ b/docs/configuration-reference/components/velero.md
@@ -37,6 +37,7 @@ Velero component configuration example:
 # velero.lokocfg
 component "velero" {
 
+  # provider = "azure/openebs/restic"
   # azure {
   #   # Required arguments.
   #   subscription_id = "9e5ac23c-6df8-44c4-9790-6f6decf96268"
@@ -110,6 +111,7 @@ Table of all the arguments accepted by the component.
 | Argument                                             | Description                                                                                                                 | Default                                           | Type   | Required |
 |------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------|---------------------------------------------------|--------|----------|
 | `namespace`                                          | Namespace to install Velero.                                                                                                | "velero"                                          | string | false    |
+| `provider`                                           | Provider sets which provider block to use for the configuration. Supported values are: `azure`, `openebs` and `restic`.     | -                                                 | string | true     |
 | `metrics`                                            | Configure Prometheus to scrape Velero metrics. Needs the [Prometheus Operator component](prometheus-operator.md) installed. | -                                                 | object | false    |
 | `metrics.enabled`                                    | Adds Prometheus annotations to Velero deployment if enabled.                                                                | false                                             | bool   | false    |
 | `metrics.service_monitor`                            | Adds ServiceMonitor resource for Prometheus. Requires `metrics.enabled` as true.                                            | false                                             | bool   | false    |
--- a/docs/configuration-reference/components/velero.md
+++ b/docs/configuration-reference/components/velero.md
@@ -147,10 +147,10 @@ Table of all the arguments accepted by the component.
 | `openebs.volume_snapshot_location.s3_url`            | S3 API URL.                                                                                                                 | -                                                 | string | false    |
 | `restic`                                             | Configure Restic provider for Velero.                                                                                       | -                                                 | object | false    |
 | `restic.credentials`                                 | Content of cloud provider credentials.                                                                                      | -                                                 | string | true     |
-| `restic.backup_storage_location.provider`            | Cloud provider name for storing backups.                                                                                    | -                                                 | string | false    |
+| `restic.backup_storage_location.provider`            | Cloud provider name for storing backups.                                                                                    | -                                                 | string | true     |
 | `restic.backup_storage_location.bucket`              | Cloud storage bucket name for storing backups.                                                                              | -                                                 | string | true     |
-| `restic.backup_storage_location.name`                | Name for backup location object on the cluster.                                                                             | -                                                 | string | false    |
-| `restic.backup_storage_location.region`              | Cloud provider region for storing snapshots.                                                                                | `eu-west-1`                                       | string | false    |
+| `restic.backup_storage_location.name`                | Name for backup location object on the cluster.                                                                             | "default"                                         | string | false    |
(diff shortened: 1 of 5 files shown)


knrt10:
Can you please address check linter and build errors?

ipochi:
> Can you please address linter and build errors?

yep, I am looking into the bug introduced by this fix.

pkg/components/velero/component.go
```diff
@@ -72,6 +72,9 @@ func newComponent() *component {
 			Enabled:        false,
 			ServiceMonitor: false,
 		},
+		Azure:   &azure.Configuration{},
```
invidian, line 75: How is this related to default restic configuration removal?

pkg/components/velero/restic/restic.go
```diff
@@ -44,11 +44,7 @@ type BackupStorageLocation struct {
 // NewConfiguration returns the default restic configuration.
 func NewConfiguration() *Configuration {
 	return &Configuration{
-		BackupStorageLocation: &BackupStorageLocation{
```
invidian, line 47: Do we also need a documentation change for this?

ipochi: I don't think so.

invidian: All those 3 parameters are marked in documentation as `required = false`. I though this is because we were setting a default values for them. So they are actually optional?

invidian (commented):
LGTM, except the docs for second commit (discussed in other comment).

pkg/components/velero/component.go
```diff
@@ -189,15 +193,15 @@ func (c *component) getSupportedProviders() []string {
 func (c *component) getProvider() (provider, error) {
 	providers := []provider{}
 
-	if c.Azure != nil {
+	if c.Provider == "azure" {
```
invidian, line 196: Actually, append logic doesn't make sense anymore if we can only have one provider configured at all times. I'd just replace entire function body with either switch/case or map lookup.

invidian (changes requested):

pkg/components/velero/restic/restic.go
ipochi: updated the docs.

docs/configuration-reference/components/velero.md
```diff
@@ -37,6 +37,7 @@ Velero component configuration example:
 # velero.lokocfg
 component "velero" {
 
+  #provider = "azure/openebs/restic"
```
knrt10, line 40: I don't know but no space here seems a little bit weird.

invidian: What space you mean? Blank line separating parameter from block below?

docs/configuration-reference/components/velero.md
```diff
@@ -110,6 +111,7 @@ Table of all the arguments accepted by the component.
 | Argument                                             | Description                                                                                                                 | Default                                           | Type   | Required |
 |------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------|---------------------------------------------------|--------|----------|
 | `namespace`                                          | Namespace to install Velero.                                                                                                | "velero"                                          | string | false    |
+| `provider`                                           | Provider sets which provider block to use for the configuration. Supported values `azure`, `openebs`, `restic`.             | -                                                 | string | true     |
```
invidian, line 114:
```suggestion
| `provider`                                           | Provider sets which provider block to use for the configuration. Supported values `azure`, `openebs` and `restic`.             | -                                                 | string | true     |
```

docs/configuration-reference/components/velero.md
```diff
@@ -145,10 +147,10 @@ Table of all the arguments accepted by the component.
...
-| `restic.backup_storage_location.name`                | Name for backup location object on the cluster.                                                                             | -                                                 | string | false    |
-| `restic.backup_storage_location.region`              | Cloud provider region for storing snapshots.                                                                                | `eu-west-1`                                       | string | false    |
+| `restic.backup_storage_location.name`                | Name for backup location object on the cluster.                                                                             | "default"                                         | string | false    |
+| `restic.backup_storage_location.region`              | Cloud provider region for storing snapshots. Required if `restic.backup_storage_location.provider = aws`                    | -                                                 | string | false    |
```
invidian, line 153:
```suggestion
| `restic.backup_storage_location.region`              | Cloud provider region for storing snapshots. Required if `restic.backup_storage_location.provider = aws`.                   | -                                                 | string | false    |
```

invidian (changes requested):
Some nits.

docs/configuration-reference/components/velero.md
knrt10: Yes, I meant `# provider` as we are using it in line with provider name in test file.

docs/configuration-reference/components/velero.md
```diff
@@ -110,6 +111,7 @@ Table of all the arguments accepted by the component.
 | Argument                                             | Description                                                                                                                 | Default                                           | Type   | Required |
 |------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------|---------------------------------------------------|--------|----------|
 | `namespace`                                          | Namespace to install Velero.                                                                                                | "velero"                                          | string | false    |
+| `provider`                                           | Provider sets which provider block to use for the configuration. Supported values `azure`, `openebs` and `restic`.          | -                                                 | string | true     |
```
invidian, line 114:
```suggestion
| `provider`                                           | Provider sets which provider block to use for the configuration. Supported values are: `azure`, `openebs` and `restic`.          | -                                                 | string | true     |
```

invidian (dismissed):
LGTM

invidian (approved):

knrt10 (approved):

PR #1160, opened 2020-11-03
Fix installation guide and add mention about it to release documentation

invidian:


docs/installer/lokoctl.md
```diff
@@ -17,19 +17,19 @@ These binaries can be manually downloaded and installed.
    keys](https://github.com/kinvolk/lokomotive/blob/master/docs/KEYS.md).
 
 ```console
-gpg --verify lokoctl_0.1.0_linux_amd64.tar.gz.sig
+gpg --verify lokoctl_0.5.0_linux_amd64.tar.gz.sig
```
ipochi, line 20: Why have a hard coded number ?

```suggestion
gpg --verify lokoctl_0.X.Y_linux_amd64.tar.gz.sig
```

and point to the releases page. This way master branch will have the `0.x.y` reference and each release tag branch will have the correct version.

ipochi (changes requested):

docs/installer/lokoctl.md
invidian: Why not X.Y.Z then? I think having a hardcoded number is better, as it's something you can copy-paste to the terminal and run.

ipochi: Yes its true its better but not for main branch.

I am fine with keeping the latest release number in the main branch if we add this task to the release checklist of updating everytime.

ipochi (approved):
Thanks for the PR @invidian 

LGTM :tada:

PR #1167, opened 2020-11-06
pkg/components/cluster-autoscaler: fix duplicate device error

invidian:
Packet API seems to have a bug, that when you use Project API key
for querying devices, it returns duplicated entries. However IDs
of the devices should be the same, so it should be safe to relax
the check on those, as if there would actually be 2 different devices
with the same name, they should have different IDs.


surajssd (dismissed):

invidian:
Found a typo in comment and commit message, had to re-push.

iaguis (approved):
lgtm

invidian:
Created also #1168 to fix flaky FLUO test.

PR #1168, opened 2020-11-06
test/components/flatcar-linux-update-operator: fix flaky idle test

invidian:
**Includes commits from #1167, as otherwise `Packet` pipeline would fail**


iaguis (approved):
lgtm

Issue #1171, opened 2020-11-09
OpenEBS validating webhook times out on component uninstalling causing CI be flaky


invidian:
Most likely introduced in #1095 

Seems related to https://github.com/openebs/openebs/issues/3046.

Failing CI jobs example:
- https://yard.lokomotive-k8s.net/teams/main/pipelines/pr-packet/jobs/test-lokomotive-packet-deployment/builds/1155
- https://yard.lokomotive-k8s.net/teams/main/pipelines/pr-aws/jobs/test-lokomotive-aws-deployment/builds/1169
- https://yard.lokomotive-k8s.net/teams/main/pipelines/pr-packet/jobs/test-lokomotive-packet-deployment/builds/1155.1

ipochi:
@Mateusz This is not a regression but a long standing issue.

Webhook policy in the previous versions after 1.10 has been set to `Fail`. Since we were running 1.12.0 before #1095 , it was already `Fail`.

Workaround mentioned in to change that to `Ignore`.

invidian:
>@mateusz This is not a regression but a long standing issue.

It's a regression in Lokomotive, as CI was not failing on that before updating OpenEBS.

>Webhook policy in the previous versions after 1.10 has been set to Fail. Since we were running 1.12.0 before #1095 , it was already Fail.

> Workaround mentioned in to change that to Ignore.

I'm fine with the workaround, if that's what upstream recommends. Can you take care of implementing it @ipochi ?

invidian:
Resolved in #1171

PR #1172, opened 2020-11-09
fix: openebs-operator webhook issue

ipochi:
This commit addresses #1171

OpenEBS by default ships the failurePolicy as 'Fail', however this
causes sometimes to unexpectedly hit a known bug in OpenEBS
https://github.com/openebs/openebs/issues/3046.
One of the workarounds suggested is to change the failurePolicy to
`Ignore`.

### DIFF
--- a/pkg/components/openebs-operator/component.go
+++ b/pkg/components/openebs-operator/component.go
@@ -77,8 +77,12 @@ ndmOperator:
     "{{ .NDMSelectorLabel }}": "{{ .NDMSelectorValue }}"
 {{- end }}
 
-{{- if and .NDMSelectorLabel .NDMSelectorValue }}
 webhook:
+  # OpenEBS by default ships the failurePolicy as 'Fail', however this causes sometimes to unexpectedly
+  # hit a known bug in OpenEBS https://github.com/openebs/openebs/issues/3046.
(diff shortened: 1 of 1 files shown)


pkg/components/openebs-operator/component.go
```diff
@@ -79,6 +79,10 @@ ndmOperator:
 
 {{- if and .NDMSelectorLabel .NDMSelectorValue }}
 webhook:
+  # OpenEBS by default ships the failurePolicy as 'Fail', however this causes sometimes to unexpectedly
+  # hit a known bug in OpenEBS https://github.com/openebs/openebs/issues/3046.
+  # One of the workarounds suggested is to change the failurePolicy to "Ignore".
+  failurePolicy: 'Ignore'
```
invidian, line 85: `if` condition must be moved below, otherwise it will be ignored only if selectors are specified.

invidian (changes requested):

pkg/components/openebs-operator/component.go
ipochi: Oops. Will fix.

invidian (approved):
LGTM

invidian (approved):

surajssd (approved):
LGTM

PR #1177, opened 2020-11-09
packet: ensure Calico CDRs gets created before use

### LINKED ISSUE #1175: Bootkube occasionally fails trying to install calico-host-protection chart before Calico
Due to fact, that we enabled [--wait](1085) in bootkube and that we are not able to define the order of which bootkube charts are being installed, all charts are now installed in parallel.

This is problematic, as Calico chart installs CRDs, which are then instantiated by `calico-host-protection`. It sometimes happens, that `calico-host-protection` gets installed before `Calico` chart, which results in following error message:
```
module.packet-ci1604930008-rt.null_resource.bootkube-start (remote-exec): Nov 09 14:02:33 ci1604930008-rt-controller-0 bootkube-start[1865]: Installing chart "calico-host-protection" in namespace "kube-system" failed: unable to build kubernetes objects from release manifest: unable to recognize "": no matches for kind "HostEndpoint" in version "crd.projectcalico.org/v1"kube-system/calico: Clearing discovery cache
```

I'd say this happen sporadically, but this is still something we should deal with.

Possible solutions:
- Copy Calico CRDs to `manifests` directory when bootkube runs, so they will be registered before charts are installed. A bit hacky, but should be simple to implement.
- Move `calico-host-protection` component installation outside of bootkube to be installed. This seem like a reasonable thing to do, as right now we abuse bootkube a bit with installation of those components, and while working on that we could perhaps also fix #478.
- Implement `calico-host-protection` chart as values for Calico chart. As mid-term solution, that would be possible, but may end up overly complex, given the amount of resources created by `calico-host-protection` chart. Perhaps we should avoid this solution.

invidian:
Currently, bootkube installs all charts in parallel, which causes
calico-host-protection chart installation to fail ocasionally, as it
gets installed before Calico chart itself.

This commit adds a workaround for this, by making an extra copy of
Calico CRDs into bootkube manifests directory, as those will be applied
before Helm charts. This ensures, that the CRDs are created before
calico-host-protection is installed, however it is rather hacky and
should not be considered as final solution for the issue, but a
temporary workaround to ensure cluster bootstrapping remains reliable.

More details in issue #1175.

Closes #1175

### DIFF
--- a/assets/terraform-modules/packet/flatcar-linux/kubernetes/calico-host-protection.tf
+++ b/assets/terraform-modules/packet/flatcar-linux/kubernetes/calico-host-protection.tf
@@ -1,3 +1,20 @@
+// TODO: Currently, bootkube installs all charts in parallel, which causes
+// calico-host-protection chart installation to fail ocasionally, as it
+// gets installed before the Calico chart itself.
+//
+// To workaround this issue, we make an extra copy of the Calico CRDs into
+// the bootkube manifests directory, as those will be applied
+// before all Helm charts. This ensures that the CRDs are created before
+// the calico-host-protection chart is installed.
+//
+// See https://github.com/kinvolk/lokomotive/issues/1175 for more details.
+resource "local_file" "calico_crds" {
+  for_each = fileset("${var.asset_dir}/charts/kube-system/calico/crds", "*.yaml")
+
+  content  = file("${var.asset_dir}/charts/kube-system/calico/crds/${each.value}")
+  filename = "${var.asset_dir}/manifests-calico-cdrs/${each.value}"
+}
+
 resource "local_file" "calico_host_protection" {
   content = templatefile("${path.module}/calico-host-protection.yaml.tmpl", {
     host_endpoints = [
--- a/assets/terraform-modules/packet/flatcar-linux/kubernetes/ssh.tf
+++ b/assets/terraform-modules/packet/flatcar-linux/kubernetes/ssh.tf
@@ -70,6 +70,7 @@ resource "null_resource" "copy-assets-dir" {
     module.bootkube,
     null_resource.copy-controller-secrets,
     local_file.calico_host_protection,
+    local_file.calico_crds,
   ]
 
   connection {
@@ -102,7 +103,10 @@ resource "null_resource" "bootkube-start" {
 
   provisioner "remote-exec" {
     inline = [
-      "sudo mv $HOME/assets /opt/bootkube",
+      "sudo mv $HOME/assets /opt/bootkube/",
+      # This is needed, as the bootkube-start script will move all files matching
+      # /opt/bootkube/asssets/manifests-*/*.yaml into /opt/bootkube/assets/manifests.
+      "sudo mkdir /opt/bootkube/assets/manifests",
       "sudo systemctl start bootkube || (sudo journalctl -u bootkube --no-pager; exit 1)",
     ]
   }


ipochi:
would it not be easy to have a  directory structure in bootkube with prefix `00-` / `01-` before chart directories and control the helm chart execution order there ?

invidian:
>would it not be easy to have a directory structure in bootkube with prefix 00- / 01- before chart directories and control the helm chart execution order there ?

It would not be easier, as it would require bootkube development, we would start abusing file-system for non-related functionality (storing properties in directory names) and would make setup more complex. After ordering, we may need to add conditional wait etc etc in the future, so it is better to keep the features we implement there to the minimum.

ipochi:
> > would it not be easy to have a directory structure in bootkube with prefix 00- / 01- before chart directories and control the helm chart execution order there ?
> 
> It would not be easier, as it would require bootkube development, we would start abusing file-system for non-related functionality (storing properties in directory names) and would make setup more complex. After ordering, we may need to add conditional wait etc etc in the future, so it is better to keep the features we implement there to the minimum.

Adding number prefixes to directory/file name is pretty common in Kubernetes ecosystem. Besides other use cases you've mentioned is futuristic, not pressing nor visible at present.

invidian:
Also, this workaround only affects Packet platform, developing something on bootkube would affect all platforms, so also the impact of that would be bigger. I'm not a fan of that, as I think installing `calico-host-protection` using bootkube is an abuse already, as bootkube should only be responsible for starting the controlplane.

ipochi:
> Also, this workaround only affects Packet platform, developing something on bootkube would affect all platforms, so also the impact of that would be bigger. I'm not a fan of that, as I think installing `calico-host-protection` using bootkube is an abuse already, as bootkube should only be responsible for starting the controlplane.

Ok, I hear you. So shouldn't our effort/ideas be around removing `calico-host-protection` from being installed as part of bootkube ?

We do want a default pre-installed list of components as part of `lokoctl cluster apply` why not try to work on that and `calico-host-protection` be the first one that gets installed for Packet.

invidian:
>Ok, I hear you. So shouldn't our effort/ideas be around removing calico-host-protection from being installed as part of bootkube ?

>We do want a default pre-installed list of components as part of lokoctl cluster apply why not try to work on that and calico-host-protection be the first one that gets installed for Packet.

Yes, this is something we should focus on mid-term.

Short term, I want to fix a stability issue on Packet, to offer best possible reliability we can, with this workaround PR, which fortunately is low impact.

assets/terraform-modules/packet/flatcar-linux/kubernetes/calico-host-protection.tf
```diff
@@ -1,3 +1,10 @@
+resource "local_file" "calico_crds" {
```
iaguis, line 1: Can we add a TODO comment here? Ideally pointing to an issue to fix this properly.

assets/terraform-modules/packet/flatcar-linux/kubernetes/ssh.tf
```diff
@@ -102,7 +103,8 @@ resource "null_resource" "bootkube-start" {
 
   provisioner "remote-exec" {
     inline = [
-      "sudo mv $HOME/assets /opt/bootkube",
+      "sudo mv $HOME/assets /opt/bootkube/",
+      "sudo mkdir /opt/bootkube/assets/manifests",
```
iaguis, line 107: Why is this needed now? Is `mkdir -p` better here just in case?

assets/terraform-modules/packet/flatcar-linux/kubernetes/calico-host-protection.tf
invidian: I'm not sure to what issue I should point this to, so I'll put a link to the old issue for now.

I think conceptually it's correct that this chart gets installed in bootkube, as it handles the firewall, which we want to have as soon as possible (BTW, it still doesn't provide 100% protection, as bootstrap kube-apiserver will be exposed to the internet).

In the future, we may have a closed cluster from the internet until components operator kicks in and provisions those policies, then it can be moved from bootkube.

assets/terraform-modules/packet/flatcar-linux/kubernetes/ssh.tf
invidian: This is needed, because `bootkube-start` does the following:
```
          [ -n "$(ls /opt/bootkube/assets/manifests-*/* 2>/dev/null)" ] && mv /opt/bootkube/assets/manifests-*/* /opt/bootkube/assets/manifests && rm -rf /opt/bootkube/assets/manifests-*
```

If this directory does not exist, it fails.

Added a comment about this.

assets/terraform-modules/packet/flatcar-linux/kubernetes/calico-host-protection.tf
```diff
@@ -1,3 +1,20 @@
+// TODO: Currently, bootkube installs all charts in parallel, which causes
+// calico-host-protection chart installation to fail ocasionally, as it
+// gets installed before Calico chart itself.
```
iaguis, line 3:
```suggestion
// gets installed before the Calico chart itself.
```

assets/terraform-modules/packet/flatcar-linux/kubernetes/calico-host-protection.tf
```diff
@@ -1,3 +1,20 @@
...
+//
+// To workaround  this issue, we make an extra copy of Calico CRDs into
+// bootkube manifests directory, as those will be applied
+// before all Helm charts. This ensures, that the CRDs are created before
+// calico-host-protection chart is installed.
```
iaguis, line 8:
```suggestion
// To workaround this issue, we make an extra copy of the Calico CRDs into
// the bootkube manifests directory, as those will be applied
// before all Helm charts. This ensures that the CRDs are created before
// the calico-host-protection chart is installed.
```

assets/terraform-modules/packet/flatcar-linux/kubernetes/ssh.tf
```diff
@@ -102,7 +103,12 @@ resource "null_resource" "bootkube-start" {
 
   provisioner "remote-exec" {
     inline = [
-      "sudo mv $HOME/assets /opt/bootkube",
+      "sudo mv $HOME/assets /opt/bootkube/",
+      # This is needed, as bootkube-start script will move all files matching
+      # /opt/bootkube/asssets/manifests-*/*.yaml into /opt/bootkube/assets/manifests.
+      #
+      # -p is not needed here, as we create assets directory in the command above.
```
iaguis, line 110: I think we can remove the `-p` explanation, it was just me missing the previous commands
```suggestion
```

assets/terraform-modules/packet/flatcar-linux/kubernetes/ssh.tf
iaguis: I guess my question was more "why was this working before"?

assets/terraform-modules/packet/flatcar-linux/kubernetes/ssh.tf
```diff
@@ -102,7 +103,12 @@ resource "null_resource" "bootkube-start" {
 
   provisioner "remote-exec" {
     inline = [
-      "sudo mv $HOME/assets /opt/bootkube",
+      "sudo mv $HOME/assets /opt/bootkube/",
+      # This is needed, as bootkube-start script will move all files matching
```
iaguis, line 107:
```suggestion
      # This is needed, as the bootkube-start script will move all files matching
```

assets/terraform-modules/packet/flatcar-linux/kubernetes/ssh.tf
invidian: Because `[ -n "$(ls /opt/bootkube/assets/manifests-*/* 2>/dev/null)" ]` returns error :) We do not use `manifests` directory anymore.

iaguis (approved):
lgtm

PR #1183, opened 2020-11-11
docs: adapt to our rendering system

iaguis:
This PR makes changes needed to adapt to our new rendering system so we can show documentation on the Kinvolk website.

The changes needed break the current internal links between the markdown files, so we should minimize the time between merging and having the docs up on the website.

TODO:

- [x] fix CLI reference generator to output the right headers and links


Makefile
```diff
@@ -154,6 +154,8 @@ docker-update-dependencies: docker-build
 .PHONY: docs
 docs:
 	GO111MODULE=on go run -mod=$(MOD) -buildmode=exe cli/cmd/document/main.go $(DOCS_DIR)
+	for f in $(DOCS_DIR)/lokoctl*md; do sed -i 's/\[\(lokoctl.*\)\](\(.*\)\.md)/\[\1\](\.\.\/\2)/g' $$f; done
```
invidian, line 157: I thought this could be done in Go, but it seems shell is easier :see_no_evil:

README.md
```diff
@@ -5,7 +5,7 @@ Kubernetes.
 It focuses on being minimal, easy to use, and secure by default.
 
 Lokomotive is fully self-hosted and includes [Lokomotive
-Components](docs/concepts/components.md) adding essential functionality for production
+Components](https://kinvolk.io/docs/lokomotive/0.5/concepts/components) adding essential functionality for production
```
invidian, line 8: Do we have to hardcode the version here? What do we do at 0.6 release? Where should commits go between the releases? I think we should have developer documentation describing new process.

README.md
```diff
@@ -26,28 +26,35 @@ not shipped with upstream Kubernetes.
 ## Installation
 
 Lokomotive provides the lokoctl CLI tool to manage clusters.
-Check the [installation guide](docs/installer/lokoctl.md) to install it.
+Check the [installation guide](https://kinvolk.io/docs/lokomotive/0.5/installer/lokoctl) to install it.
```
invidian, line 29: Why we can't use relative links BTW? I think it would me MUCH better. Also, if we run in preview, then links will be wrong as well?

README.md
```diff
@@ -65,7 +72,7 @@ have it assigned to yourself.
 
 ## Contributing
 
-Check out our [contributing guidelines](docs/CONTRIBUTING.md).
+Check out our [contributing guidelines](https://kinvolk.io/docs/lokomotive/0.5/contributing).
```
invidian, line 75: Maybe we can have some global contributing page then?

docs/CONTRIBUTING.md
```diff
@@ -1,16 +1,7 @@
-# Contribution Guidelines
-
-## Contents
-
-- [Code of Conduct](#code-of-conduct)
```
invidian, line 5: Hm, having a ToC in markdown files is useful when browsing via GitHub.

docs/configuration-reference/components/prometheus-operator.md
```diff
@@ -19,13 +13,13 @@ instances.
 
 * A Lokomotive cluster with a
   [PersistentVolume](https://kubernetes.io/docs/concepts/storage/persistent-volumes/) plugin, e.g.
-  [OpenEBS](openebs-operator.md) or one of the
+  [OpenEBS](../openebs-operator) or one of the
   [built-in](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#types-of-persistent-volumes)
   plugins.
 
-* An ingress controller such as [Contour](contour.md) for HTTP ingress.
+* An ingress controller such as [Contour](../contour) for HTTP ingress.
```
invidian, line 20: This change breaks links in GitHub rendered markdown. How can we preview the changes to ensure things are not broken?

invidian (changes requested):
Some questions/concerns which I would like to get addressed before we merge this.

README.md
joaquimrocha: We will have a redirect from `https://kinvolk.io/docs/lokomotive/latest/concepts/components` -> `https://kinvolk.io/docs/lokomotive/0.5/concepts/components`. So all these links should have latest instead of 0.5.

README.md
joaquimrocha: The README should link to the rendered docs. So this is fine here. Otherwise you'll see the markdown files and Github is nice enough but in the current format of docs, the docs portal will be better.

README.md
joaquimrocha: We have a global cointributon guide, but each project may have its own pages like this. It's okay.

docs/CONTRIBUTING.md
joaquimrocha: But the new docs rendered creates this by default. Unfortunately we cannot have both.

docs/configuration-reference/components/prometheus-operator.md
joaquimrocha: The idea with the docs portal is that it will be used primarily for reading docs from now on. Unfortunately the links don't use `.md` in the portal.

invidian: Okay, so as a contributor to the docs, how I can see that my changes looks correct?

vbatts: this is a gap between hugo rendering and github linking. 
As for previewing the rendered docs, i'm sure we'll smooth it out, or even better, have a canary netflify deployment for review in the github bots

iaguis: Yes, I think we should add a GitHub check in each PR linking to a netlify preview. We still need some work to get that set up though.

invidian: Even manual instruction how to test them would be sufficient for now, but I think it should be documented.

README.md
iaguis: I've changed them to latest

docs/configuration-reference/components/prometheus-operator.md
iaguis: As discussed OOB, we've agreed on a way to test the final rendering internally. Contributors can work against the raw Markdown and we'll check if the rendering is correct, we expect most contributions won't need to check the final rendering. We'll later follow up with a way for contributors to see the rendered docs.

invidian (dismissed):
OK

invidian (approved):

PR #1201, opened 2020-11-19
Implement automatic certificate rotation

### LINKED ISSUE #1215: Implement subcommand for certificate rotation and document it's usage
To track work on #1201 and to superseed #1031, we should implement a separate subcommand, which will:
- schedule all cluster certificates for renewal
- re-create certificates using Terraform
- deploy new certificates to the cluster
- ensure that all cluster-components are restarted to pick up new certificates

Additionally, we should tell user, what are the manual steps after certificate rotation. I think doing a rolling reboot on all nodes in the cluster (perhaps using FLUO) should be sufficient, as it will re-create/restart all the pods, which should be enough to propagate new certificates.

invidian:
Draft on top of #1193

Closes #1215 

TODO:
- [ ] Convert from running on every `cluster apply` to separate sub-command
- [ ] Perhaps don't run updates when rotating the certificates.
- [ ] Figure out how to tell user what's the expiration time of certificates.
- [ ] Add documentation for certificate rotation process, which should mention that ideally all nodes in the cluster should be drained and rebooted after rotation to ensure that all pods using CA certificate etc got restarted.
- [ ] Figure out automated restart of kube-proxy, Calico-node, CoreDNS and other core components.

### DIFF
--- a/assets/charts/control-plane/kube-apiserver/templates/kube-apiserver-ds.yaml
+++ b/assets/charts/control-plane/kube-apiserver/templates/kube-apiserver-ds.yaml
@@ -24,6 +24,8 @@ spec:
       annotations:
         checkpointer.alpha.coreos.com/checkpoint: "true"
         seccomp.security.alpha.kubernetes.io/pod: 'docker/default'
+        # Automatically rolls update when secret changes.
+        checksum/secret: {{ include (print $.Template.BasePath "/kube-apiserver-secret.yaml") . | sha256sum }}
     spec:
 {{- template "containers" . }}
 {{- end }}
--- a/assets/charts/control-plane/kube-apiserver/templates/kube-apiserver.yaml
+++ b/assets/charts/control-plane/kube-apiserver/templates/kube-apiserver.yaml
@@ -25,6 +25,8 @@ spec:
       annotations:
         checkpointer.alpha.coreos.com/checkpoint: "true"
         seccomp.security.alpha.kubernetes.io/pod: 'docker/default'
+        # Automatically rolls update when secret changes.
+        checksum/secret: {{ include (print $.Template.BasePath "/kube-apiserver-secret.yaml") . | sha256sum }}
     spec:
 {{- template "containers" . }}
 {{- end }}
--- a/assets/charts/control-plane/kubelet/templates/kubelet-ds.yaml
+++ b/assets/charts/control-plane/kubelet/templates/kubelet-ds.yaml
@@ -17,7 +17,23 @@ spec:
         tier: node
         k8s-app: kubelet
     spec:
-      automountServiceAccountToken: false
+      initContainers:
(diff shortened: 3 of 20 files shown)


meyskens (commented):
Leaving some thoughts mainly for myself when picking this up.

I have one main concern which is that is the CA rotation. Rotating the CA by replacing it means the cluster will be non-responsive quite soon when kubelets pick up the new CA but the API i the old (and vice versa) that will no longer be able to comminucate. 
To solve this we should be able to distribute an historic CA chain (just 1 old one will be enough) as trust bundle. Followed by making it trusted by all worker nodes then restarting the API server. 

Etcd certs should follow a similar strategy.

invidian:
>I have one main concern which is that is the CA rotation. Rotating the CA by replacing it means the cluster will be non-responsive quite soon when kubelets pick up the new CA but the API i the old (and vice versa) that will no longer be able to comminucate.
To solve this we should be able to distribute an historic CA chain (just 1 old one will be enough) as trust bundle. Followed by making it trusted by all worker nodes then restarting the API server.

The situation you described only occurs when you replace _private key_ of the CA certificate. We only rotate the certificate itself, so the signatures remains the same, so the communication remains functional. This is a deliberate decision as first step, as changing only the certificates is relatively easy. Changing private keys is indeed harder and remains chaining and further cleanup of old CA certificates etc.

meyskens:
I over read that we didn't taint the keys so that indeed helps a lot, in the first iteration of this. 
I also looked a bit it my kubeadm built cluster (just by coincidence my admin user cert expired) and i confirmed my suspicion that the CA certs kubadm generates are valid for 10 years and client only for 1/2 years (1.5 i guess, need to look into the code just to be sure) which brings me to another thing i wanted to discuss which is having a longer lived CA duration in Lokomtive, but that of course also depends on how we'll handle the certificate rollovers.

meyskens:
Also sorry I am throwing a lot of things here :) Some might just be better as a follow up PR in a new version ;)

invidian:
>I also looked a bit it my kubeadm built cluster (just by coincidence my admin user cert expired) and i confirmed my suspicion that the CA certs kubadm generates are valid for 10 years and client only for 1/2 years (1.5 i guess, need to look into the code just to be sure) which brings me to another thing i wanted to discuss which is having a longer lived CA duration in Lokomtive, but that of course also depends on how we'll handle the certificate rollovers.

We already have CA expiration time configurable, so user may choose to extend it, but I think by default we should offer short-lived certificates and, ideally, fully automated rotation.

meyskens:
> We already have CA expiration time configurable

I only find one certificate lifetime variable used everywhere, am I missing a variable?

invidian:
Ah yes, we only have `certs_validity_period_hours`. Hmm...

invidian:
Closing in favor of #1435.

PR #1207, opened 2020-11-23
bootstrap-secrets: always quote values

invidian:
They should always be strings. If Terraform generates them to be
numbers-only, Helm will fail installing it.


surajssd (approved):
LGTM

PR #1238, opened 2020-12-03
tinkerbell: fix printing bootkube logs on error

invidian:
Instead of piping everything to stderr we should use --no-pager to get
logs printed correctly.


vbatts (approved):
LGTM

Issue #1096, opened 2020-10-19
configmaps "extension-apiserver-authentication" is forbidden


johananl:
kube-scheduler constantly logs the following:

```
E1019 11:57:21.407120       1 reflector.go:127] k8s.io/apiserver/pkg/server/dynamiccertificates/configmap_cafile_content.go:206: Failed to watch *v1.ConfigMap: failed to list *v1.ConfigMap: configmaps "extension-apiserver-authentication" is forbidden: User "system:serviceaccount:kube-system:kube-scheduler" cannot list resource "configmaps" in API group "" in the namespace "kube-system"
E1019 11:57:57.428130       1 reflector.go:127] k8s.io/apiserver/pkg/server/dynamiccertificates/configmap_cafile_content.go:206: Failed to watch *v1.ConfigMap: failed to list *v1.ConfigMap: configmaps "extension-apiserver-authentication" is forbidden: User "system:serviceaccount:kube-system:kube-scheduler" cannot list resource "configmaps" in API group "" in the namespace "kube-system"
E1019 11:58:31.721129       1 reflector.go:127] k8s.io/apiserver/pkg/server/dynamiccertificates/configmap_cafile_content.go:206: Failed to watch *v1.ConfigMap: failed to list *v1.ConfigMap: configmaps "extension-apiserver-authentication" is forbidden: User "system:serviceaccount:kube-system:kube-scheduler" cannot list resource "configmaps" in API group "" in the namespace "kube-system"
E1019 11:59:20.845730       1 reflector.go:127] k8s.io/apiserver/pkg/server/dynamiccertificates/configmap_cafile_content.go:206: Failed to watch *v1.ConfigMap: failed to list *v1.ConfigMap: configmaps "extension-apiserver-authentication" is forbidden: User "system:serviceaccount:kube-system:kube-scheduler" cannot list resource "configmaps" in API group "" in the namespace "kube-system"
```

I'm not sure if this is an indication for a real problem, but anyway we probably want to either fix the problem or eliminate the unnecessary logging.

Observed with lokoctl `v0.4.1-207-gd814fcc0` on Packet.

invidian:
Perhaps `extension-apiserver-authentication-reader` role should be assigned.

niels-s:
I wanted to report the same issue, it's producing quite some logs, and I wonder if this means something is broken?

invidian:
[...]

PR #1254, opened 2020-12-07
kube-scheduler: fix access to cluster certificates ConfigMap

knrt10:
added a rolebinding for role `extension-apiserver-authentication-reader` for kube-scheduler

closes: #1096

### DIFF
--- a/assets/charts/control-plane/kubernetes/templates/kube-scheduler-role-binding.yaml
+++ b/assets/charts/control-plane/kubernetes/templates/kube-scheduler-role-binding.yaml
@@ -10,3 +10,16 @@ subjects:
 - kind: ServiceAccount
   name: kube-scheduler
   namespace: kube-system
+---
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: kube-scheduler-extension-apiserver-authentication-reader
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: extension-apiserver-authentication-reader
+subjects:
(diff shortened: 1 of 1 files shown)


assets/charts/control-plane/kubernetes/templates/kube-scheduler-role-binding.yaml
```diff
@@ -10,3 +10,16 @@ subjects:
 - kind: ServiceAccount
   name: kube-scheduler
   namespace: kube-system
+---
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: system:configmaps-reader
```
invidian, line 17: Name should probably be unique. Like `kube-scheduler-extension-apiserver-authentication-reader` or something.

invidian (changes requested):
Please add an explanation why we add it in the commit message.

invidian (changes requested):
How about the following for the commit message?
```
kube-scheduler: fix access to cluster certificates ConfigMap

Currently kube-scheduler is configured to dynamically load cluster certificates from
extension-apiserver-authentication-reader ConfigMap, but it is not authorized to do that
because its ServiceAccount is missing extension-apiserver-authentication-reader Role.

This causes kube-scheduler to not being able to read new cluster certificates dynamically
and makes it print annoying error message.

This commit adds missing RoleBinding for Role extension-apiserver-authentication-reader to
kube-scheduler ServiceAccount to fix this.

Closes: #1096

```

invidian (approved):

surajssd (approved):

PR #1259, opened 2020-12-08
Simplify docs link titles

### LINKED ISSUE #1217: Make docs more content aware
Currently when you go to https://kinvolk.io/docs/lokomotive/0.5/quickstarts/aks/, you can see "Lokomotive AKS quickstart guide" string all over the place.
Selection_281
Highlighted parts should be simplified to just "AKS". As far as I know this can be done with the following patch:
```diff
 ---
 title: Lokomotive AKS quickstart guide
+linkTitle: AKS
 weight: 10
 ---
```

We should of course go trough the entire docs tree and apply it everywhere where content is duplicated.

kskewes:
We are already nested under a heading so we can drop that heading (eg: "quickstart..." or "..... configuration reference") from the link title.

Closes #1217 

Ran `make docs` in root of repository but didn't notice any side effects or changes.
I'm not sure how to locally validate how the docs render.
I was expecting I could do something like `hugo server` or whatever the mechanism is.

### DIFF
--- a/docs/configuration-reference/components/cert-manager.md
+++ b/docs/configuration-reference/components/cert-manager.md
@@ -1,5 +1,6 @@
 ---
 title: Cert-Manager configuration reference for Lokomotive
+linkTitle: Cert-Manager
 weight: 10
 ---
 
--- a/docs/configuration-reference/components/flatcar-linux-update-operator.md
+++ b/docs/configuration-reference/components/flatcar-linux-update-operator.md
@@ -1,5 +1,6 @@
 ---
-title: Flatcar Container Linux update operator configuration reference for Lokomotive
+title: Flatcar Container Linux Update Operator configuration reference for Lokomotive
+linkTitle: Flatcar Container Linux Update Operator
 weight: 10
 ---
 
--- a/docs/configuration-reference/backend/local.md
+++ b/docs/configuration-reference/backend/local.md
@@ -1,5 +1,6 @@
 ---
 title: Local backend configuration reference
+linkTitle: Local
 weight: 10
 ---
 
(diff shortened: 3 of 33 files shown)


docs/configuration-reference/components/cert-manager.md
```diff
@@ -1,5 +1,6 @@
 ---
 title: Cert-Manager configuration reference for Lokomotive
+linkTitle: Cert-Manager
```
invidian, line 3: While we're on it, maybe we can change the name to `cert-manager`? This is official name of the project as far as I know.

docs/configuration-reference/components/flatcar-linux-update-operator.md
```diff
@@ -1,5 +1,6 @@
 ---
-title: Flatcar Container Linux update operator configuration reference for Lokomotive
+title: Flatcar Container Linux Update Operator configuration reference for Lokomotive
+linkTitle: Flatcar Container Linux Update Operator
```
invidian, line 3: Hmm, same here. It should be `Flatcar Linux Update Operator`.

invidian (approved):
Thanks a lot for another PR @kskewes! Changes looks good to me, but perhaps someone else needs to look into this PR as well, to make sure this is what we want, as @blixtra had some concerns. Maybe @iaguis?

>Ran make docs in root of repository but didn't notice any side effects or changes.
I'm not sure how to locally validate how the docs render.
I was expecting I could do something like hugo server or whatever the mechanism is.

Yeah, `make docs` is only used right now to generate CLI reference documentation.

At the moment we have no way of locally previewing this unfortunately, though I'm sure @joaquimrocha is working on it :wink:

docs/configuration-reference/components/cert-manager.md
kskewes: Sure, done.

docs/configuration-reference/components/flatcar-linux-update-operator.md
kskewes: Sure, done.

PR #1274, opened 2020-12-18
calico-host-protection: add psp rule

knrt10:
Calico host endpoint controller only needs to talk to the apiserver and
should not be granted privileged PSP.

Update deprecated apiVersion for clusterrole and clusterrolebinding.

closes: #287

### DIFF
--- a/assets/charts/control-plane/calico-host-protection/templates/host-endpoint-controller.yaml
+++ b/assets/charts/control-plane/calico-host-protection/templates/host-endpoint-controller.yaml
@@ -47,7 +47,7 @@ spec:
 
 ---
 # rbac.yaml
-apiVersion: rbac.authorization.k8s.io/v1beta1
+apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
   name: calico-hostendpoint-controller-role
@@ -58,19 +58,19 @@ rules:
 - apiGroups: ["crd.projectcalico.org"]
   resources: ["hostendpoints"]
   verbs:
-   - create
-   - get
-   - list
-   - update
-   - delete
-   # To use kubectl apply on resources that already exist
-   - patch
+  - create
+  - get
+  - list
+  - update
+  - delete
+  # To use kubectl apply on resources that already exist
+  - patch
 - apiGroups: ["policy"]
   resources: ["podsecuritypolicies"]
   verbs: ["use"]
   resourceNames: ["calico-hostendpoint-controller-psp"]
 ---
-apiVersion: rbac.authorization.k8s.io/v1beta1
+apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
   name: calico-hostendpoint-controller-role-binding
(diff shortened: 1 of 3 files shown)


knrt10:
`kc auth can-i --list --as="system:serviceaccount:kube-system:calico-hostendpoint-controller"` has both 

```
podsecuritypolicies.policy                      []                  [zz-minimal]      [use]
podsecuritypolicies.policy                      []                  [zz-privileged]   [use]
```

listed as we have set privileged policy for all `Serviceaccounts` in `kube-system` namespace. 

Pod created using `calico-hostendpoint-controller`, still does not use `zz-minimal` policy. 

Trying to debug why this is happening

assets/charts/control-plane/calico-host-protection/templates/psp.yaml
```diff
@@ -0,0 +1,19 @@
...
+metadata:
+  annotations:
+    seccomp.security.alpha.kubernetes.io/allowedProfileNames: '*'
+  name: calico-hostendpoint-controller-psp
+spec:
+  privileged: false
+  fsGroup:
+    rule: RunAsAny
+  runAsUser:
+    rule: RunAsAny
```
invidian, line 12: Why it is allowed to run as any if we run it as non-root?

assets/charts/control-plane/calico-host-protection/templates/psp.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: policy/v1beta1
+kind: PodSecurityPolicy
```
invidian, line 2: I think we could make this policy more strict. Start with restricted policy and then allow what's required.

assets/charts/control-plane/calico-host-protection/templates/host-endpoint-controller.yaml
```diff
@@ -43,7 +48,7 @@ spec:
 
 ---
 # rbac.yaml
-apiVersion: rbac.authorization.k8s.io/v1beta1
+apiVersion: rbac.authorization.k8s.io/v1
```
invidian, line 51: Version bump could be done in separate commit.

assets/charts/control-plane/calico-host-protection/templates/host-endpoint-controller.yaml
```diff
@@ -33,6 +33,11 @@ spec:
       containers:
       - image: quay.io/kinvolk/calico-hostendpoint-controller:v0.0.4
         name: calico-hostendpoint-controller
+        securityContext:
+          privileged: false
```
invidian, line 37: Are all those statements required? Shouldn't just `runAsUser: 65534` be sufficient to make it not run as root?

test/platform/packet/packet_test.go
```diff
@@ -98,3 +98,22 @@ func TestWhenBGPIsNotDisabledInConfigurationServersHasBGPSessionCreated(t *testi
 		t.Fatalf("Worker pool with BGP not disabled should have at least one BGP session")
 	}
 }
+
+func TestAnnotationForCalicoHostEndpointController(t *testing.T) {
```
invidian, line 102:
```suggestion
func TestCalicoHostEndpointControllerRunsWithDedicatedPSP(t *testing.T) {
```

invidian (changes requested):

assets/charts/control-plane/calico-host-protection/templates/psp.yaml
knrt10: Fixed

invidian (approved):

surajssd (approved):
LGTM

PR #1306, opened 2021-01-08
AKS: Don't check control plane components

knrt10:
AKS does not need to ensure checking of control components as we mark it
as managed service.

closes: #1303

### DIFF
--- a/cli/cmd/cluster/apply.go
+++ b/cli/cmd/cluster/apply.go
@@ -74,7 +74,7 @@ func Apply(contextLogger *log.Entry, options ApplyOptions) error {
 
 	charts := platform.CommonControlPlaneCharts(options.UpgradeKubelets)
 
-	if !exists {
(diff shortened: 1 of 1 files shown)


ipochi (dismissed):
lgtm

cli/cmd/cluster/apply.go
```diff
@@ -93,7 +93,7 @@ func Apply(contextLogger *log.Entry, options ApplyOptions) error {
 		}
 	}
 
-	if exists && !options.SkipPreUpdateHealthCheck {
+	if exists && !options.SkipPreUpdateHealthCheck && !c.platform.Meta().Managed {
```
invidian, line 96: Hm, we don't need unpacking controlplane charts either I think.

surajssd (approved):

invidian (approved):

Issue #1300, opened 2021-01-05
Add support VMWare provider


mazzy89:
**Current situation**

Support VMware as a provider.

**Ideal future situation**

Attract a high number of customers/users.

**Additional information**

I would be interested to have VMware implemented as a provider. I have access to the VMWare platform so it would be easy for me.

invidian:
Hey @mazzy89, thanks for opening the issue. Could this be implemented using https://registry.terraform.io/providers/hashicorp/vsphere/latest/docs ?

mazzy89:
Hey Mateusz. yes the idea is right to use the VSphere provider and create VMs.

mazzy89:
Could you please Mateusz clarify one thing. I've seen around that you would like to use the TF module `controller` and `worker` to bring reusability. In case of the creation of a new platform like this case, do you recommend to use these two modules?

invidian:
>Could you please Mateusz clarify one thing. I've seen around that you would like to use the TF module controller and worker to bring reusability. In case of the creation of a new platform like this case, do you recommend to use these two modules?

Yes, using those modules would be preferred approach. Given that they are only used for one platform so far, they may require some changes to make them reusable. But using them should save some hassle implementing new platform.

mazzy89:
Ok. I will start then to put down the PR

mazzy89:
@invidian any recommendations before putting down the PR similar to the one discussed above?

invidian:
Hmm, perhaps you can look at 84821fd4e50aa3fe55e7e17b192affda8d25aa10 to get an idea what is required to get a full platform added, so Terraform, Go code and some tests. The documentation and quickstart guides can be done separately I think.

Also if you could document how to test this PR, that would simplify things a lot, as we will have to set up some CI for it later. The CI configs are also optional, though appreciated.

If you need to do some changes unrelated to new platform, for example in common modules, please commit them separately or even open a separate PR with those changes so we keep PRs reasonable small and focused.

Hope this helps. Let me know if you have more questions :)

mazzy89:
> Also if you could document how to test this PR, that would simplify things a lot, as we will have to set up some CI for it later. The CI configs are also optional, though appreciated.

Do you have guys a VMWare account in which to run integration tests? Or you mean unit tests?

invidian:
>Do you have guys a VMWare account in which to run integration tests? Or you mean unit tests?

[...]

Issue #1121, opened 2020-10-22
Scale Masters after cluster creation


justgotthis:
Currently I was planning to creating my lokomotive cluster with one master as my usage will be low and to save on costs. As my team usage increases, I would have liked to scale masters up to 3. Currently that is not possible without recreating the cluster again. This would be a useful feature to have and save on time. By running "cluster apply" and the fact that it is idempotent, it would just spin up 2 other servers instead of having to recreate the cluster from scratch.

invidian:
Hey @justgotthis, thanks for reporting. This is a known issue (#29), but I think we can keep this one open as a feature request and the other one as bug.

justgotthis:
Sounds good. Thanks for the heads up Mateusz.

*- Luis L.*

Issue #29, opened 2020-02-20
Adding/removing controller nodes to existing cluster does not work


invidian:
It seems that new etcd member is not able to join the cluster.

However, if you create 3-controller node cluster from scratch and then you need to replace one one (so the data from the node is lost), everything is fine.

invidian:
As a workaround, we could ignore changes to the controllers count parameter.

PR #1311, opened 2021-01-11
Add VMware VSphere platform

mazzy89:
# Add VMware VSphere platform

This PR adds the initial support for the VMware VSphere platform. The idea is to provide support for both VSphere and ESXi hosts as initially discussed here https://github.com/kinvolk/lokomotive/issues/1300 so to run this k8s distro also on ESXi with a free license.

# How to use

You need to first have in VCenter the Flatcar VM Template available. Download from [Flatcar Releases](https://stable.release.flatcar-linux.net/) the OVA required and then run the following command:

```sh
ovftool --name="Flatcar-stable-2605.10.0" --datastore='<datastore>' --overwrite --skipManifestCheck --noSSLVerify --allowAllExtraConfig --importAsTemplate ./flatcar_production_vmware_ova.ova 'vi://<username>@<server>/<dataceter>/host/<compute_cluster>'
```

This will import the OVA as a VM Template which will be used by Terraform provider to create all the VMware instances.

It has been provided a testing `cluster.lokocfg` file to test it. For VSphere, it is important to export environment variables needed for the Terraform `vsphere` provider and provide ` lokocfg.vars` file.

VMware instances require static IP addresses. 

```sh
export VSPHERE_USER=
export VSPHERE_PASSWORD=
export VSPHERE_SERVER=
# In the case you run VCenter behind a self-signed cert
export VSPHERE_ALLOW_UNVERIFIED_SSL=true 
lokoctl cluster apply --lokocfg ./examples/vmware-testing/cluster.lokocfg --lokocfg-vars ./examples/vmware-testing/lokocfg.vars
```

It is important to set up also the DNS provider as one of `route53`, `cloudflare` or `manual`.

# Testing done

Created a testing cluster starting from the provided `cluster.lokocfg`. Tested with `Cloudflare` as provider and `manual`.

------

# TODO

- [x] Write unit tests for VMware platform
- [x] Configure properly `bootkube`

### DIFF
--- a/assets/terraform-modules/platforms/vmware/controllers/main.tf
+++ b/assets/terraform-modules/platforms/vmware/controllers/main.tf
@@ -87,7 +87,7 @@ resource "vsphere_virtual_machine" "main" {
   }
 
   // Advanced options
-  nested_hv_enabled  = var.nested_hv_enabled
+  nested_hv_enabled = var.nested_hv_enabled
 
   depends_on = [
     vsphere_folder.main
--- a/assets/terraform-modules/platforms/vmware/controllers/main.tf
+++ b/assets/terraform-modules/platforms/vmware/controllers/main.tf
@@ -1,0 +1,98 @@
+locals {
+  nodes_mask = split("/", var.hosts_cidr)[1]
+  nodes_gw   = cidrhost(var.hosts_cidr, 1)
+}
+
+module "controller" {
+  source = "../../../controller"
+
+  count = var.node_count
+
+  cluster_name           = var.cluster_name
+  controllers_count      = var.node_count
+  dns_zone               = var.dns_zone
+  count_index            = count.index
+  cluster_dns_service_ip = module.bootkube.cluster_dns_service_ip
+  ssh_keys               = var.ssh_keys
+  cluster_domain_suffix  = var.cluster_domain_suffix
+  host_dns_ip            = var.host_dns_ip
+  apiserver              = format("%s.%s", var.cluster_name, var.dns_zone)
+  ca_cert                = module.bootkube.ca_cert
+
+  clc_snippets = concat(var.clc_snippets, [
+    <<EOF
+storage:
+  files:
+  - path: /etc/hostname
+    filesystem: root
+    mode: 0644
+    contents:
+      inline: |
+        ${var.cluster_name}-controller-${count.index}
+EOF
+    ,
+    <<EOF
+networkd:
+  units:
+  - name: 00-ens192.network
+    contents: |
+      [Match]
+      Name=ens192
+
+      [Network]
+      Address=${var.nodes_ips[count.index]}/${local.nodes_mask}
+      Gateway=${local.nodes_gw}
+      %{~for dns in var.dns_servers~}
+      DNS=${dns}
+      %{~endfor~}
+EOF
+    ,
+    ],
+  )
+}
+
+resource "vsphere_folder" "main" {
+  path          = var.folder
+  type          = "vm"
+  datacenter_id = data.vsphere_datacenter.main.id
+}
+
+resource "vsphere_virtual_machine" "main" {
+  count = var.node_count
+
+  name             = format("%s-controller-%s", var.cluster_name, count.index)
+  resource_pool_id = data.vsphere_compute_cluster.main.resource_pool_id
+  datastore_id     = data.vsphere_datastore.main.id
+  folder           = var.folder
+
+  num_cpus = var.cpus_count
+  memory   = var.memory
+
+  # https://code.vmware.com/apis/358/doc/vim.vm.GuestOsDescriptor.GuestOsIdentifier.html
+  #
+  guest_id = "other3xLinux64Guest"
+
+  network_interface {
+    network_id = data.vsphere_network.main.id
+  }
+
+  disk {
+    label            = "disk0"
+    size             = var.disk_size
+    eagerly_scrub    = data.vsphere_virtual_machine.main_template.disks[0].eagerly_scrub
+    thin_provisioned = data.vsphere_virtual_machine.main_template.disks[0].thin_provisioned
+  }
+
+  clone {
+    template_uuid = data.vsphere_virtual_machine.main_template.id
+  }
+
+  extra_config = {
+    "guestinfo.ignition.config.data.encoding" = "base64"
+    "guestinfo.ignition.config.data"          = base64encode(module.controller[count.index].clc_config)
+  }
+
+  depends_on = [
+    vsphere_folder.main
+  ]
+}
--- a/assets/terraform-modules/platforms/vmware/controllers/main.tf
+++ b/assets/terraform-modules/platforms/vmware/controllers/main.tf
@@ -89,6 +89,12 @@ resource "vsphere_virtual_machine" "main" {
   // Advanced options
   nested_hv_enabled = var.nested_hv_enabled
 
+  lifecycle {
+    ignore_changes = [
+      extra_config["guestinfo.ignition.config.data"]
+    ]
+  }
+
   depends_on = [
     vsphere_folder.main
   ]
--- a/assets/terraform-modules/platforms/vmware/controllers/main.tf
+++ b/assets/terraform-modules/platforms/vmware/controllers/main.tf
@@ -42,9 +42,6 @@ networkd:
       [Network]
       Address=${var.nodes_ips[count.index]}/${local.nodes_mask}
       Gateway=${local.nodes_gw}
-      %{~for dns in var.dns_servers~}
-      DNS=${dns}
-      %{~endfor~}
 EOF
     ,
     ],
@@ -67,10 +64,7 @@ resource "vsphere_virtual_machine" "main" {
 
   num_cpus = var.cpus_count
   memory   = var.memory
-
-  # https://code.vmware.com/apis/358/doc/vim.vm.GuestOsDescriptor.GuestOsIdentifier.html
-  #
-  guest_id = "other3xLinux64Guest"
+  guest_id = data.vsphere_virtual_machine.main_template.guest_id
 
   network_interface {
     network_id = data.vsphere_network.main.id
--- a/assets/terraform-modules/platforms/vmware/controllers/main.tf
+++ b/assets/terraform-modules/platforms/vmware/controllers/main.tf
@@ -86,6 +86,9 @@ resource "vsphere_virtual_machine" "main" {
     "guestinfo.ignition.config.data"          = base64encode(module.controller[count.index].clc_config)
   }
 
+  // Advanced options
+  nested_hv_enabled  = var.nested_hv_enabled
+
   depends_on = [
     vsphere_folder.main
   ]
--- a/assets/terraform-modules/platforms/vmware/controllers/bootkube.tf
+++ b/assets/terraform-modules/platforms/vmware/controllers/bootkube.tf
@@ -1,0 +1,30 @@
+locals {
+  api_servers = format("%s-private.%s", var.cluster_name, var.dns_zone)
+}
+
+# Self-hosted Kubernetes assets (kubeconfig, manifests).
+module "bootkube" {
+  source = "../../../bootkube"
+
+  cluster_name = var.cluster_name
+
+  api_servers          = [local.api_servers]
+  api_servers_external = [format("%s.%s", var.cluster_name, var.dns_zone)]
+  api_servers_ips      = var.nodes_ips
+  etcd_servers         = module.controller[0].etcd_servers
+
+  asset_dir             = var.asset_dir
+  network_mtu           = var.network_mtu
+  pod_cidr              = var.pod_cidr
+  service_cidr          = var.service_cidr
+  cluster_domain_suffix = var.cluster_domain_suffix
+  enable_reporting      = var.enable_reporting
+  enable_aggregation    = var.enable_aggregation
+
+  certs_validity_period_hours = var.certs_validity_period_hours
+
+  bootstrap_tokens            = concat(module.controller.*.bootstrap_token, var.worker_bootstrap_tokens)
+  enable_tls_bootstrap        = true
+  disable_self_hosted_kubelet = false
+  conntrack_max_per_core      = var.conntrack_max_per_core
+}
--- a/assets/terraform-modules/platforms/vmware/controllers/data.tf
+++ b/assets/terraform-modules/platforms/vmware/controllers/data.tf
@@ -1,0 +1,23 @@
+data "vsphere_datacenter" "main" {
+  name = var.datacenter
+}
+
+data "vsphere_datastore" "main" {
+  name          = var.datastore
+  datacenter_id = data.vsphere_datacenter.main.id
+}
+
+data "vsphere_compute_cluster" "main" {
+  name          = var.compute_cluster
+  datacenter_id = data.vsphere_datacenter.main.id
+}
+
+data "vsphere_network" "main" {
+  name          = var.network
+  datacenter_id = data.vsphere_datacenter.main.id
+}
+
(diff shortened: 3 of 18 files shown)


mazzy89:
@invidian I've wired here the first iteration for the VMware support.

I've been wondering which would be the best approach here for the DNS. Would we go like with Packet where we support some external providers or like in Bare Metal?

Which is your idea?

invidian:
Hey @mazzy89, thanks for opening the PR! Sorry I wasn't able to get back to you in the #1300 in timely manner. I will have time to have a deeper look next week.

About the DNS, given that vSphere does not offer any DNS solution out of the box (e.g. libvirt does), I think we could still follow the same patterns as Packet platform does. Our provider-agnostic DNS implementation has also `manual` provider, which can be used the same as with bare-metal platform. While not perfect, I think this will make it easier to try out the platform.

>VMware installations usually run in private networks not exposed to the Internet.

This shouldn't be a problem, as it's okay'ish to put private IPs in public DNS records and nodes must most likely have internet access, for example to pull the images.

mazzy89:
Hey, @mazzy89 no worries. Thank you to get back here.


I agree with you with the DNS. I will then follow the Packet way as it is already now and I will tune up this PR.

pothos:
FYI:
For a static IP address setup via CLC snippets and a definition of `coreos-metadata.service` you can have a look here: https://kinvolk.io/docs/flatcar-container-linux/latest/installing/cloud/vmware/ (`coreos-metadata.service` must be created by Lokomotive because for VMware the network setup is custom, so Flatcar doesn't have a unit for VMware).

mazzy89:
Thank you @pothos 

I haven't looked at that yet. The idea would be to wire the static ip assigment feature in a next iteration of the provider.

What do you think? 

Should we introduce it in this PR?

pothos:
Depending on the configuration of the DHCP server there may be no guarantee that the DNS records you set up are valid after a reboot or expiry of the DHCP address lease. Using static IP addresses takes DNS out of the equation and you can refer to the API server by the IP address. Therefore, I think it makes sense to go with static addresses from the beginning. However, I haven't had a closer look and if you know better, then you can go another route, I just try to help.

mazzy89:
That makes totally sense.

If there is no component running in the cluster deployed by this distro that keep those records up to date then it makes more sense to go with the static IPs approach from the first shot.

In Kops for instance there was a controller namely protokube that would run in the cluster and keep DNS entries up to date. This is the best case for environment where instances are ephemeral like the cloud.

I haven't looked at it deeply here but seems there is no controller that would keep entries up to date in our case. 

I will then implement static IP in this PR. 

Please advice @pothos in case there is anything planned in the horizon that would come in handy for updating DNS entries.

invidian:
>Depending on the configuration of the DHCP server there may be no guarantee that the DNS records you set up are valid after a reboot or expiry of the DHCP address lease. Using static IP addresses takes DNS out of the equation and you can refer to the API server by the IP address. Therefore, I think it makes sense to go with static addresses from the beginning. However, I haven't had a closer look and if you know better, then you can go another route, I just try to help.

I'm not sure if using IP addresses directly make sense here. I'd rather put static host entires into `/etc/hosts` then, which can be done using CLC snippets.

mazzy89:
@invidian I think it would be fine. We would have VMs with static IPs and then `/etc/hosts` entries in the case the DNS is set to manual. In the other cases we would use the public DNS

mazzy89:
@invidian the build process here is complaining but I'm not sure why.

locally I'm running the command `make check-working-tree-clean` and it returns w/o issues

assets/terraform-modules/platforms/vmware/controllers/main.tf
```diff
@@ -1,6 +1,12 @@
 locals {
   nodes_mask = split("/", var.hosts_cidr)[1]
   nodes_gw   = cidrhost(var.hosts_cidr, 1)
+
+  clc_storage = templatefile("${path.module}/clc_storage.tmpl", {
+    dns_provider : var.dns_provider,
+    etcd_fqdn : { for i, d in var.nodes_ips : format("%s-etcd%d.%s.", var.cluster_name, i, var.dns_zone) => d }
```
mazzy89, line 7: @invidian these two are the entries we push into the `/etc/hosts` of the controller in case the DNS provider is set to `manual`

mazzy89:
@invidian still need to add unit tests for the VMware config.

The idea here is now to start testing it on my side and check how it comes up on VMware. Now the VMware machines (controllers and workers) are setup to use static IP addresses and create `/etc/hosts` entries in case the DNS provider is set to `manual`.

In case you have spare time please take an unofficial look so I can make any adjustments along the way. I will open for the review by the end of the week.

mazzy89:
I gave a quick round of testing this and unfortunately, the customization of the instance was not possible. I get this

image

Perl is missing on the guest instances.

The customization is needed to set up `linux_options` which is required in the case of Linux systems

```sh
Error: linux_options must exist in VM customization options for Linux operating systems

  on ../terraform-modules/platforms/vmware/controllers/main.tf line 33, in resource "vsphere_virtual_machine" "main":
  33: resource "vsphere_virtual_machine" "main" {
```

Any idea on this?

invidian:
>Any idea on this?

Can this customization be done with CLC snippets maybe?

mazzy89:
I would need then to try what happens when I remove the entire `customize` block from the Terraform resource and configure everything via CLC like setting up `systemd` network units. Would then the VMware instance pick up the IP address correctly? I will try and let you know,.

mazzy89:
@invidian unfortunately it does not work. In the case when the VMware instance is attached to a network with a DHCP server, the machine gets an IP address despite a networkd unit is set up for the main interface `eth0`. In this case, it is attached to a network w/o DHCP server then the VMware instance comes up w/o IPv4 address.

Not many ideas here for the moment. I believe the case of bootstrapping Flatcar in VMware with static IPs using Terraform is not supported so we can't do much here unless the VMware box can run VMware customization commands.

invidian:
@mazzy89 hmm, and what about public DNS then?

In the future, we may support e.g. PowerDNS, which could run locally in VMware network, so DNS entries could be created there.

mazzy89:
It's more a problem of setting up static IPs on the NIC rather than DNS.
 
Nevertheless, in case we would use a public DNS then we would ask as a requirement that the VMware instances would run in a network where a DHCP server is deployed so it can provide IP addresses to the instances.

invidian:
>>It's more a problem of setting up static IPs on the NIC rather than DNS.

> Nevertheless, in case we would use a public DNS then we would ask as a requirement that the VMware instances would run in a network where a DHCP server is deployed so it can provide IP addresses to the instances.

Ah, of course, you need static IP addresses to be able to create DNS entries :facepalm: I think I didn't wake up properly yet :smile:

And are you able to read the IP addresses of the machines using Terraform?

mazzy89:
> Ah, of course, you need static IP addresses to be able to create DNS entries 🤦 I think I didn't wake up properly yet 😄

no worries. It's Friday. all allowed 😇

> And are you able to read the IP addresses of the machines using Terraform?

In the PR I've implemented, the IP addresses are set as input variables and passed into the controllers and workers.

A `networkd` unit named `00-eth0.network` is then created as CLC snippet. I see that it is correctly parsed by Terraform and transpiled as Ignition. But no IP address is set to the interface.

invidian:
>A networkd unit named 00-eth0.network is then created as CLC snippet. I see that it is correctly parsed by Terraform and transpiled as Ignition. But no IP address is set to the interface.

Right. I looked at the code now. And the interface is for sure `eth0`? I don't have experience with VMware, but by looking at the code, it seems it should work for me. What does `networkctl` says about the interface on the created machine?

mazzy89:
> And the interface is for sure eth0?

Maybe I should wake properly too 😇 the interface was not named `eth0` but `ens192`. IPs are assigned correctly now.

mazzy89:
```sh
❯ kubectl --kubeconfig=$PWD/lokomotive-assets/cluster-assets/auth/kubeconfig get pods -nkube-system
NAME                                                     READY   STATUS    RESTARTS   AGE
calico-kube-controllers-855c8775f9-mxts6                 1/1     Running   0          6m21s
calico-node-8g4qg                                        1/1     Running   0          4m44s
calico-node-dnf2k                                        1/1     Running   0          6m16s
coredns-7d799bc4c8-kvmdd                                 1/1     Running   0          6m21s
kube-apiserver-5bc9dd4f8-5h76x                           1/1     Running   0          6m21s
kube-controller-manager-66cc9486cc-6llvf                 1/1     Running   0          6m21s
kube-proxy-s78ph                                         1/1     Running   0          6m16s
kube-proxy-xxj7v                                         1/1     Running   0          4m44s
kube-scheduler-845bddddfd-999wf                          1/1     Running   0          6m21s
kubelet-bcbgw                                            1/1     Running   0          4m44s
kubelet-xgw85                                            1/1     Running   0          6m11s
metrics-server-dfd577c9d-mlv8b                           1/1     Running   0          90s
pod-checkpointer-dcv4s                                   1/1     Running   0          4m44s
pod-checkpointer-dcv4s-lokomotive-cluster-controller-0   1/1     Running   0          3m11s
```

```sh
You can find the logs in "lokomotive-assets/terraform/logs/16421.log"

Your configurations are stored in ./lokomotive-assets

Now checking health and readiness of the cluster nodes ...

Node                                      Ready    Reason          Message

lokomotive-cluster-controller-0           True     KubeletReady    kubelet is posting ready status
lokomotive-cluster-worker-my-wp-name-0    True     KubeletReady    kubelet is posting ready status

Success - cluster is healthy and nodes are ready!
INFO[0459] Applying component configuration              args="[]" command="lokoctl cluster apply"
Applying component 'metrics-server'...
Successfully applied component 'metrics-server' configuration!
Applying component 'flatcar-linux-update-operator'...
Successfully applied component 'flatcar-linux-update-operator' configuration!
```

@invidian  I was able to get a cluster 🆙 and 🏃‍♂️ 💥

Still missing units tests for the config and then ready for review.

Some notes of the current implementation:

when the DNS provider is set to `manual`, entries in `/etc/hosts` will be created.

It has been necessary to set the `bootkube` variable to

```hcl
  api_servers_external = var.nodes_ips
```

otherwise the kubeconfig created would have as API server the cluster name and DNS zone which would not be solvable.

invidian:
Awesome news @mazzy89, I'm glad you got it working! 

>when the DNS provider is set to manual, entries in /etc/hosts will be created.

Hm, I not sure if this is the right way to implement this, as when DNS provider is `manual`, I think code should not be opinionated how DNS entries are created. If we require `controller_ip_addresses` from user anyway, I think we could only suggest via documentation, that DNS entries can be also created using CLC snippet. I would not recommend using `/etc/hosts` except for testing.

>otherwise the kubeconfig created would have as API server the cluster name and DNS zone which would not be resolvable.

I think `kubeconfig` should have the DNS name for API address. Bootkube runs in Docker, which means `/etc/hosts` should be copied into the container on start. See #409 for more details. Perhaps something changed (I'm aware of rkt -> Docker change, but I would expect it to still work) and this is no longer functional?

mazzy89:
> if we require controller_ip_addresses from user anyway, I think we could only suggest via documentation, that DNS entries can be also created using CLC snippet. I would not recommend using /etc/hosts except for testing.

yeah I completely agree with you. 

> I think kubeconfig should have the DNS name for API address.

yeah I agree too here

let me make all the requested changed. I will open in a while the PR for review.

mazzy89:
@invidian The PR is ready for review. I've adjusted the description above reporting the required steps. At the moment there is no documentation or quick start guide but as I recall you said that we would introduce it in a next PR.

mazzy89:
I've been continuing the testing.

One thing I've noticed is that in case of a bigger cluster 3 nodes and 3 workers it takes almost forever to bootstrap. Around 30/40 mins. So I started to debug and noticed that the traffic there is heavily capped. In fact the pulling of the docker images fails multiple times and get successful only after a while.

Now I'm quite sure is something related to my underlying network combined with the default settings of the VMware Terraforn provider. I would need to debug deeper and look at the bandwidth option and in case tune up them and give the possibility to adjust them.

mazzy89:
I've fixed the issue. I had a problem with DNS. Yeah I know It's ALWAYS DNS 🤠

On my network, I had to point the DNS to my recursive one and not the default one set up which is Google on `8.8.8.8`.

Now the cluster with 3 controllers and 3 workers goes ⬆️ in `took 14m52s`

mazzy89:
Currently, the update of a cluster does not go well. I get the following error in the `controller-0`:

```
Jan 24 11:08:10 lokomotive-cluster-controller-0 ba2b539592f5[850]: 2021-01-24 11:08:10.391154 E | rafthttp: request cluster ID mismatch (got a6dd8d5938a30d31 want 9cb898310a8dd110)
Jan 24 11:08:10 lokomotive-cluster-controller-0 ba2b539592f5[850]: 2021-01-24 11:08:10.470701 E | rafthttp: request cluster ID mismatch (got a6dd8d5938a30d31 want 9cb898310a8dd110)
Jan 24 11:08:10 lokomotive-cluster-controller-0 ba2b539592f5[850]: 2021-01-24 11:08:10.570811 E | rafthttp: request cluster ID mismatch (got a6dd8d5938a30d31 want 9cb898310a8dd110)
Jan 24 11:08:10 lokomotive-cluster-controller-0 ba2b539592f5[850]: 2021-01-24 11:08:10.591355 E | rafthttp: request cluster ID mismatch (got a6dd8d5938a30d31 want 9cb898310a8dd110)
Jan 24 11:08:10 lokomotive-cluster-controller-0 ba2b539592f5[850]: 2021-01-24 11:08:10.670797 E | rafthttp: request cluster ID mismatch (got a6dd8d5938a30d31 want 9cb898310a8dd110)
```

and the process gets stuck within:

```
Now checking health and readiness of the cluster nodes ...
```

**How to reproduce:**

1. Create the cluster with 1 controller and 1 worker.
2. Once the cluster is successfully created, scale up the cluster with 3 controllers and 3 workers. In other terms we add more resources to the cluster that is a very common scenario.
3. At the point the API becomes unavailable and in the controller, we get the error as reported above.

I haven't yet investigated deeper.

Do we guys support rolling updates? Which is the strategy here to upgrade a cluster?

invidian:
@mazzy89 adding/removing controller nodes is currently not supported nor tested. See #1121 and #29 for details.

mazzy89:
@mazzy89 oh indeed. makes sense then.

mazzy89:
@invidian Should then, for this reason, ignore changes to the user data of the controllers like we do for AWS provider?

invidian:
>@invidian Should then, for this reason, ignore changes to the user data of the controllers like we do for AWS provider?

Yes, I think this should be safe to do for now.

mazzy89:
@invidian I've done with the tests on my side. Everything looks good.

mazzy89:
May I help here anyhow to move this further?

iaguis:
> May I help here anyhow to move this further?

Hey @mazzy89, sorry this is taking so long. I'll review today or tomorrow.

mazzy89:
No problems. Ping me whenever you want if you need any clarification and give you more context. Thanks again

iaguis:
> No problems. Ping me whenever you want if you need any clarification and give you more context. Thanks again

Hy @mazzy89, I didn't forget about this. I've been trying to see how we can run CI on this and it seems there's no way around getting actual vSphere, I'll see what I can do.

mazzy89:
Thanks @iaguis. Yeah CI on Vsphere is the biggest problem here. Equinix provides ESXi machines with free license unfortunately (no VSphere). I could look at it as well.

mazzy89:
@iaguis do you have updates here? I was thinking if you can grab from VMware a vlab to use as CI env test.do you think is doable?

mazzy89:
Feel free to close it whether it is not possible to find out a solution or not much business interest on your side.

PR #1319, opened 2021-01-14
Add 'terraform fmt' check and fix files formatting in assets

### LINKED ISSUE #115: Run 'terraform fmt' as part of CI
So all Terraform files are consistently formatted.

Just calling `terraform fmt -recursive assets/` should be enough, though we need to make sure Terraform binary is available in CI image.

When we do it, we could also remove `e2e` tag from Terraform executor test actually.

invidian:
Closes #115

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -196,8 +196,12 @@ check-vendor: check-working-tree-clean vendor
 check-docs: check-working-tree-clean docs
 	@test -z "$$(git status --porcelain)" || (echo "Please run make docs and commit the changes."; git status; exit 1)
 
+.PHONY: check-terraform
+check-terraform:
+	terraform fmt -check -recursive ./assets/
+
 .PHONY: ci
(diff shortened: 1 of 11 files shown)


knrt10:
Can you please check the CI errors @invidian ?

knrt10:
@invidian for second commit 

> Found using Terraform 0.14.4

But required version that we suggest is `>= 0.13, < 0.14`. should we lint with v0.14+ ?

invidian:
>But required version that we suggest is >= 0.13, < 0.14. should we lint with v0.14+ ?

We should actually upgrade to Terraform 0.14 everywhere. See #1262.

knrt10:
Cool, then this should be merged after we update to terraform v0.14+

invidian:
>Cool, then this should be merged after we update to terraform v0.14+

No. I think it's fine to merge it right now.

knrt10 (approved):
+1

PR #1325, opened 2021-01-15
Update Tinkerbell to latest version

invidian:


invidian:
Hmm, that might still not work without updated Tinkerbell Terraform provider.

PR #1332, opened 2021-01-19
Add diff flag to terraform check

ipochi:
This PR adds the adds a `-diff` flag to `terraform fmt`, so that the user can see the output what needs to be fixed.

### DIFF
--- a/assets/terraform-modules/packet/flatcar-linux/kubernetes/workers/workers.tf
+++ b/assets/terraform-modules/packet/flatcar-linux/kubernetes/workers/workers.tf
@@ -59,7 +59,7 @@ data "ct_config" "ignitions" {
       cluster_domain_suffix = var.cluster_domain_suffix
       node_labels = merge({
         "node.kubernetes.io/node"                 = "",
-        "lokomotive.alpha.kinvolk.io/bgp-enabled" = format("%t", !var.disable_bgp),
+        "lokomotive.alpha.kinvolk.io/bgp-enabled" = format("%t", ! var.disable_bgp),
       }, var.labels)
       taints               = var.taints
       setup_raid           = var.setup_raid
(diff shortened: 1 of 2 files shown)


assets/terraform-modules/packet/flatcar-linux/kubernetes/workers/workers.tf
```diff
@@ -59,7 +59,7 @@ data "ct_config" "ignitions" {
       cluster_domain_suffix = var.cluster_domain_suffix
       node_labels = merge({
         "node.kubernetes.io/node"                 = "",
-        "lokomotive.alpha.kinvolk.io/bgp-enabled" = format("%t", !var.disable_bgp),
+        "lokomotive.alpha.kinvolk.io/bgp-enabled" = format("%t", ! var.disable_bgp),
```
invidian, line 62: What version of Terraform you use? For me, 0.14.x allows `!var.x` formatting.

invidian: It seems GitHub Action agrees with me :smile:

ipochi: I believe the `terraform fmt` should be based on 0.13. Github Actions needs to change.
What do you think ?

invidian: Not sure how to proceed here :thinking:

invidian: I'm not even sure why GitHub Action has Terraform already installed actually :smile: But yeah, we can change GitHub action to have 0.13.x then.

PR #1333, opened 2021-01-19
[WIP] Baremetal improvements

pothos:
- baremetal: use grub.cfg file for kernel arguments
    While the kernel arguments can be set for PXE, they have not been
    applied to the final installation. Also, the Ignition URL got lost
    because it was downloaded in the installer and put to
    /usr/share/oem/config.ign through the -i ignition.json flag.
    Use the /usr/share/oem/grub.cfg file to permanently set the kernel
    arguments and the Ignition URL, so that on each boot the final
    installation uses the kernel arguments provided by the user and
    that Ignition is able to fetch the any config changes from Matchbox.
- baremetal: integrate provisioning logic    
    The bare-metal platform lacked any understanding of whether an instance
    is actually running the configuration that lokoctl put into matchbox,
    because, when the configuration was updated, there was no notification
    to the user that PXE booting has to be done again for this instance.
    Also, when to boot from PXE was also clear to the user because it must
    happen after lokoctl populated matchbox with the (new) configuration
    but before any other steps are timing out.
    A null resource and is used with the possibility to run user-defined
    commands that can automate PXE provisioning at the right time, i.e.,
    initially at the first lokoctl run and after each Ignition
    configuration change. However, PXE booting is a long and maybe even
    manual process, or even impossible at production due to the lack of
    an appropriate DHCP server. We can rely on Ignition to simulate
    reprovisioning by creating the first_boot flag file via SSH and
    issuing a reboot, which makes Ignition fetch the configuration from
    matchbox, and if we make sure to clean the root filesystem the result
    is the same as if reprovisioned with a PXE boot.
- baremetal: rework (re-)provisioning logic to detect first run
    The PXE provisioning logic should only be done if it's either the first
    run or if a worker node gets reconfigured. For controllers this is
    currently not supported due to losing etcd state and ensuring there is
    always the quorum held. It should also be robust against any races and
    wait verify that the reprovisioning actually took place.
    The new way of doing it works with local state files under the asset
    folder. This serves as source of truth and also gives the user a way to
    force a PXE install if needed (reminder: all this exercise here is done
    because we don't have a Terraform provider doing this for us). A flag
    file that Ignition creates and that a validation step removes further
    helps to prevent any races where the worker secrets SSH step would
    already run before the reprovisioning took place.
- baremetal: introduce CLC snippets for the installer
    The installer OS also needs to be able to get its network configuration
    as CLC snippet if DHCP is not used.
- baremetal: Do not hardcode kernel console setting
- baremetal: Add pre-reboot command for installation process
- baremetal: let installer.service retry on failure
    In case the network only has a temporary problem, the installer service
    should start again on failure.
    Convert it to a oneshot service to set RemainAfterExist, so that it is
    not triggered twice if anything that depends on it is pulling it in
    again after it finished.
- baremetal: run copy-controller-secrets step after recreating the node
- baremetal: update the kernel arguments when skipping a PXE boot
    When the PXE boot is skipped, the new Ignition config is already
    fetched but it does not contain the kernel arguments because these
    have to be set outside of Ignition or otherwise they would only
    apply to the following boots. The kernel arguments are kept in the
    grub config and this needs to be updated out of band (kernel console
    and Ignition URL are also part of the kernel arguments).
- baremetal: add etcd-rejoin script from AWS/Packet for control plane HA
- baremetal: fix wrong etcd service name

# How to use

I added an empty PXE command because the CI already boots the VMs up at the right time. This could be simplified by supplying the config `pxe_commands = "sudo virt-install … --network=bridge:pxe0,mac=$mac …" ` (or in this case `scripts/pxe-testbed … create`) but it needs an additional step of leaving the container (`nsenter …`) because the VMs are created outside of the lokoctl container.

# Testing done

- Updating the worker's CLC snipped triggered a reboot and the change was applied.
- Deleting the worker VM and making a config change to trigger a recreating worked, too.
- Reprovisioning a baremetal controller in a 3 controller cluster and have etcd rejoin
   `rm lokoctl-assets/cluster-assets/MACADDR # grep controller2 lokomotive/lokoctl-assets/cluster-assets/*`
   `vi cl/controller2.lokomotive.yaml`
   `lokoctl cluster apply --skip-pre-update-health-check -v`


assets/terraform-modules/bare-metal/flatcar-linux/kubernetes/profiles.tf
```diff
@@ -68,6 +69,7 @@ resource "matchbox_profile" "cached-flatcar-linux-install" {
     container_linux_oem      = var.container_linux_oem
     ssh_keys                 = jsonencode(var.ssh_keys)
     install_to_smallest_disk = var.install_to_smallest_disk
+    kernel_args              = join(" ", var.kernel_args)
```
invidian, line 72: When we use `templatefile`, we can call functions inside the template, which would be preferred way, as we keep high-level type as long as possible and we do type conversion when it's actually needed.

assets/terraform-modules/bare-metal/flatcar-linux/kubernetes/cl/install.yaml.tmpl
```diff
@@ -40,9 +39,13 @@ storage:
...
             -V ${os_version} \
             -o "${container_linux_oem}" \
-            ${baseurl_flag} \
-            -i ignition.json
+            ${baseurl_flag}
           udevadm settle
+          OEM_DEV="$(blkid -t "LABEL=OEM" -o device)"
+          mkdir -p /tmp/oemfs
+          mount "$${OEM_DEV}" /tmp/oemfs
+          echo 'set linux_append="${kernel_args} ignition.config.url=${ignition_endpoint}?{{.request.raw_query}}&os=installed"' | tee -a /tmp/oemfs/grub.cfg
```
invidian, line 47: What happens when ignition endpoint becomes unavailable? Flatcar will still be able to boot?

pothos: Yes, Ignition will only query it if it runs and by default it only runs once. `touch /boot/flatcar/first_boot` will allow to trigger it again.

pothos:
new state in https://github.com/kinvolk/lokomotive/pull/1402

pothos:
(please keep branch around for some time)

PR #1345, opened 2021-01-26
pkg/components/util: Fix bug with multiple hooks in same manifest in helm chart

jbanety:
# Allow multiple hooks in same manifest in helm chart

If a helm chart contains multiple hooks in one manisfest file, the hooks result are overridden because the map key is the same for each one.

# How to use

Use chart with multiple hooks in the same manifest file.

# Testing done

Consider a manifest with two TLS certs generated by hooks.

```yaml
---
apiVersion: v1
kind: Secret
metadata:
  name: {{ $fullName }}-controller-tls
  annotations:
    "helm.sh/resource-policy": "keep"
    "helm.sh/hook": "pre-install"
    "helm.sh/hook-delete-policy": "before-hook-creation"
    "directives.qbec.io/update-policy": "never"
type: kubernetes.io/tls
data:
  tls.crt: "[...]"
  tls.key: "[...]"
  ca.crt: "[...]"
---
apiVersion: v1
kind: Secret
metadata:
  name: {{ $fullName }}-client-tls
  annotations:
    "helm.sh/resource-policy": "keep"
    "helm.sh/hook": "pre-install"
    "helm.sh/hook-delete-policy": "before-hook-creation"
    "directives.qbec.io/update-policy": "never"
type: kubernetes.io/tls
data:
  tls.crt: [...]
  tls.key: [...]
  ca.crt: [...]
```

With the current behavior, only the second one is kept.
With the patch, all manifests are stored.

### DIFF
--- a/pkg/components/util/helm.go
+++ b/pkg/components/util/helm.go
@@ -59,7 +59,11 @@ func RenderChart(helmChart *chart.Chart, name, namespace, values string) (map[st
 
 	// Include hooks
 	for _, m := range template.Hooks {
-		ret[m.Path] = m.Manifest
+		if val, ok := ret[m.Path]; ok {
+			ret[m.Path] = val + "\n---\n" + m.Manifest
+		} else {
+			ret[m.Path] = m.Manifest
+		}
 	}
 
 	// CRD are not rendered, so do this manually
--- a/pkg/components/util/helm.go
+++ b/pkg/components/util/helm.go
@@ -59,11 +59,11 @@ func RenderChart(helmChart *chart.Chart, name, namespace, values string) (map[st
 
 	// Include hooks
 	for _, m := range template.Hooks {
-		if val, ok := ret[m.Path]; ok {
-			ret[m.Path] = val + "\n---\n" + m.Manifest
-		} else {
-			ret[m.Path] = m.Manifest
-		}
+                if _, ok := ret[m.Path]; !ok {
+                        ret[m.Path] = ""
+                }
+                
+                ret[m.Path] += "---\n" + m.Manifest
 	}
 
 	// CRD are not rendered, so do this manually
--- a/pkg/components/util/helm.go
+++ b/pkg/components/util/helm.go
@@ -59,11 +59,11 @@ func RenderChart(helmChart *chart.Chart, name, namespace, values string) (map[st
 
 	// Include hooks
 	for _, m := range template.Hooks {
-                if _, ok := ret[m.Path]; !ok {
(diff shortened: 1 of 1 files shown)


pkg/components/util/helm.go
```diff
@@ -59,7 +59,11 @@ func RenderChart(helmChart *chart.Chart, name, namespace, values string) (map[st
 
 	// Include hooks
 	for _, m := range template.Hooks {
-		ret[m.Path] = m.Manifest
+		if val, ok := ret[m.Path]; ok {
+			ret[m.Path] = val + "\n---\n" + m.Manifest
+		} else {
+			ret[m.Path] = m.Manifest
+		}
```
invidian, line 66: I think this code can be simplified to something like:
```suggestion
                if _, ok := ret[m.Path]; !ok {
                        ret[m.Path] = ""
                }
                
                ret[m.Path] += "---\n" + m.Manifest
```

invidian (changes requested):
Thanks for the PR @jbanety! The patch overall looks good to me, nice catch!

I wonder if this fix might address #1128.

When we fix this behavior for `Hooks`, I think we should also fix it for `CRDs()` below, which I suppose might also be affected.

Do you mind adding some unit tests for this code? Similar test-cases are already implemented, so I think it shouldn't be too hard.

pkg/components/util/helm.go
jbanety: OK. I didn't read contrib guidlines. Shame on me....
I'll update my PR asap.

invidian: Ping @jbanety, are you still interested in doing this work?

jbanety: Hi @invidian,
Sorry. Yes I am. Little busy at the moment.
I'll try to work on it this week

jbanety:
Hi @invidian,

> When we fix this behavior for `Hooks`, I think we should also fix it for `CRDs()` below, which I suppose might also be affected.

CRDs are indexed by name, not by file path so there is no problem.

> Do you mind adding some unit tests for this code? Similar test-cases are already implemented, so I think it shouldn't be too hard.

Since we need to load a component chart, I can't create example with existing components in lokomotive.

> I would also prefix commit message with `pkg/components/util:`, following our [contribution guidelines](https://github.com/kinvolk/contribution/blob/master/topics/git.md).

Done

invidian:
@knrt10 maybe you could have a look to add implementing unit tests for it? It shouldn't be too hard.

PR #1317, opened 2021-01-14
Baremetal platform feature parity

ipochi:
This PR adds configuration knobs to Lokomotive which are present on the Terraform side such as:

`install_disk`
`network_ip_autodetection`
`kernel_args`
`download_protocol`

Also adds a new configuration option: `install_to_smallest_partition` which installs Flatcar Container Linux on the smallest disk partition.

PR #1363, opened 2021-02-05
baremetal: use grub.cfg file for kernel arguments

ipochi:
This commit is cherry picked from b79eacc9002292deabe0f4fe878ffdce1caea5a6 because this is a fix for the Kernel arguments
variable that was added as part of Baremetal feature parity [here](https://github.com/kinvolk/lokomotive/pull/1317/commits/f603a05c93679a6d8768bd9624daef83b8e8a003).

This commit, IMO should be part of the 0.6.1 patch release rather than continue in the PR #1333


assets/terraform-modules/bare-metal/flatcar-linux/kubernetes/cl/install.yaml.tmpl
```diff
@@ -40,9 +39,13 @@ storage:
             -C ${os_channel} \
             -V ${os_version} \
             -o "${container_linux_oem}" \
-            ${baseurl_flag} \
```
invidian, line 43: Hm, I agree about preserving kernel arguments, but I think it's OK that Ignition is only applied during installation. We do not support running Ignition multiple times and I have doubts this is something we want to support. Or at least we should have a conversation about this.

ipochi:
closing this after discussing with @pothos that #1333 should not be split.

invidian:
@ipochi can branch be removed?

Issue #1247, opened 2020-12-07
Figure out how to run e2e tests for PRs outside the repository


invidian:
**Current situation**
Recently, we've got some PRs from outside contributors and right now we are not able to run e2e tests on them, as we only allow them for PRs originating from this repository.

**Impact**
Having no CI for external contributions may affect the quality or impact contributing community, as process for merging those PRs will be slower, as it will require manual testing. Or we will have to take over all PRs and create replacements.

**Ideal future situation**
We should have a way, similar to `ok-to-test` label from Kubernetes repositories, which enables us to allow running CI for a given commits for external contributors.

invidian:
I _think_ we could use something like https://github.com/colstrom/concourse-github-status to run tests for all pushes to this repository. This would include all feature branches and master branch.

It will also be nicer to what we currently have, as we will have a status for merge commits in the master branch.

This way, if someone opens a PR from fork, we simply need to pull it via `git pr` or with `gh pr checkout` and then re-push the branch to main repository.

knrt10:
ideal situation will be handled by bot, once we decide to move forward with it

invidian:
We could also run the CI for PRs which are approved from external contributors, as GitHub plugin we currently use for Concourse supports that. We would combine that with automatic dismissal of approvals on push then.

PR #1372, opened 2021-02-12
Remove a deprecated cert-manager namespace label

meyskens:
# Remove a deprecated cert-manager namespace label

This removed an old label on the Kubernetes namespace that is no longer needed for cert-manager and would no longer work after 0.11 ;) 

# How to use

no new features added, just should be tested that everything stays working

# Testing done

none, any preferred testing I should perform?

### DIFF
--- a/pkg/components/cert-manager/component.go
+++ b/pkg/components/cert-manager/component.go
@@ -101,9 +101,6 @@ func (c *component) Metadata() components.Metadata {
 		Name: Name,
 		Namespace: k8sutil.Namespace{
 			Name: c.Namespace,
-			Labels: map[string]string{
-				"certmanager.k8s.io/disable-validation": "true",
-			},
 		},
 		Helm: components.HelmMetadata{
 			// Cert-manager registers admission webhooks, so we should wait for the webhook to


pkg/components/cert-manager/component.go
```diff
@@ -101,9 +101,6 @@ func (c *component) Metadata() components.Metadata {
 		Name: Name,
 		Namespace: k8sutil.Namespace{
 			Name: c.Namespace,
-			Labels: map[string]string{
```
invidian, line 104: We could still replace it with new label, as it's defined in https://github.com/kinvolk/lokomotive/blob/0b0c39c449f1b916182539f987de357e284b87a9/assets/charts/components/cert-manager/templates/webhook-validating-webhook.yaml#L27, but given that it was not working anyway for such a long time, I guess it's fine to remove it :smile:

I recall I had issues with webhooks timing out on some old AKS version and I had to disable the webhooks to get this working, maybe that was the cause :smile: 

I doubt we have tests for it, so as long CI is passing, it should be fine to merge.

invidian (approved):
Hey @meyskens ! Thanks for opening the PR. Because of to https://github.com/kinvolk/lokomotive/issues/1247, I'll open another temporary PR to let CI run with this patch and then we can merge :)

pkg/components/cert-manager/component.go
meyskens: guess I need to PR a cert-manager validator then ;) 
cert-manager used to create a certificate resource for itself which sometimes caused issued it now changed the way it signs it's own webhook (chicken egg issue there) for which the validating webhook not being ready should no longer cause issues

meyskens: on second thought better to double check as i saw somewhere you installed 2 clusterissuers also? Haven't taken a close look where exactly they are installed

meyskens: on third thought I just remembered that cluster issuers are of course cluster scoped (got confused with where the secrets are) :facepalm:

invidian: Yeah, we install 2 clusterissuers as part of the cert-manager itself, so we support letsencrypt out of the box, both production and staging.

invidian (approved):
CI passes in #1373, so let's get another LGTM and we're good to go.

PR #1405, opened 2021-02-25
WIP - do not review; baremetal : node-specific labels

ipochi:


assets/terraform-modules/bare-metal/flatcar-linux/kubernetes/variables.tf
```diff
@@ -203,3 +203,9 @@ variable "install_to_smallest_disk" {
   type        = bool
   default     = false
 }
+
+variable "node_specific_labels" {
+  type        = map(map(string))
+  description = "Map of node specificwq labels map."
```
knrt10, line 209:
```suggestion
  description = "Map of node specific labels map."
```

pkg/platform/baremetal/template.go
```diff
@@ -65,6 +65,20 @@ module "bare-metal-{{.ClusterName}}" {
...
+    {{- if $mapOfLabels }}
+               "{{$nodeName}}" = {
+          {{- range $key, $value := $mapOfLabels }}
+         "{{$key}}" = "{{$value}}",
+                {{- end }}
+               }
+               {{- end }}
+    {{- end}}
+  }
+  {{- end}}
```
knrt10, line 80: This can be linted I guess

test/components/kubernetes/baremetal_kubelet_labels_test.go
```diff
@@ -0,0 +1,75 @@
+// Licensed under the Apache License, Version 2.0 (the "License");
```
knrt10, line 1: Missing

`// Copyright 2021 The Lokomotive Authors`

test/components/kubernetes/baremetal_kubelet_labels_test.go
```diff
@@ -0,0 +1,75 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	pkglabels "k8s.io/apimachinery/pkg/labels"
+
+	testutil "github.com/kinvolk/lokomotive/test/components/util"
+)
+
+func TestBaremetalNodeHasNodeSpecificLabels(t *testing.T) {
+	// This map is copied from file ci/baremetal/baremetal-cluster.lokocfg.envsubst
+	// the field is `node_specific_labels`.
+	// The node name keys are costructed as follows: <cluster-name>-<controller/worker>-<index>
```
knrt10, line 31:
```suggestion
	// The node name keys are constructed as follows: <cluster-name>-<controller/worker>-<index>
```

knrt10 (changes requested):
Nits

test/components/kubernetes/baremetal_kubelet_labels_test.go
```diff
@@ -0,0 +1,75 @@
...
+	"context"
+	"testing"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	pkglabels "k8s.io/apimachinery/pkg/labels"
+
+	testutil "github.com/kinvolk/lokomotive/test/components/util"
+)
+
+func TestBaremetalNodeHasNodeSpecificLabels(t *testing.T) {
```
invidian, line 28: I think we could name this test `Test_Baremetal_NodeSpecificLabels` and then has 2 sub-tests for it, one for checking that labels has been applied, and other one to check that common labels are retained.

test/components/kubernetes/baremetal_kubelet_labels_test.go
```diff
@@ -0,0 +1,75 @@
...
+		"mercury-worker-1": {
+			"storagenode": "yes",
+		},
+	}
+
+	client := testutil.CreateKubeClient(t)
+
+	nodes, err := client.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		t.Errorf("could not list nodes: %v", err)
```
invidian, line 48: You should use `Fatalf` here, otherwise `nodes` might be nil and code will panic when you dereference `nodes` for `Items` field.

test/components/kubernetes/baremetal_kubelet_labels_test.go
```diff
@@ -0,0 +1,75 @@
...
+			LabelSelector: labels,
+		})
+		if err != nil {
+			t.Errorf("expected node with name %q not found: %v", nodeName, err)
+		}
+
+		// Currently we assume that the labels added to the node in CI are unique to the worker node.
+		if len(nodes.Items) > 1 {
+			t.Errorf("more than one worker node found having labels: %q", labels)
+		}
```
invidian, line 67: I think this check doesn't make much sense. We should check that there is at least one instead, as this is exactly what we expect. Currently we test here CI configuration, which shouldn't matter.

invidian (changes requested):
Mostly looking good, I have some suggestions we could apply.

test/components/kubernetes/baremetal_kubelet_labels_test.go
ipochi: We can do a check that only one node should be listed with the provided values in LabelSelector and that node should match the name as a reassuring check.

test/components/kubernetes/baremetal_kubelet_labels_test.go
ipochi: I'd rather append the labels provided in ci configuration in
```
    nodeNamesAndExpectedLabelsInCI := map[string]map[string]string{
      "mercury-controller-0": {
        "testkey": "testvalue",
        "node.kubernetes.io/master": "",
        "node.kubernetes.io/controller": "true",
      },
      "mercury-worker-0": {
        "ingressnode": "yes",     
        "testing.io": "yes",
        "roleofnode": "testing",
      },
      "mercury-worker-1": {
        "storagenode": "yes",
        "testing.io": "yes",
        "roleofnode": "testing"
      },
    }
    ```
    
and pass map of specific node to LabelSelector, which would confirm that node specifc labels were applied as well as  confirm that common labels were retained.

File: test/components/kubernetes/baremetal_kubelet_labels_test.go (hunk 7 above)
Comment (ipochi) [reply test/components/kubernetes/baremetal_kubelet_labels_test.go:67]:
```
    // Currently we assume that the labels added to the node in CI are unique to the worker node.
    // This check confirms that the expected set of labels as per CI configuration were
    // present on the correct node.
    if len(nodes.Items) == 1 && nodes.Items[0].Name != nodeName {
      t.Errorf("expected node %q to have the labels %s", nodes.Items[0].Name, labels)
    }
```
even this won't be needed if we do this https://github.com/kinvolk/lokomotive/pull/1405/commits/f5f9248a5557c2bcccd614f4db4ccbe8e18ddae2#r589997444

test/components/kubernetes/baremetal_kubelet_labels_test.go
invidian: Sounds good.

test/components/kubernetes/baremetal_kubelet_labels_test.go
invidian: Sounds good :+1:

invidian (approved):

knrt10 (approved):
+1

PR #1406, opened 2021-03-01
Add cpu_manager_policy to workers

knrt10:
This add flag to worker pools also added a knob for the user to configure
it.

closes: #1337


surajssd:
I was adding this feature to baremetal and was wondering how to restrict this to certain workers since there are no worker pools for it as of now. WDYT if I skip adding support to baremetal?

assets/terraform-modules/packet/flatcar-linux/kubernetes/workers/variables.tf
```diff
@@ -195,3 +195,21 @@ variable "enable_tls_bootstrap" {
...
 }
+
+variable "cpu_manager_policy" {
+  description = "CPU Manager policy to use for the worker pool. Possible values: `none`, `static`."
+  default     = "none"
+  type        = string
+}
+
+variable "kube_reserved_cpu" {
+  description = "CPU cores reserved for the Worker Kubernetes components like kube-proxy, kubelet, etc."
```
invidian, line 206: `kube-proxy` runs as a pod, so it should not be included here.

assets/terraform-modules/packet/flatcar-linux/kubernetes/workers/variables.tf
```diff
@@ -195,3 +195,21 @@ variable "enable_tls_bootstrap" {
...
+  default     = "2000m"
+  type        = string
+}
+
+variable "system_reserved_cpu" {
+  description = "CPU cores reserved for the Worker services like docker, sshd, kernel, etc."
```
invidian, line 212:
```suggestion
  description = "CPU cores reserved for the host services like Docker, sshd, kernel, etc."
```

assets/terraform-modules/packet/flatcar-linux/kubernetes/workers/variables.tf
```diff
@@ -195,3 +195,21 @@ variable "enable_tls_bootstrap" {
...
+
+variable "cpu_manager_policy" {
+  description = "CPU Manager policy to use for the worker pool. Possible values: `none`, `static`."
+  default     = "none"
+  type        = string
+}
+
+variable "kube_reserved_cpu" {
+  description = "CPU cores reserved for the Worker Kubernetes components like kube-proxy, kubelet, etc."
+  default     = "2000m"
```
invidian, line 207: 2 cores only for kubelet and etcd? That seems like a lot. How this has been estimated?

invidian:
>I was adding this feature to baremetal and was wondering how to restrict this to certain workers since there are no worker pools for it as of now. WDYT if I skip adding support to baremetal?

Lack of worker pools on Bare metal (actually Matchbox) seems like ongoing issue with feature parity.

assets/terraform-modules/packet/flatcar-linux/kubernetes/workers/variables.tf
surajssd: Good point. We have this enabled only for workers not for control plane. So etcd is out of the picture. Talking about the idle kubelet yeah 2 cores can be a lot for it. I will reduce it to 500m.

invidian: On my nodes with ~30 pods, CPU does not exceed 0.15 core. What's our max pods per node here? We could calculate as such perhaps.

invidian: Selection_441
CPU usage per pod per kubelet instance.

pkg/platform/aws/aws.go
```diff
@@ -31,23 +31,24 @@ import (
...
-	DiskIOPS     int               `hcl:"disk_iops,optional"`
-	SpotPrice    string            `hcl:"spot_price,optional"`
-	TargetGroups []string          `hcl:"target_groups,optional"`
-	CLCSnippets  []string          `hcl:"clc_snippets,optional"`
-	Tags         map[string]string `hcl:"tags,optional"`
-	LBHTTPPort   int               `hcl:"lb_http_port,optional"`
-	LBHTTPSPort  int               `hcl:"lb_https_port,optional"`
+	Name             string            `hcl:"pool_name,label"`
+	Count            int               `hcl:"count"`
+	CPUManagerPolicy string            `hcl:"cpu_manager_policy,optional"`
```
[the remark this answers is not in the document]
surajssd: Yep empty == none.

assets/terraform-modules/packet/flatcar-linux/kubernetes/workers/variables.tf
surajssd: Is `# of pods` proportional to kubelet CPU usage? Also what about the spike of CPU usage before the self-hosted takes over?

pkg/platform/aws/aws.go
invidian, line 36: Value of this field should be validated I think.

pkg/platform/aws/template.go
```diff
@@ -198,6 +198,8 @@ module "worker-pool-{{ $index }}" {
   clc_snippets          = {{ (index $.WorkerpoolCfg $index "clc_snippets") }}
   {{- end }}
 
+  cpu_manager_policy = "{{$pool.CPUManagerPolicy}}"
```
invidian, line 201: Terraform variable description says `Possible values: `none`, `static`.`, while this will put `""`, which IIRC is different from `null` or not specifying it, so it can pick up the default value.

surajssd:
@invidian one more round of review?

invidian:
@surajssd I already left some comments, I don't see any replies or new code to review. Also, PR has conflicts which must be resolved.

surajssd:
> @surajssd I already left some comments, I don't see any replies or new code to review. Also, PR has conflicts which must be resolved.

My bad, I had this page open for many days and did not see those comments.

pkg/platform/aws/template.go
surajssd: Yep "" is put in here, which is equally treated as empty. And the default value of `non` is picked up.

pkg/platform/aws/aws.go
surajssd: done

pkg/platform/aws/aws.go
```diff
@@ -410,3 +412,22 @@ func (c *config) checkWorkerPoolNamesUnique() hcl.Diagnostics {
 
 	return diagnostics
 }
+
+func (c *config) checkCPUManagerPolicy() hcl.Diagnostics {
```
invidian, line 416: Perhaps we should have some tests for this.

pkg/platform/aws/aws.go
invidian: Thanks!

pkg/platform/aws/aws.go
surajssd: Added.

assets/terraform-modules/packet/flatcar-linux/kubernetes/workers/variables.tf
surajssd: They are limits for the kubelet when we enable the cpu-manager-policy. Because rest of the CPU is allocated to others exclusively. 

About number of CPUs I followed [this guide](https://cloud.redhat.com/blog/full-cluster-part-2-protecting-nodes) which also recommends `# of pods * 5m`. I doubled it just to be conservative.

invidian: Do we set max number of pods or have it configurable?

assets/terraform-modules/aws/flatcar-linux/kubernetes/workers/variables.tf
```diff
@@ -173,3 +173,22 @@ variable "enable_csi" {
...
+
+variable "kube_reserved_cpu" {
+  description = "CPU cores reserved for the Worker Kubernetes components like kubelet, etc."
+  default     = "300m"
+  type        = string
+}
+
+variable "system_reserved_cpu" {
+  description = "CPU cores reserved for the host services like Docker, sshd, kernel, etc."
+  default     = "1500m"
```
invidian, line 192: This together seems a lot if you use an instance, which has only 2 cores for example. It leaves only 200m for controlplane.

pkg/platform/aws/aws.go
```diff
@@ -410,3 +412,22 @@ func (c *config) checkWorkerPoolNamesUnique() hcl.Diagnostics {
...
+
+	for _, w := range c.WorkerPools {
+		switch w.CPUManagerPolicy {
+		case "", "none", "static":
+			continue
+		default:
+			diagnostics = append(diagnostics, &hcl.Diagnostic{
+				Severity: hcl.DiagError,
+				Summary:  "invalid cpu_manager_policy",
+				Detail:   fmt.Sprintf("Worker pool '%s' has invalid cpu_manager_policy '%s'", w.Name, w.CPUManagerPolicy),
```
invidian, line 427: `%q` should be used for quoting.
```suggestion
				Detail:   fmt.Sprintf("Worker pool %q has invalid cpu_manager_policy %q", w.Name, w.CPUManagerPolicy),
```

pkg/platform/aws/aws_internal_test.go
```diff
@@ -222,3 +223,64 @@ func TestConfigurationIsValidWhen(t *testing.T) {
...
+		},
+	}
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			c := &config{
+				WorkerPools: tt.WorkerPools,
+			}
+
+			if got := c.checkCPUManagerPolicy(); !reflect.DeepEqual(got, tt.want) {
```
invidian, line 281: DeepEqual on diagnostic text seems annoying to maintain. Why can't we just use `HasErrors()`?

Also, `cmp.Diff` usually is nicer to use than `DeepEqual` :)

pkg/platform/aws/aws_internal_test.go
```diff
@@ -222,3 +223,64 @@ func TestConfigurationIsValidWhen(t *testing.T) {
 		})
 	}
 }
+
+func Test_config_checkCPUManagerPolicy(t *testing.T) {
```
invidian, line 227:
```suggestion
func Test_config_checkCPUManagerPolicy(t *testing.T) {
  t.Parallel()
  
```

pkg/platform/aws/aws_internal_test.go
```diff
@@ -222,3 +223,64 @@ func TestConfigurationIsValidWhen(t *testing.T) {
...
+					Summary:  "invalid cpu_manager_policy",
+					Detail:   "Worker pool 'bar' has invalid cpu_manager_policy 'bar'",
+				},
+			},
+		},
+	}
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			c := &config{
```
invidian, line 277:
```suggestion
		t.Run(tt.name, func(t *testing.T) {
		        t.Parallel()
		        
			c := &config{
```

pkg/platform/packet/packet.go
```diff
@@ -614,6 +616,25 @@ func (c *config) validateOSVersion() hcl.Diagnostics {
...
+			diagnostics = append(diagnostics, &hcl.Diagnostic{
+				Severity: hcl.DiagError,
+				Summary:  "invalid cpu_manager_policy",
+				Detail:   fmt.Sprintf("Worker pool '%s' has invalid cpu_manager_policy '%s'", w.Name, w.CPUManagerPolicy),
+			})
+		}
+	}
+
+	return diagnostics
+}
```
invidian, line 636: Can we make this type and validation rule generic so there is no duplicated code and tests?

assets/terraform-modules/packet/flatcar-linux/kubernetes/workers/variables.tf
surajssd: We use the defaults.

assets/terraform-modules/aws/flatcar-linux/kubernetes/workers/variables.tf
surajssd: Is 2 core machine feasible for production use? Also this change is only for workers not controller nodes.

invidian: With 1.5 core overhead, probably not :upside_down_face: I'm not sure, but I feel we can do better at estimating it + make it configurable.

assets/terraform-modules/packet/flatcar-linux/kubernetes/workers/variables.tf
invidian: If we use the default of 110, then I think having it at 500m or something around make sense then :+1:

pkg/platform/aws/aws_internal_test.go
surajssd: Can't make it parallel twice:

```bash
go test -timeout 30s -run TestCheckCPUManagerPolicy github.com/kinvolk/lokomotive/pkg/platform

--- FAIL: TestCheckCPUManagerPolicy (0.00s)
panic: testing: t.Parallel called multiple times [recovered]
	panic: testing: t.Parallel called multiple times

goroutine 20 [running]:
testing.tRunner.func1.2(0x7f3f00, 0xa441b0)
	/usr/local/go/src/testing/testing.go:1144 +0x332
testing.tRunner.func1(0xc00008a900)
	/usr/local/go/src/testing/testing.go:1147 +0x4b6
panic(0x7f3f00, 0xa441b0)
	/usr/local/go/src/runtime/panic.go:965 +0x1b9
testing.(*T).Parallel(0xc00008a900)
	/usr/local/go/src/testing/testing.go:1036 +0x36f
github.com/kinvolk/lokomotive/pkg/platform_test.TestCheckCPUManagerPolicy(0xc00008a900)
	/home/hummer/work/lokomotive/pkg/platform/platform_test.go:236 +0x569
testing.tRunner(0xc00008a900, 0x9d2a90)
	/usr/local/go/src/testing/testing.go:1194 +0xef
created by testing.(*T).Run
	/usr/local/go/src/testing/testing.go:1239 +0x2b3
FAIL	github.com/kinvolk/lokomotive/pkg/platform	0.016s
FAIL
```

surajssd: My bad I had put it in the for loop out of `t.Run` :laughing:

pkg/platform/platform_test.go
```diff
@@ -17,6 +17,8 @@ package platform_test
 import (
 	"testing"
 
+	"github.com/google/go-cmp/cmp"
+	"github.com/hashicorp/hcl/v2"
 	"github.com/kinvolk/lokomotive/pkg/platform"
```
invidian, line 22: Local module imports should be separate from 3rd party.
```suggestion
	"github.com/google/go-cmp/cmp"
	"github.com/hashicorp/hcl/v2"

	"github.com/kinvolk/lokomotive/pkg/platform"
```

pkg/platform/packet/packet.go
```diff
@@ -614,6 +616,15 @@ func (c *config) validateOSVersion() hcl.Diagnostics {
 	return diagnostics
 }
 
+func (c *config) checkCPUManagerPolicy() hcl.Diagnostics {
+	wp := make(map[string]string)
+	for _, w := range c.WorkerPools {
+		wp[w.Name] = w.CPUManagerPolicy
+	}
```
invidian, line 623: I wonder if we could extend `platform.WorkerPool` to make this simpler.

pkg/platform/platform.go
```diff
@@ -281,3 +282,36 @@ func WorkerPoolNamesUnique(pools []WorkerPool) hcl.Diagnostics {
 
 	return d
 }
+
+// CheckCPUManagerPolicy takes a map with key as worker pool name and value as the CPU Manager
+// Policy value. This function checks if the values are valid.
+func CheckCPUManagerPolicy(wp map[string]string) hcl.Diagnostics {
```
invidian, line 288: `wp` doesn't seem to me like a clear name here. And function name does not help either :smile: I know the comment, but still :)

assets/terraform-modules/aws/flatcar-linux/kubernetes/workers/variables.tf
invidian, line 192: I still think this is too high.

invidian (commented):
Perhaps we could use some other reviewers for this PR :)

pkg/platform/packet/packet.go
surajssd: I tried doing that but right now there are conflicts in the struct and the method names if I try to satisfy that interface.

assets/terraform-modules/aws/flatcar-linux/kubernetes/workers/variables.tf
surajssd: Let us start somewhere and adjust value as we see more data around this?

pkg/platform/platform.go
surajssd: Modified the name, also added more context to the command.

assets/terraform-modules/aws/flatcar-linux/kubernetes/workers/variables.tf
```diff
@@ -173,3 +173,22 @@ variable "enable_csi" {
   type        = bool
   default     = false
 }
+
+
```
invidian, line 177:
```suggestion

```

invidian (dismissed):
LGTM

assets/terraform-modules/aws/flatcar-linux/kubernetes/workers/variables.tf
surajssd: Made it 500m.

invidian (approved):

PR #1416, opened 2021-03-04
pkg/components/util: retain all hooks from a single file while rendering

invidian:
With existing implementation, if there is more than one hook defined in
a single file, last hook will overwrite the previous one when rendering,
as we index hooks using Path and 2 hooks will have the same path.

CRDs are not affected, as we index them by name and there should be no
duplicate CRDs in chart defined.

Closes #1345


iaguis (approved):
lgtm when the linter passes :)

iaguis (approved):

PR #1423, opened 2021-03-10
Update certified Kubernetes logo target URL

mauriciovasquezbernal:
The old URL is deprecated and
https://landscape.cncf.io/?selected=lokomotive should be used instead.

# Testing done

Click on the logo and check that not deprecated URL message is shown in the CNCF landscape site.


surajssd (approved):
LGTM

PR #1434, opened 2021-03-17
[WIP] Support the ARM64 platform on bare-metal

meyskens:
# Support the ARM64 platform on bare-metal
This adds the OS Arch option to the bare-metal option to support ARM64.
The changes are taken from the Packet module.

# How to use

set `os_arch = "arm64"`

# Testing done

Needs to be tested tested on my Raspberry Pi 4, will update this a (draft) PR when that is finished. 
Happy to also work on a QEMU baded arm64 CI test if that is desired.


pothos:
Would it make sense first to revert the reverts and then modify the shared module instead of doing this now for the to-be-removed code?

meyskens:
@pothos it probably does... happy to rebase this after those

PR #1472, opened 2021-05-21
baremetal: enable and document "pod_cidr" and "service_cidr" options

### LINKED ISSUE #1471: Allow customizing Pod CIDR and Service CIDR on bare metal
It should be simple exposing the knob which is already used for other platforms.

Useful links for newcomers:
- Where to add fields: https://github.com/kinvolk/lokomotive/blob/a313f02df067c734902deb256dbb64a0146b0fdf/pkg/platform/baremetal/baremetal.go#L33
- What to add: https://github.com/kinvolk/lokomotive/blob/a313f02df067c734902deb256dbb64a0146b0fdf/pkg/platform/aws/aws.go#L74-L75
- Terraform template example: https://github.com/kinvolk/lokomotive/blob/a313f02df067c734902deb256dbb64a0146b0fdf/pkg/platform/aws/template.go#L49-L54
- Docs should be updated as well: https://github.com/kinvolk/lokomotive/blob/9065238f2c2b1e0262363eff4f1bab58ef66092c/docs/configuration-reference/platforms/baremetal.md#attribute-reference

Tests are welcome, but not requires I guess, as we don't have examples.

akurz:
closes #1471

# Allow the customization of pod and service CIDRs also for baremetal setups 

This is a trivial change to the baremetal platform package that allows a user to define custom values for `pod_cidr` and `service_cidr`. Thanks to @invidian for pointing out the necessary code changes :+1:

### DIFF
--- a/docs/configuration-reference/platforms/baremetal.md
+++ b/docs/configuration-reference/platforms/baremetal.md
@@ -82,6 +82,10 @@ cluster "bare-metal" {
 
   network_mtu = 1500
 
+  pod_cidr = "10.2.0.0/16"
+
+  service_cidr = "10.3.0.0/16"
+
   controller_domains = var.controller_domains
 
   controller_macs = var.controller_macs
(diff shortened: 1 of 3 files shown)


akurz:
Anything I can do here to get this PR accepted @invidian?

invidian:
@surajssd @ipochi @iaguis maybe one of you can have a look?

PR #1478, opened 2021-06-02
baremetal: fix terraform non-parallel execution

ipochi:
In order to limit the number of cnocurrent operations as Terraform walks
the graph, `-parallelism=n` is used, with a default value of 10.

When rotating certificates, it is desirable to limit parallelism to `1`,
so that certificates are rotated one by one and we still have quorum.

In baremtal platform, there is a case of missing hiphen before
`parallelism`:
```
current - return ex.Apply([]string{"parallelism=1"})
correct - return ex.Apply([]string{"-parallelism=1"})
```

Hence the following error:

```
stat parallelism=1: no such file or directory
```

This commit provides the fix by correcting the code in question.


invidian:
I don't think there are 3 reviewers required for such trivial fix.

PR #1402, opened 2021-02-25
[WIP] - Baremetal reprovisioning

ipochi:
replacement of #1333 

Includes all the changes of the latest master, #1387 , #1405

PR #1481, opened 2021-06-04
pkg/terraform: increase the default parallelism

ipochi:
The default of 10 parallel steps does not limit creation of a small cluster but on a cluster with more than 10 nodes it leads to creation in sequential batches of 10, slowing the bring up down.

This PR is in a series of multiple PRs that will be created to ease the review/acceptance/merge of a large PR #1402

### DIFF
--- a/pkg/platform/aws/aws.go
+++ b/pkg/platform/aws/aws.go
@@ -140,7 +140,7 @@ func (c *config) Apply(ex *terraform.Executor) error {
 		return err
 	}
 
-	return ex.Apply(nil)
+	return ex.Apply([]string{terraform.WithParallelism})
 }
 
 // ApplyWithoutParallel applies Terraform configuration without parallel execution.
@@ -149,7 +149,7 @@ func (c *config) ApplyWithoutParallel(ex *terraform.Executor) error {
 		return fmt.Errorf("initializing Terraform configuration: %w", err)
 	}
 
-	return ex.Apply([]string{"-parallelism=1"})
+	return ex.Apply([]string{terraform.WithoutParallelism})
 }
 
 func (c *config) Destroy(ex *terraform.Executor) error {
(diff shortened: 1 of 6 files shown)


pkg/platform/aws/aws.go
```diff
@@ -140,7 +140,7 @@ func (c *config) Apply(ex *terraform.Executor) error {
 		return err
 	}
 
-	return ex.Apply(nil)
+	return ex.Apply([]string{"-parallelism=100"})
```
invidian, line 143: Maybe this can be part of executor default arguments? Or maybe at least we can put it into a function, so we de-duplicate the code?

invidian: Also, why 100 is a good number? Doesn't that overwhelm Terraform/CPU/RAM?

ipochi: > Maybe this can be part of executor default arguments? Or maybe at least we can put it into a function, so we de-duplicate the code?

@invidian 

I think its not big of a deal to add extra arguments this way, this is also the same for where we pass `-parallelism=1` for terraform execution without parallelism.

One thing we can do to improve is to to convernt `--parallelism=X` as a const so that we we don't have to change everywhere if there is ever a value change for `X`.

pothos: 10 meant that with 30 resources to create it took 3 times as long as needed. I just chose 100 arbitrarily with the assumption that people have less resources to create even for a very large cluster. Having 100 processes at maximum should not be an issue for normal user machines as most of these processes can share pages and don't use the CPU much but rather have to wait for an API call to finish.

invidian: I think having it as a constant and documented, that it's an arbitrary number is acceptable. In the long term, this sort of logic should be platform-independent (I'm sure there is an issue for this).

invidian (dismissed):

surajssd (approved):
LGTM

pothos (dismissed):

invidian (approved):

pothos (approved):

pothos (approved):

pothos (approved):

PR #1482, opened 2021-06-04
cli/cmd/cluster: allow to skip the control plane update

ipochi:
When the cluster is not fully set up and lokoctl is run again to fix that
problem, the user may want to avoid the control plane update happen at the
same time, because it can cause issues.

This PR is in a series of multiple PRs that will be created to ease the review/acceptance/merge of a large PR #1402


invidian (approved):
Seems OK

surajssd (approved):

Issue #441, opened 2020-05-19
Unify and document our approach on setting hostnames for cluster nodes


invidian:
Currently, we have all variety of solutions for setting hostnames/node names:
- on **AWS** we allow AWS to set machine hostnames and we use those for Node names (which together with lack of node roles (#227) gives completely ambiguous names unless you always print the labels)
- on **Packet** we set hostnames manually to format `${var.cluster_name}-controller-${count.index}` and `${var.cluster_name}-${var.pool_name}-worker-${count.index}`
- on **AKS** hostnames are controller by AKS
- on **bare-metal** we require hostnames to be provided by the user as cluster configuration

As an end user, I would expect my nodes to be consistently named when possible or to be able to set my own names on all platforms, where it is technically possible.

This issue affects also our code base, where:
- we use `--hostname-override` flag on kubelet only on `bare-metal` and `kvm-libvirt` platforms.
- each new platform requires a decision of how to handle hostnames
- different handling per platform blocks de-duplication of configuration across platforms code

vbatts:
do we even need to do `--hostname-override` at all? perhaps not?

PR #1488, opened 2021-06-09
baremetal: sets hostname controller_names/worker_names

ipochi:
Sets the hostname from the `<cluster_name>-worker-<count_index>` which
is set when `set_standard_hostname` is true, to
`controller_names<count_index>` for controllers and
`worker_names<count_index>` for workers

### DIFF
--- a/assets/terraform-modules/bare-metal/flatcar-linux/kubernetes/controller.tf
+++ b/assets/terraform-modules/bare-metal/flatcar-linux/kubernetes/controller.tf
@@ -10,6 +10,17 @@ module "controller" {
   apiserver              = format("%s.%s", var.cluster_name, var.k8s_domain_name)
   ca_cert                = module.bootkube.ca_cert
   kubelet_labels         = lookup(var.node_specific_labels, var.controller_names[count.index], {})
-  clc_snippets           = lookup(var.clc_snippets, var.controller_names[count.index], [])
-  set_standard_hostname  = true
+  set_standard_hostname  = false
+  clc_snippets = concat(lookup(var.clc_snippets, var.controller_names[count.index], []), [
+    <<EOF
+storage:
+  files:
+    - path: /etc/hostname
+      filesystem: root
+      mode: 0644
+      contents:
+        inline: ${var.controller_names[count.index]}
+EOF
+    ,
+  ])
 }
--- a/assets/terraform-modules/bare-metal/flatcar-linux/kubernetes/worker.tf
+++ b/assets/terraform-modules/bare-metal/flatcar-linux/kubernetes/worker.tf
@@ -9,7 +9,18 @@ module "worker" {
   apiserver              = format("%s.%s", var.cluster_name, var.k8s_domain_name)
   kubelet_labels         = merge(lookup(var.node_specific_labels, var.worker_names[count.index], {}), var.labels)
   cluster_name           = var.cluster_name
-  clc_snippets           = lookup(var.clc_snippets, var.worker_names[count.index], [])
-  set_standard_hostname  = true
+  set_standard_hostname  = false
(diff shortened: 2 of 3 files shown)


pothos:
To add, the hostname change got introduced by https://github.com/kinvolk/lokomotive/pull/1477 where it was set from the template variable `domain_name` coming from `var.controller_domains[count.index]`.

pothos (dismissed):

pothos:
Can you add my comment to the commit message?

assets/terraform-modules/bare-metal/flatcar-linux/kubernetes/controller.tf
```diff
@@ -10,6 +10,17 @@ module "controller" {
   apiserver              = format("%s.%s", var.cluster_name, var.k8s_domain_name)
   ca_cert                = module.bootkube.ca_cert
   kubelet_labels         = lookup(var.node_specific_labels, var.controller_names[count.index], {})
-  clc_snippets           = lookup(var.clc_snippets, var.controller_names[count.index], [])
-  set_standard_hostname  = true
+  set_standard_hostname  = false
+  clc_snippets = concat(lookup(var.clc_snippets, var.controller_names[count.index], []), [
+    <<EOF
+storage:
```
invidian, line 16: We should make sure user can still override that.

invidian (commented):
We could have some tests for this.

In general, I'd like #441 to be resolved in the first place, as it seems to me recently done changes more unify things, rather then retain old behavior, so I'm not sure about this PR. Perhaps someone more involved should decide.

assets/terraform-modules/bare-metal/flatcar-linux/kubernetes/controller.tf
pothos: You mean the user would use a different name for the machine when defining it in the configuration and setting up a domain name than the hostname would be? I think it's good to check first what the behavior of CLC snippets is when the same file gets specified by the user again and if it doesn't work, document that the user has to set the hostname through a one-time service unit.

pothos: Since this is a bug fix and does not change something related to how the user could overwrite the hostname, I think we should merge.

pothos (dismissed):

pothos (approved):

pothos (changes requested):
The `mercury-controller-0 name` in `Test_Baremetal_NodeSpecificLabels` is mismatching because the K8s node name is `node1` based on the hostname. I gues the K8s node name used to be overwritten and now isn't anymore. I think it's good to just use the user-provided name and later change the CI node naming to be more specific than `node1, `node2`, `node3`.

pothos (changes requested):
The test error message is not specific enough:
```
                if len(nodes.Items) == 1 && nodes.Items[0].Name != nodeName {
                        t.Errorf("expected node %q to have the labels %s", nodes.Items[0].Name, labels)
                }
```

I think it should be

```
                if len(nodes.Items) == 1 && nodes.Items[0].Name != nodeName {
                        t.Errorf("expected node %q but found %q to have the labels %s", nodeName, nodes.Items[0].Name, labels)
                } else {
                        t.Errorf("expected a single node, got %d", len(nodes.Items))
                }
```

invidian:
```go
                if len(nodes.Items) == 1 {
                                if nodes.Items[0].Name != nodeName  {
                                        t.Errorf("expected node %q but found %q to have the labels %s", nodeName, nodes.Items[0].Name, labels)
                                }
                } else {
                        t.Errorf("expected a single node, got %d", len(nodes.Items))
                }
```

I'd make that two independent asserts with `t.Fatal`.

pothos (approved):

pothos:
Will merge now and start a new PR based on this

PR #1489, opened 2021-06-09
baremetal: configure and persist kernel args

pothos:
- baremetal: Do not hardcode kernel console setting
    The kernel console setting was hardcoded and only used for PXE boots.
    Make the setting a variable and also apply it for the final
    installation.
- baremetal: use grub.cfg file for kernel arguments
    While the kernel arguments can be set for PXE, they have not been
    applied to the final installation.
    Use the /usr/share/oem/grub.cfg file to permanently set the kernel
    arguments, so that on each boot the final installation uses the
    kernel arguments provided by the user.


docs/configuration-reference/platforms/baremetal.md
```diff
@@ -190,7 +190,8 @@ os_version = var.custom_default_os_version
...
+| `kernel_args`                     | Additional kernel args to provide at PXE boot and in /usr/share/oem/grub.cfg.                                                                                                                                                                                                                                                                                                             | []                     | list(string)      | false    |
+| `kernel_console`                  | The kernel arguments to configure the console at PXE boot and in `/usr/share/oem/grub.cfg`.                                                                                                                                                                                                                                                                                               | ["console=tty0", "console=ttyS0"] | list(string) | false    |
```
ipochi, line 194: This is not properly formatted.

ipochi (changes requested):
small nit w.r.t formatting of the markdown table. 

Ltgm otherwise.

docs/configuration-reference/platforms/baremetal.md
pothos: I think it is, what would you change?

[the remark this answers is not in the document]
pothos: That would align it with the longer column width from `kernel_console`. `download_protocol` is also longer than the rest. Since the `kernel_args` line can be aligned with the rest, I would leave it as is.

ipochi (approved):
lgtm

PR #1490, opened 2021-06-09
baremetal: let installer.service retry on failure

pothos:
In case the network only has a temporary problem, the installer service
should start again on failure.
Convert it to a oneshot service to set RemainAfterExist, so that it is
not triggered twice if anything that depends on it is pulling it in
again after it finished.


ipochi (approved):
lgtm

ipochi:
@pothos Please rebase, that should fix the tinkerbell CI (fix was merged)

ipochi (approved):
lgtm

PR #1500, opened 2021-06-17
cert-rotation: print journal on error when restarting etcd

iaguis:
We observed CI failures when restarting etcd after copying its
certificates. This prints logs when restarting fails which is
interesting in general.


invidian (approved):

pothos:
In my case the reason was a race in Docker (it's a known bug): https://github.com/kinvolk/lokomotive/issues/1496#issuecomment-863506160

surajssd (approved):

PR #1502, opened 2021-06-17
baremetal: integrate automated (re-)provisioning logic

pothos:
-  baremetal: introduce CLC snippets for the installer    
    The installer OS needs to be able to get its network configuration as
    CLC snippet if DHCP is not used.
    
    Add CLC snippets for the installer and migrate the matchbox-specific
    CLC template to the normal CLC template which is needed for CLC
    snippets. The request template variable is replaced by the actually
    used "mac" parameter.
- baremetal: use ignition.config.url kernel parameter
    The Ignition configuration currently gets written out only once to the
    OEM partition and any updates to it in the matchbox store are not
    propagated to the node's Ignition environment when Ignition is forced
    to run again.
    
    Use the Ignition kernel parameter to point the node's Ignition
    environment directly to matchbox where the userdata lies.
- baremetal: integrate automated (re-)provisioning logic
    The bare-metal platform currently lacks any understanding of whether an
    instance is actually running the configuration that lokoctl put into
    matchbox, because, when the configuration was updated, there is no
    notification to the user that PXE booting has to be done again for this
    instance.
    Also, it is not clear to the user when to boot from PXE because the PXE
    boot must happen after lokoctl populated matchbox with the (new)
    configuration but before any other steps time out.
    The goal of this patch is to bring the baremetal platform to an
    actually usable level and support automated provisioning and
    reprovisioning, in a configurable way regardless if IPMI is used or VMs
    are created. In addition, we don't want to require a complicated PXE
    boot for each configuration update because it is slow, fragile and
    needs a special DHCP infrastructure which may be lacking in a
    production environment.
    
    Add user-defined commands in the "pxe_commands" variable to perform
    automated PXE provisioning at the right time, i.e. initally at the
    first run or when recreating a node.
    To address the problem that PXE booting is a long and maybe even manual
    process, or even impossible at production side, we can rely on Ignition
    to simulate reprovisioning by creating the "first_boot" flag file via
    SSH and issuing a reboot, which makes Ignition fetch the configuration
    from matchbox, and if we make sure to clean the root filesystem by
    formatting it, the result is the same as if reprovisioned was done with
    a PXE boot.
    The logic is achieved by a "null resource" in Terraform that executes a
    helper script which either does a PXE boot or uses SSH to trigger a
    reprovisioning with Ignition. It also handles the case of ignoring
    userdata changes for controller nodes to prevent losing etcd state.
    Since there is no notion of a baremetal node on the Terraform level
    (reminder: all this exercise here is done because we don't have a
    Terraform provider doing this for us) a local flag file is created
    under the asset directory on the machine which runs lokoctl. If it
    exists, the node was provisioned with PXE and SSH will be used for
    reprovisioning, if it does not exist, it will be provisioned with PXE
    during inital setup and for the next reprovisioning because the user
    forced recreating the node by deleting the flag file.
    Another flag file on the node is used to check whether a node was
    successfully reprovisioned.
    When SSH is used to reprovision, the kernel parameters for GRUB are
    updated directly because they are not part of the Ignition
    configuration.
    The "copy-controller-secrets" step is run after recreating a controller
    node, again since there is no notion of a node object this is solved by
    depending on the variables itself which define the node state.
    Also add user-defined commands in the "install_pre_reboot_cmds"
    variable to run after the PXE OS installation and before booting into
    the final OS, needed to set up persistent booting from disk after the
    PXE booting was configured in "pxe_commands".
    The whole patch is used by Racker (https://github.com/kinvolk/racker),
    and can be tested either with the "bootstrap/prepare.sh" script to
    create VMs with lokoctl or by running Racker in the QEMU IPMI simulator
    environment through the "racker-sim/ipmi-env.sh" script and a Racker
    Docker image built with "installer/conf.yaml" pointing to this
    Lokomotive branch.

### DIFF
--- a/assets/terraform-modules/bare-metal/flatcar-linux/kubernetes/controller.tf
+++ b/assets/terraform-modules/bare-metal/flatcar-linux/kubernetes/controller.tf
@@ -13,8 +13,22 @@ module "controller" {
   set_standard_hostname  = false
   clc_snippets = concat(lookup(var.clc_snippets, var.controller_names[count.index], []), [
     <<EOF
+filesystems:
+  - name: root
+    mount:
+      device: /dev/disk/by-label/ROOT
+      format: ext4
+      wipe_filesystem: true
+      label: ROOT
 storage:
   files:
+    - path: /ignition_ran
+      filesystem: root
+      mode: 0644
+      contents:
+        inline: |
+          Flag file indicating that Ignition ran.
+          Should be deleted by the SSH step that checks it.
     - path: /etc/hostname
       filesystem: root
       mode: 0644
--- a/assets/terraform-modules/bare-metal/flatcar-linux/kubernetes/controller_profiles.tf
+++ b/assets/terraform-modules/bare-metal/flatcar-linux/kubernetes/controller_profiles.tf
@@ -1,8 +1,10 @@
 module "controller_profile" {
   source                   = "../../../matchbox-flatcar"
   count                    = length(var.controller_names)
+  asset_dir                = var.asset_dir
   node_name                = var.controller_names[count.index]
   node_mac                 = var.controller_macs[count.index]
+  node_domain              = var.controller_domains[count.index]
   download_protocol        = var.download_protocol
   os_channel               = var.os_channel
   os_version               = var.os_version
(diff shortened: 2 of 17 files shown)


ipochi (dismissed):
If you can provide some more explanation that would be easy for others to review.

ipochi (approved):
lgtm

PR #1516, opened 2021-06-28
cluster/apply: honor --upgrade-kubelets

### LINKED ISSUE #1480: upgrade-kubelets flag is not taken into account in cluster apply
While in the pre-update health check [we use a charts list](https://github.com/kinvolk/lokomotive/blob/v0.8.0/cli/cmd/cluster/apply.go#L104-L108) that [takes into account the flag](https://github.com/kinvolk/lokomotive/blob/v0.8.0/cli/cmd/cluster/apply.go#L75), in the actual upgrade [we don't](https://github.com/kinvolk/lokomotive/blob/v0.8.0/cli/cmd/cluster/apply.go#L132) so we update the kubelet no matter what.

We should pass the flag along.

iaguis:
It was being ignored and kubelets were being upgraded no matter what.

Fixes #1480

### DIFF
--- a/cli/cmd/cluster/certificate-rotator.go
+++ b/cli/cmd/cluster/certificate-rotator.go
@@ -99,7 +99,9 @@ func rotateControlPlaneCerts(contextLogger *log.Entry, cc clusterConfig) error {
 
 	contextLogger.Log(log.InfoLevel, "Applying a controlplane update with the new CA")
 
-	if err := c.upgradeControlPlane(contextLogger, kubeconfig); err != nil {
+	upgradeKubelets := true
+
+	if err := c.upgradeControlPlane(contextLogger, kubeconfig, upgradeKubelets); err != nil {
 		return fmt.Errorf("running controlplane upgrade: %v", err)
 	}
 
--- a/cli/cmd/cluster/apply.go
+++ b/cli/cmd/cluster/apply.go
@@ -130,7 +130,7 @@ func Apply(contextLogger *log.Entry, options ApplyOptions) error {
 	if exists && !options.SkipControlPlaneUpdate && !c.platform.Meta().Managed {
 		fmt.Printf("\nEnsuring that cluster controlplane is up to date.\n")
 
-		if err := c.upgradeControlPlane(contextLogger, kubeconfig); err != nil {
+		if err := c.upgradeControlPlane(contextLogger, kubeconfig, options.UpgradeKubelets); err != nil {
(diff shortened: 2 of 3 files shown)


cli/cmd/cluster/certificate-rotator.go
```diff
@@ -99,7 +99,7 @@ func rotateControlPlaneCerts(contextLogger *log.Entry, cc clusterConfig) error {
 
 	contextLogger.Log(log.InfoLevel, "Applying a controlplane update with the new CA")
 
-	if err := c.upgradeControlPlane(contextLogger, kubeconfig); err != nil {
+	if err := c.upgradeControlPlane(contextLogger, kubeconfig, true); err != nil {
```
iaguis, line 102: Maybe it makes sense to add an option here to skip rotating the kubelet?

invidian: >Maybe it makes sense to add an option here to skip rotating the kubelet?

Any specific reason for it? I guess that would be a feature request?

invidian: Maybe this for better readability? Boolean parameters in functions are smells, so maybe at least we can name them :)
```suggestion
        upgradeKubelets := true`
        
	if err := c.upgradeControlPlane(contextLogger, kubeconfig, upgradeKubelets); err != nil {
```

invidian (commented):
Just one nit, otherwise looks OK

cli/cmd/cluster/certificate-rotator.go
iaguis: Yeah, I don't think skipping rotating the kubelet makes sense in this PR.

Sure, I'll add a bool variable.

invidian (dismissed):

surajssd (dismissed):

invidian:
CI on AWS failed at:
```
module.aws-ci1624949536-yy.null_resource.copy-controller-secrets[0] (remote-exec): Jun 29 06:53:34 ip-10-0-8-208 docker[2219]: 2021-06-29 06:53:34.597141 C | etcdmain: open /etc/ssl/etcd/etcd/peer.crt: no such file or directory
```

Does this PR include f66d9f30b7e22c1621efda056ff55ef4f98cb9ef? :thinking:

invidian (approved):

iaguis:
The only change from last reviews is a rebase so I'll merge this.

PR #1517, opened 2021-06-28
cluster/apply: enable upgrade kubelets by default

### LINKED ISSUE #110: Enable self-hosted kubelet updates by default

iaguis:
It has been tested enough and it works reliably.

Fixes #110


surajssd (approved):
LGTM

PR #1537, opened 2021-07-22
Rename Packet to Equinix Metal in existing Documentation.

ipochi:
This PR  replaces the name `Packet` with `Equinix Metal` in all the existing documentation resources under `docs/`

Fixes: #1060 

This PR only takes care of renaming on the documentation side, more work and careful thought process would be required in migrating from Packet terraform provider to Equinix Metal terraform provider, such that there is little or no effect on existing clusters during upgrade.

More specifically, this migration guide will need to be followed 
https://registry.terraform.io/providers/equinix/metal/latest/docs/guides/migration_guide_packet


invidian (approved):
LGTM, though I have some concerns about documentation talking about EM and code saying Packet all over the place. This is confusing for end user. I'd expect code changes to fall into the same release as this PR, so transition period is only "internally".

knrt10 (approved):
+1. But same concern as @invidian. I think we can create a different issue than #1060. for addressing changes on terraform level.

ipochi:
@invidian @rata @knrt10 

The user facing documentation is whats visible to the end user more so that any other piece of code/config.

I understand your concern however, making it part of the single PR would have meant a possible delay of getting the user documentation changes merged, which was top priority anyway.

Hence I closed the original issue #1060 , created another issue #1546 that tracks the code and config related changes from Packet to Equinix Metal.

#1545 tries to address #1546

PR #1553, opened 2021-08-18
metallb: update helm chart and latest version

knrt10:
Part of #1443


assets/charts/components/metallb/Chart.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v2
+name: metallb
+description: A Helm chart for MetalLB
+type: application
+version: 0.1.0
+appVersion: "v0.1.0-789-g85b7a46a"
```
invidian: Why don't we use upstream helm chart for MetalLB?

johananl: Most likely because the chart doesn't support peer autodiscovery (which isn't merged yet).

invidian: I'd still feel more comfortable using a official chart + some custom modifications, rather than still using plain manifests (which are not very good either).

knrt10: @invidian I agree. I have a different branch https://github.com/kinvolk/lokomotive/tree/knrt10/metallb-update with code from upstream and updated as per our components needs. It is throwing various errors. Once fixed I will update the PR

knrt10:
Closing in favour of https://github.com/kinvolk/lokomotive/pull/1555

PR #1561, opened 2021-08-30
baremetal: Ensure Backwards compatiblity with Lokomotive v0.8   

pothos:
- baremetal: add ignore_worker_changes option
  The baremetal platform defaults to replacing worker nodes on changes
through a simple reboot with rerunning Ignition.
This does not work for pre-0.9 deployed workers because they don't
use the ignition kernel flag to fetch the configuration. These nodes
either need a PXE boot for reprovisioning or, if the user wants to
keep the state, an additional variable to prevent them being replaced.
This additional variable is also useful if users have a workload that
doesn't play well with replacing the worker.
  Add a new variable to ignore worker config changes, similar to what
is already done on Equinix Metal, but keep replacing the workers on
by default.
-  docs/quickstarts/baremetal: use new value for k8s domain name
  The k8s domain name variable was used as control plane domain name in
Lokomotive 0.8 but afterwards got the same meaning as the DNS zone
variable in, e.g., AWS. Therefore, it is now only the suffix and the
complete control plane domain name consists of cluster name plus k8s
domain name.

  

## Testing done

Tested with https://github.com/kinvolk/racker/pull/94


pothos:
The release notes for 0.9 should contain something like that as note for existing baremetal clusters deployed with Lokomotive 0.8:

1. You must change the lokocfg value of `k8s_domain` to be only the suffix without the value of `cluster_name`, e.g., `k8s_domain_name= "k8s.localdomain"` instead of `k8s_domain_name= "mycluster.k8s.localdomain"`.
2. You must add a `pxe_commands` entry which lokoctl uses to automate the PXE (re)provisioning. For existing clusters you can use `pxe_commands = "true"` to have no PXE automation.
3. You should also set `ignore_worker_changes = true`  to prevent the worker nodes from being reprovisioned - you can disable it again later if you have a meaningful`pxe_commands` value and force a PXE provisioning process by removing the flag file (see below). After one PXE reinstallation all worker config changes are applied through a reboot via SSH by reformatting the root fs, as long as `ignore_worker_changes` is not set.
4.  Run `echo DOMAIN_NAME > lokoctl-assets/cluster-assets/MAC` for every controller and worker node to mark the node as provisioned. By removing the file you can force a PXE reprovisioning the using `pxe_commands` the next time a config change is done and `ignore_worker_changes` is not set. (With the file existing and `ignore_worker_changes` not being set, worker config changes are applied through a reboot via SSH and reformatting the root fs.)
5. Finally, run `lokoctl cluster apply` to upgrade to Lokomotive 0.9

pothos:
I updated the release note guide because reprovisioning the workers actually works out of the box once we work around a matchbox provider bug.

assets/terraform-modules/bare-metal/flatcar-linux/kubernetes/variables.tf
```diff
@@ -240,6 +240,12 @@ variable "install_pre_reboot_cmds" {
   description = "shell commands to execute on the provisioned host after installation finished and before reboot, e.g., docker run --privileged --net host --rm debian sh -c 'apt update && apt install -y ipmitool && ipmitool chassis bootdev disk options=persistent'"
 }
 
+variable "ignore_worker_changes" {
```
invidian, line 243: If this is causing troubles, perhaps we can turn the logic, so re-provisioning is optional, as on other platforms, but it's opt-in for interested parties?

This will also prevent accidental breakage on existing clusters while upgrading, if user forgets to set a specific config option.

We could also think about extending that feature, to only allow certain domains to be re-provisioned.

pothos: We can change the default, yes. Is this consistent for all platforms?

pothos: (It's not really causing troubles unless the user has important data in storage volumes)

invidian: I think it's consistent. At least on Packet we do not replace nodes during upgrades.

ipochi: @invidian  @pothos 

I've revered the behavior. It ignores the changes by default.

ipochi: > We could also think about extending that feature, to only allow certain domains to be re-provisioned.

We can think about it as a new feature. Probably out of scope for this PR.

ipochi:
@pothos @invidian @iaguis 

Tested multiple upgrade paths :

1. Ignoring the worker changes and updating to the 0.9
2. Re-provision the worker nodes whilst updating to 0.9
3. Updating user data but not re-provisioning whilst updating to 0.9

surajssd (dismissed):

iaguis (changes requested):
Just a commit message nit

assets/terraform-modules/bare-metal/flatcar-linux/kubernetes/bootkube.tf
```diff
@@ -5,7 +5,7 @@ module "bootkube" {
   cluster_name = var.cluster_name
   api_servers  = [format("%s.%s", var.cluster_name, var.k8s_domain_name)]
   # Each instance of controller module generates the same set of etcd_servers.
-  etcd_servers                    = module.controller[0].etcd_servers
+  etcd_servers                    = concat(var.controller_domains, module.controller[0].etcd_servers)
```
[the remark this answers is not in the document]
invidian: Also, commit with count variable adding should ideally go before this one, otherwise it creates broken commit.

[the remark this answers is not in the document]
ipochi: done.

assets/terraform-modules/bare-metal/flatcar-linux/kubernetes/variables.tf
ipochi: @invidian Default is `true` , i.e it ignores the worker changes and doesn't reprovision.

If one wants re-provisioning of the workers the user must set `ignore_worker_changes = false`

invidian: But this is default in Terraform, which will be overwritten by Go template, as zero value of bool is `false` in Go.

iaguis (approved):
lgtm. Thanks!

assets/terraform-modules/bare-metal/flatcar-linux/kubernetes/variables.tf
ipochi: @invidian updated.

iaguis (approved):

invidian (approved):
Looks good

PR #1564, opened 2021-09-08
kubernetes: Update to v1.22.1

ipochi:
Update Kubernetes to the latest version. v1.22.1


knrt10 (dismissed):
+1

iaguis:
From talking with @invidian it seems updating to v1.22 breaks may things due to removed APIs so I think we should hold off this to after v0.9.0.

invidian:
Yes, it seems velero, gangway, httpbin, web-ui, and prometheus-operator all use some API types which has been removed. Also some apps don't support 1.22 yet, like external-dns.

ipochi:
Yes, I thought of giving it a try as it was a low hanging fruit for 0.9.
It seems bootstrap control plane pods itself didn't start.

PR #1568, opened 2021-09-09
docs: update build instructions for kubelet

ipochi:
The earlier instructions to build the kubelet image builds the same
image for both amd64 and arm64 if building both on the same machine.

Using docker's buildx we can build multiple architecture images on the
same machine.


ipochi:
closing as this is not needed anymore.
Covered in https://github.com/kinvolk/lokomotive/pull/1567/commits/ca039c01bb62b514967fbe704290d3f06f85029d

